PR #1, opened 2015-05-25
Custom message when max connections are reached

arvindr21:
Custom message when max connections are reached - updated: source code, test & docs


nfriedly:
Awesome, thanks!

arvindr21:
:+1:

PR #2, opened 2015-06-28
Fix reset max connections reached after a accept new connections from…

MikeBild:
… blocked IP.

Issue: after an reset of a blocked IP all connections still open, because hits[ip] === NaN.
Fix: Reset hits[ip] to a Number

### DIFF
--- a/lib/express-rate-limit.js
+++ b/lib/express-rate-limit.js
@@ -33,6 +33,10 @@ function RateLimit(options) {
             }
         }, options.windowMs);
 
+        if(!hits[ip]) {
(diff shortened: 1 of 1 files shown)


nfriedly:
Hi Mike,

Maybe I'm just groggy, but I think this change is unnecessary. Did you run into the issue you described? Can you include a test case that fails before but passes after this change?

nfriedly:
So, my apologies for being groggy, there was a bug here. Thanks for your help in getting it fixed!  `v1.0.3` is now live on npm.

PR #3, opened 2015-07-27
FIX: Initialize counter to 1

MatthewPhinney:
If the counter is initialized to zero, the decrement function is called one too many times.  It resets hits[ip] to undefined when it reaches zero, but then it decrements again, which sets hits[ip] = NaN (`undefined--` evaluates to `NaN`.  `typeof NaN === 'number'`, so this error is never corrected, and the limiter fails to work correctly for all future requests.

### DIFF
--- a/lib/express-rate-limit.js
+++ b/lib/express-rate-limit.js
@@ -28,7 +28,7 @@ function RateLimit(options) {
         setTimeout(function() {
             // cleanup
             hits[ip]--;
-            if (hits[ip] <=0 ) {
+            if (hits[ip] < 0 ) {
                 delete hits[ip];
             }
         }, options.windowMs);
--- a/test/test.js
+++ b/test/test.js
@@ -168,4 +168,25 @@ describe('express-rate-limit node module', function() {
             });
         }, 60);
     });
+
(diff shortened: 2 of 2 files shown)


nfriedly:
Good catch, but now the tests are failing :/

Can you make it keep the same behavior for the normal case? (I think it might be as simple as changing `>=` to just `>` on [line 36](https://github.com/nfriedly/express-rate-limit/blob/master/lib/express-rate-limit.js#L36).)

Also, can you add a test that reproduces this condition and fails on the current master branch but passes on yours?

nfriedly:
So, I thought about it a little more and realized that that just the `>=` to `>` wouldn't be enough, but I do like the idea of starting at 1 because it just makes mores sense. So I included your change and reworked the other things to make the tests pass.

I also added a new test that goes through the entire flow of new -> blocked -> unblocked -> blocked again -> unblocked again to ensure this dosen't pop up again. It failed on the old code but passes now. 

Thanks for the help!

MatthewPhinney:
Awesome -- thanks for publishing the new version :)

nfriedly:
Oh, right, I forgot to mention that but looks like you found it. You're welcome :)

PR #10, opened 2016-01-18
Add a handler option

MtDalPizzol:
The handler defaults to a function which receives a response object of Express
and negociates the content of the response based on the Contetn-Type header of the request.

Files changed:
modified:   README.md
modified:   lib/express-rate-limit.js

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -22,7 +22,16 @@ describe('express-rate-limit node module', function() {
     function createAppWith(limit) {
         app = express();
         app.all('/', limit, function(req, res) {
-            res.send('response!');
+          res.format({
+            html: function () {
+              res.send('response!');
+            },
+            json: function () {
+              res.json({
+                message: /response!/
+              });
+            }
+          });
         });
         // helper endpoint to know what ip test requests come from
         // set in headers so that I don't have to deal with the body being a stream
@@ -49,11 +58,47 @@ describe('express-rate-limit node module', function() {
             });
     }
 
+    function goodJsonRequest(errorHandler, successHandler) {
+        request(app)
+            .get('/')
+            .set('Accept', 'application/json')
+            .expect('Content-Type', /json/)
+            .expect(200, {
+              message: /response!/
+            })
+            .end(function(err, res) {
+                if (err) {
+                    return errorHandler(err);
+                }
+                delay = Date.now() - start;
+                if (successHandler) {
+                    successHandler(null, res);
+                }
+            });
+    }
+
     function badRequest(errorHandler, successHandler) {
         request(app)
             .get('/')
             .expect(429)
             .expect(/Too many requests/)
+            .end(function(err, res) {
+                if (err) {
+                    return errorHandler(err);
+                }
+                delay = Date.now() - start;
+                if (successHandler) {
+                    successHandler(null, res);
+                }
+            });
+    }
+
+    function badJsonRequest(errorHandler, successHandler) {
+        request(app)
+            .get('/')
+            .set('Accept', 'application/json')
+            .expect('Content-Type', /json/)
+            .expect(429, { message: 'Too many requests, please try again later.' })
             .end(function(err, res) {
                 if (err) {
                     return errorHandler(err);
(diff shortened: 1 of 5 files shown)


nfriedly:
Hey, thanks!

I have a few requests and a question before merging this in:

Question: should we instead just throw an exception and let users have their own express error-handler handle it? That would be a simple way of allowing this, although I'm honestly not sure if it's better than your approach. (The main reason that it comes to mind is to get a complete, styled html page that matches the template from the rest of the site.)

Requests:

1. Merge/rebase `master` now that the tests are passing (sorry, this was my bad.)
2. Add a couple of tests to verify expected behavior - one with an `Accept: application/json`, and one with a "real" `Accept:` header lifted from a browser.
3. I think we should only specify `application/json` and a `default` handler: skip `text/plain` because the default should catch it, and skip `text/html` because a) anything that can handle html can handle text, and b) that's not valid html and I don't think it's worth the effort to produce valid html.

dougwilson:
> should we instead just throw an exception and let users have their own express error-handler handle it? That would be a simple way of allowing this, although I'm honestly not sure if it's better than your approach.

If it helps, I may be able to provide a little bit of context here. Typically for this type of flow, I usually lean more towards recommending the `next(err)` approach, otherwise you may find you have to start adding a new `handler` option for each kind of way people want to reply to error conditions. I usually find that is provides the benefit that applications can have much more centralized error-handling, in the form on an error-handler.

A good example is the returning a fully-formatted error page, where an application only needs to define this handling and rendering in a single location, rather than duplicating the logic into this handler option. Typically APIs also like to have a centralized error-handing as well, to provide a consistent JSON error payload, for example.

If you do go the error approach, Express has a built-in default error handler that will use properties from the `err` object to construct the request (currently only `err.statusCode` or `err.status`), but adding others that signal information about the rate limit can simply be added and documented to the users.

Ultimately, the API choice is up to the module, though :) Just my 2 cents from the Gitter question :)

nfriedly:
Thanks @dougwilson!

(BTW, @MtDalPizzol, I also asked about which way to handle it on https://gitter.im/strongloop/express - And, obviously, if we just let express handle it, then requests 3 and 4 don't apply.)

MtDalPizzol:
Hey everyone... thanx for the feedback. Actually, having a handler with a signature ```function (req, res, next)``` was my first idea, then I thought something stupid and went with a different aproach. I'll take care of this today.

MtDalPizzol:
So, what I'm going to do is simply to add a **handler** option that defaults to this:

```js
function (req, res, next)  {
   res.format({
       html: function () {
           res.status(options.statusCode).end(options.message);
       },
       json: function () {
           res.json.status(options.statusCode).json({
               message: options.message
           });
       }
   });
}
```

Then people are free to do whatever they think is best...

I have a doubt, since I'm new to Github. Should I send the next PR with the version of the module updated in the package.json?

nfriedly:
No, you should leave updating package.json to me, there's a `npm version` command that does it automatically. 

Also, you don't need to create a new PR, any changes you push to your `master` branch will automatically go with this PR.

nfriedly:
Also, I think that this is probably good for a minor version release (v2.1.0), but I may still go back and switch it to passing an error instead of sending the response directly. However, that would be a major version change (v3.0.0).

MtDalPizzol:
Oh! Ok then! Thanx  for the pacience... baby steps here...

MtDalPizzol:
Almost done here. Only a few more things.

1. To get a middleware as the default handler option and the tests working, I had to set .jshint ```unused: false```, since the ```next``` function isn't called into the default handler. Is it ok?

2. I wrote a test for the ```Accept: application/json``` and it is working well. But, sorry for the ignorance: I didn't understood what did you meant by "one with a 'real' ```Accept:``` header lifted from a browser".

nfriedly:
For 1), just skip the `next` argument. If you don't use it, then you don't need it in the function's signature.  (You could leave it commented out like `handler: function(req, res /*, next */) {...}` if you want to keep a reminder there that it's available.) 

I admit that this is somewhat of a failing of jshint, but I don't want `unused: false` because it does help catch real bugs sometimes.

For 2), I meant, use the same `Accept:` header that a browser would send when requesting a page. So for example, I just loaded github.com with firefox's developer tools open and checked the request in the network tab. It listed all of the headers that firefox sent and all of the ones that the github.com server responded with. Firefox's accept header was `Accept: text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8` - so that one should be good for your test. Just verify that it responds to that header with text and not JSON. 

Also, not sure why I didn't think about this sooner, but there should also be one more test that supplies a custom `handler` function and verifies that it was actually called and used to respond to the request.

MtDalPizzol:
Hey guys... I already have pushed the latest changes. All the 15 tests are passing here. Any ideas why Travis is failing?

nfriedly:
Arc, its a latency test and Travis is apparently too slow to run it reliably. I need to put some more work in to that. But, your changes look good. I'm on my phone at the moment, but I'll merge them and ship a new release later on today.

test/test.js
```diff
@@ -22,7 +22,16 @@ describe('express-rate-limit node module', function() {
...
         app = express();
         app.all('/', limit, function(req, res) {
-            res.send('response!');
+          res.format({
+            html: function () {
+              res.send('response!');
+            },
+            json: function () {
+              res.json({
+                message: /response!/
```
nfriedly: Shouldn't this be `message: "response"`?

test/test.js
```diff
@@ -49,6 +58,25 @@ describe('express-rate-limit node module', function() {
             });
     }
 
+    function goodJsonRequest(errorHandler, successHandler) {
+        request(app)
+            .get('/')
+            .set('Accept', 'application/json')
+            .expect('Content-Type', /json/)
+            .expect(200, {
+              message: /response!/
```
nfriedly: Ditto.

nfriedly:
I went and changed it to a string after merging (regexes aren't really valid in JSON. It happened to work here but I'd prefer to stick to valid JSON for the testing.)

And, I released it as v2.1.0 on npm. Thanks for the help!

If I ever get to it, I think v3 is going to drop the `handler` method in favor of calling `next(err)`, but this should work for now.

MtDalPizzol:
Yeah! You're right! My bad on that one with the regex...
I'm going to pull the new version to my project right now!
Thanks for the work on the module and the tips with the PR!

nfriedly:
You're welcome - thanks for doing all the hard work!

PR #14, opened 2016-03-10
customize how IP is obtained from req object

yufeiliu:
`req.ip` does not reliably represent client's IP address in load-balanced environments like Google App Engine and Heroku. This change allows users to define how to retrieve IP address from a request object, so they could use a custom header field if needed.


nfriedly:
Hi @yufeiliu

Thanks for the PR, but I don't think it should be necessary - express should handle this automatically when call `app.enable('trust proxy');`  

I haven't tried GAE, but it's worked well for me on Heroku and Bluemix.

See https://github.com/expressjs/express/blob/master/lib/request.js#L326 and the package that it uses, https://www.npmjs.com/package/proxy-addr

If you've found a case where that doesn't work, it should probably be filed as a bug on the `proxy-addr` module or else handled in custom middleware rather than by modifying this one.

PR #23, opened 2016-04-22
Custom condition

karatheodory:
Dear Sir,

In this pull request I have added a key generation function which allows the users to implement custom keys, not only users' IPs.

Thanks!

### DIFF
--- a/lib/express-rate-limit.js
+++ b/lib/express-rate-limit.js
@@ -70,13 +70,9 @@ function RateLimit(options) {
         delete hits[key];
     }
 
-    // Backward compatibility function
-    function resetIp(ip) {
-        return resetKey(ip);
-    }
-
     rateLimit.resetKey = resetKey;
-    rateLimit.resetIp = resetIp;
+    // Backward compatibility function
+    rateLimit.resetIp = resetKey;
 
     return rateLimit;
 }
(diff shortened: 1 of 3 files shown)


lib/express-rate-limit.js
```diff
@@ -62,11 +66,16 @@ function RateLimit(options) {
...
+    }
 
+    // Backward compatibility function
     function resetIp(ip) {
-        delete hits[ip];
+        return resetKey(ip);
     }
 
+    rateLimit.resetKey = resetKey;
     rateLimit.resetIp = resetIp;
```
nfriedly: You could just set `ratelimit.resetIp = resetKey;` and then get rid of the `resetIp()` function all together.

nfriedly:
Hey, nice work @karatheodory, this looks like a smart change.

I left one comment about a slight simplification, and then I'd also like to see a test that validates the functionality. (Maybe include a token in the querystring and a custom function that uses the token instead of the IP - check that that one token gets blocked after too many requests, and also check that a second token does not get blocked once just because the first one is.

lib/express-rate-limit.js
karatheodory: That's right :smile: Thanks for your comment, I will fix it later today.

karatheodory:
@nfriedly Hello! Can you please review the updated pull request?

nfriedly:
Yep, it looks good. Nice work!

nfriedly:
It should be released as v0.2.2 on npm automatically after Travis CI finishes it's thing.

PR #24, opened 2016-05-09
The ability to use a custom source if IP to be used as key

bendpx:
the problem to solve is that in cases where a proxy or load balancer is on top of the express application, req.ip might be populated with the lb/proxy ip. and then the limiter will block every user.

this way you can define the source of the IP, a custom header or request param, where it will usually be populated in.


nfriedly:
Hi @bendpx - thanks for putting in the work to add this, but I think it shouldn't be necessary. For the case you describe, express handles this automatically - just set `app.enable('trust proxy');` and then `req.ip` will automatically be set based on headers from the proxy/load balancer/etc. More details on that at http://expressjs.com/en/4x/api.html#app.settings.table

If express's built-in ip-handling doesn't work for some reason, then I'd recommend using the `keyGenerator` option - for example: 

```js
var limiter = new RateLimit({
  keyGenerator: function(req, res) {
    return req.realIpHeader;
  }
});
```

Is that sufficient, or is there another angle that I'm missing here?

PR #25, opened 2016-05-18
Abstracted backend store out from RateLimit

wyattjoh:
I really liked the simplicity of the package, but I wanted to extend it for use with other stores.

I've abstracted out the Memory Store into a model that can now be plugged in as need be with other databases. The only requirements of any other store is that it must provide the following methods:

- `incr`: This method must accept a key, and a callback, the key will be incremented and the new value returned in the callback, along with any error that occurred.
- `resetKey`: This method resets the key indicated in an atomic operation
- `resetAll` (optional): This method resets all the keys stored for the rate limiting middleware

Would love for any feedback :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,7 +123,7 @@ var SomeStore = function() {
     };
 
     /**
-     * This callback is called buy the underlying store when an answer to the
+     * This callback is called by the underlying store when an answer to the
      * increment is available.
      * @callback Store~incrCallback
      * @param {?object} err - The error from the underlying store, or null if no
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@
 
 Basic rate-limiting middleware for Express. Use to limit repeated requests to public APIs and/or endpoints such as password reset.
 
-Note: this module does not share state with other processes/servers.
+Note: this module does not share state with other processes/servers by default.
 If you need a more robust solution, I recommend checking out [strict-rate-limiter](https://www.npmjs.com/package/strict-rate-limiter) or [express-brute](https://www.npmjs.com/package/express-brute), both are excellent pieces of software.
 
 
@@ -106,6 +106,40 @@ function (req, res, /*next*/) {
     }
   });
 }
+```
+* **store**: The storage to use when persisting rate limit attempts. By default, the [MemoryStore](lib/memory-store.js) is used. It must implement the following in order to function:
+```js
+var SomeStore = function() {
+    /**
+      * Increments the value in the underlying store for the given key.
+      * @method function
+      * @param {string} key - The key to use as the unique identifier passed
+      *                     down from RateLimit.
+      * @param {Store~incrCallback} cb - The callback issued when the underlying
+      *                                store is finished.
+      */
+    this.incr = function(key, cb) {
+      // ...
+    };
+
+    /**
+     * This callback is called buy the underlying store when an answer to the
+     * increment is available.
+     * @callback Store~incrCallback
+     * @param {?object} err - The error from the underlying store, or null if no
+     *                      error occurred.
+     * @param {number} value - The current value of the counter
+     */
+
+    /**
+     * Resets a value with the given key.
+     * @method function
+     * @param  {[type]} key - The key to reset
+     */
+    this.resetKey = function(key) {
+      // ...
+    };
+};
 ```
 
 The `delayAfter` and `delayMs` options were written for human-facing pages such as login and password reset forms.
--- a/lib/express-rate-limit.js
+++ b/lib/express-rate-limit.js
@@ -84,10 +84,10 @@ function RateLimit(options) {
         });
     }
 
-    rateLimit.resetKey = options.store.resetKey.bind(options.store);;
+    rateLimit.resetKey = options.store.resetKey.bind(options.store);
 
     // Backward compatibility function
-    rateLimit.resetIp = options.store.resetKey.bind(options.store);;
+    rateLimit.resetIp = options.store.resetKey.bind(options.store);
 
     return rateLimit;
 }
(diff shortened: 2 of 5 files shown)


lib/express-rate-limit.js
```diff
@@ -27,52 +53,41 @@ function RateLimit(options) {
         }
     });
 
+    options = defaults(options, {
```
nfriedly: can this be merged with the previous `defaults()` call?

wyattjoh: I needed a valid (default or not) `options.windowMs`, and that wasn't available until after the defaults were applied first on it.

lib/express-rate-limit.js
```diff
@@ -27,52 +53,41 @@ function RateLimit(options) {
...
     }
 
-    // simply reset ALL hits every windowMs
-    setInterval(resetAll, options.windowMs);
-
-    // export an API to allow hits from one or all IPs to be reset
-    function resetKey(key) {
-        delete hits[key];
-    }
+    rateLimit.resetKey = options.store.resetKey;
```
nfriedly: this should have it's scope bound:

    rateLimit.resetKey = options.store.resetKey.bind(options.store);

 (MemoryStore doesn't happen to use `this` but presumably other stores could.)

lib/express-rate-limit.js
```diff
@@ -27,52 +53,41 @@ function RateLimit(options) {
...
 
-    rateLimit.resetKey = resetKey;
     // Backward compatibility function
-    rateLimit.resetIp = resetKey;
+    rateLimit.resetIp = options.store.resetKey;
```
nfriedly: Ditto above, except that you can re-use the bound function above instead of re-binding it:

  rateLimit.resetIp = rateLimit.resetKey;

lib/express-rate-limit.js
nfriedly: aah, of course. Sorry, I'm running low on sleep.

nfriedly:
Cool, nice work. I was a little bit hesitant when I first saw this, because one of my goals with this module is simplicity, but I think I'll go ahead and merge it in with a couple more changes:

* Can you move the memory store to it's own file?
* The storage model needs to be documented in the readme. You can more or less copy-paste from the pull request though, but the bit about "this module does not share state with other processes/servers" should say "by default", and have a note about plugin in a custom store to achieve that.
* There needs to be some additional tests mock the store to ensure that RateLimit calls the correct methods (I suppose just one for checking `incr`.) 
* Ideally the tests would be specific to RateLimit and to MemoryStore, and probably split into two test files.. but that might be a bit much to ask. We'll call that one extra credit.

Oh, and just out of curiosity, why not use one of the other

lib/express-rate-limit.js
nfriedly: Ok, one idea, and this is a small thing, but I think an improvement. When you call defaults and pass an object with anew MemoryStore instance, that MemoryStore is created weather it ends up being use or not. So if the user passes in a custom store, the default MemoryStore will still be created and use up a (tiny) bit of resources.

If you did this instead, it would only create a MemoryStore instance when the user didn't specify any options.store:

    options.store = options.store || new MemoryStore(options);

This works because 

1. JS lets you use the logical or (`||`) on objects, returning the first "truthy" one (meaning that it exists) and 
2. it uses "short circuit evaluation" - if the left side of `||` is truthy, the the right side doesn't even get evaluated, so no new MemoryStore instance is created.

wyattjoh:
I'm currently in the process of adding the enhancements you specified.

I didn't use Brute as their API to me seemed pretty complicated, and I honestly had difficulty finding out how to do what I want to do. If you feel that this PR isn't in scope with what you'd rather have the project do, I wouldn't be offended if you rejected it. I thought that a simple abstraction without preference to a particular store might have allowed extension rather than locking you to the simple memory solution.

nfriedly:
Yea, while it does add a small amount of complexity, I think this is a good call. I think I'm going to release a v3 sometime soon that removes the delay option because no one seems to use it, so maybe that will be my complexity counterbalance ;)

wyattjoh:
I should have hit all your suggestions (including the extra credit :) ). Let me know if there are any issues.

README.md
```diff
@@ -107,6 +107,40 @@ function (req, res, /*next*/) {
...
+      *                     down from RateLimit.
+      * @param {Store~incrCallback} cb - The callback issued when the underlying
+      *                                store is finished.
+      */
+    this.incr = function(key, cb) {
+      // ...
+    };
+
+    /**
+     * This callback is called buy the underlying store when an answer to the
```
nfriedly: Typo: it should be "by", not "buy" ;)

lib/express-rate-limit.js
```diff
@@ -1,11 +1,12 @@
 'use strict';
 var defaults = require('defaults');
+var MemoryStore = require('./memory-store');
 
 function RateLimit(options) {
 
     options = defaults(options, {
         // window, delay, and max apply per-key unless global is set to true
-        windowMs: 60 * 1000, // milliseconds - how long to keep records of requests in memory
+        windowMs: 60 * 1000, // milliseconds - how long to keep records of requests in memory: Only used when store is MemoryStore
```
nfriedly: I think we should omit this addition... presumably other stores could also use windowMs if they wanted.

nfriedly:
I mentioned a couple of little comment things, but this looks good overall. Nice work!

nfriedly:
Ok, this should be out in v2.3.1 as soon as the CI loop finishes. (I put it in v2.3.0 but then decided to add a `"use strict"; to the top of `memory-store.js`.)

wyattjoh:
Nice! Created this one btw also: https://github.com/wyattjoh/rate-limit-redis

PR #26, opened 2016-05-29
RateLimit Header

Pe46dro:
Added the possibility to show request limit and remaining request
available in HTTP response header

To Do:
- Add reset timer in header

### DIFF
--- a/lib/express-rate-limit.js
+++ b/lib/express-rate-limit.js
@@ -49,28 +49,29 @@ function RateLimit(options) {
 
         options.store.incr(key, function(err, current) {
             if (err) {
-            	return next(err);
-            }
-
-            if(options.headers){
-                res.setHeader('X-RateLimit-Limit', options.max);
-                res.setHeader('X-RateLimit-Remaining', ((options.max-current) > 0)?(options.max-current):0);
-            }else{
-                req.rateLimit = ({
-                    limit: options.max,
-                    remaining: ((options.max-current) > 0)?(options.max-current):0
-                });
-            }
-
-            if (options.max && current > options.max) {
-                return options.handler(req,res, next);
+              return next(err);
             }
 
             if (options.delayAfter && options.delayMs && current > options.delayAfter) {
-            	var delay = (current - options.delayAfter) * options.delayMs;
-                setTimeout(next, delay);
+              var delay = (current - options.delayAfter) * options.delayMs;
+              setTimeout(next, delay);
             } else {
+
+              if (options.headers) {
+                res.setHeader('X-RateLimit-Limit', options.max);
+                res.setHeader('X-RateLimit-Remaining', ((options.max - current) > 0) ? (options.max - current) : 0);
+              }
+
+              req.rateLimit = ({
+                limit: options.max,
+                remaining: ((options.max - current) > 0) ? (options.max - current) : 0
+              });
+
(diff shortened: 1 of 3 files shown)


nfriedly:
Hey, thanks for this,

I was hoping there would be some standard for these headers, but based on http://stackoverflow.com/questions/16022624/examples-of-http-api-rate-limiting-http-response-headers#16022625 there doesn't really seem to be.

I like the idea here, the only part I'm iffy about is adding another arbitrary function to the options - I think this is simple enough that it could be handled more simply, perhaps just a single boolean and then expose the values on the req object for users that want to customize things.

So, could I ask for an alternate version that does something like this: setting `options.headers === true` (should probably be the default) adds the headers, while `options.headers === false` prevents them. And then either way, we set `req.rateLimit` to something like 

```
{
 limit: 456
 remaining: 123
 reset: new Date(/*...*/)
}
```

For anyone who wants different header names, they can set options.headers to false and then use the req.rateLimit object to set their own headers.

Actually, the reset part is trickey due to the store controlling it and being abstracted.. lets just skip that for now. I need to release a v3 with some breaking changes one of these days, the reset time might make sense there.

nfriedly:
Oh, and it needs some new tests also. Shouldn't be too hard, especially if we reduce the config to just a boolean.

Pe46dro:
Hey,  sorry for the delay getting back to you, i just committed the module with your suggestion, i still have to do tests because i have to study supertest's documentation.

lib/express-rate-limit.js
```diff
@@ -48,20 +48,30 @@ function RateLimit(options) {
...
+            }
+
+            if(options.headers){
+                res.setHeader('X-RateLimit-Limit', options.max);
+                res.setHeader('X-RateLimit-Remaining', ((options.max-current) > 0)?(options.max-current):0);
+            }else{
+                req.rateLimit = ({
+                    limit: options.max,
+                    remaining: ((options.max-current) > 0)?(options.max-current):0
+                });
```
nfriedly: This part doesn't need to be in an else block, it should happen either way.

lib/express-rate-limit.js
```diff
@@ -48,20 +48,31 @@ function RateLimit(options) {
...
-          if (options.max && current > options.max) {
-              return options.handler(req,res, next);
-          }
+            if (options.headers) {
+              res.setHeader('X-RateLimit-Limit', options.max);
+              res.setHeader('X-RateLimit-Remaining', ((options.max - current) > 0) ? (options.max - current) : 0);
+            }
 
-          if (options.delayAfter && options.delayMs && current > options.delayAfter) {
+            req.rateLimit = ({
```
nfriedly: Hey, sorry I missed that you had updated this PR. You don't need parenthesis here, but I'm going to go ahead and merge this anyways

lib/express-rate-limit.js
```diff
@@ -48,20 +48,31 @@ function RateLimit(options) {
...
+              limit: options.max,
+              remaining: ((options.max - current) > 0) ? (options.max - current) : 0
```
nfriedly: This would be simpler with `Math.max`, but again, good enough for now

nfriedly:
This should be released in `v2.4.0` once the CI loop finishes

PR #29, opened 2016-09-01
Allow to skip requests

bogini:
Useful for whitelisting.


RyanTheAllmighty:
@bogini any plans to make those additions suggested or should another PR be made by another user so we can get this merged in?

PR #30, opened 2016-09-07
Make it possible to clear the memory store interval

samshull:


nfriedly:
Hi @samshull,

Can I ask what the need / reasoning is behind this?

samshull:
@nfriedly When performing an integration test against an app that uses this middleware, the process is left open because the interval cannot be cleared. Currently, the only way to work around that is `process.exit()`. This would allow the interval to be cleaned up without forcing the process to end.

nfriedly:
Aah, that's a fair point. Although, I think the correct solution is to use https://nodejs.org/api/timers.html#timers_timeout_unref - setTimeout(...).unref();

Can you update the PR with that?

samshull:
@nfriedly updated, and I agree much more elegant

samshull:
This needs to be published, please.

nfriedly:
Doh, sorry. I think I did the merge on my phone. Should be out in `v2.5.0` once the CI loop finishes.

PR #45, opened 2017-05-15
Feature events

tspoke:
Hi,

I needed an event to log infos when an user reach the limit and I didn't want to override default handler method only for that. This PR add a function to this use-case.

Thanks !

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,6 +116,12 @@ function (req, res, /*next*/) {
   });
 }
 ```
+* **onLimitReached**: Function to listen each time the limit is reached. You can use it to debug/log. Defaults:
+```js
+function (req, res, options) {
+  /* empty */
+}
+```
 * **store**: The storage to use when persisting rate limit attempts. By default, the [MemoryStore](lib/memory-store.js) is used. It must implement the following in order to function:
 ```js
 function SomeStore() {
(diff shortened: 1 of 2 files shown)


nfriedly:
Hi @tspoke,

Thanks! 

I see what you're going for, but I'm a little bit iffy about the implementation, because I feel like it could cause some confusion.  

What would you say to making the default function accessible, and then to logging would look something like this?

```js
var RateLimit = require('express-rate-limit');

var limiter = new RateLimit({
  handler: function(req, res, next) {
    console.log('limit reached');
    return limiter.defaultHandler(req, res, next);
  }
});
```
Actually, I'm not sure I really like that API either now that I'm looking at it... 

I don't know, do you have any ideas here?

tspoke:
Hi @nfriedly ,

Thank you for your fast answer !

I understand your point about my implementation. 

You solution is interesting but I think it's important to make the difference between the `handler()` method and the potential observers attached to it. In your example you link together the event and the effect into one function. Adding `onLimitReached` as another attribute split the two behaviors.

Your call about this, I can change the PR to match your solution ;)

nfriedly:
Yea... lets ship it like this and see what happens. I have some plans to do a bit of simplification for v3, so if we come up with any better ideas, we can try them out there.

nfriedly:
Should be released as v2.8 once it passes CI

tspoke:
Thank you @nfriedly !

PR #48, opened 2017-07-24
Runs in electron where setInterval() returns only a number

adiel:
We run integration tests in Electron where we were see a different return value from `setInterval()` than in pure node causing `unset` to be undefined.

### DIFF
--- a/lib/memory-store.js
+++ b/lib/memory-store.js
@@ -22,7 +22,10 @@ function MemoryStore(windowMs) {
   };
 
   // simply reset ALL hits every windowMs
-  setInterval(this.resetAll, windowMs).unref();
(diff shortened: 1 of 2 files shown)


nfriedly:
Weird.. but OK. I'll get this merged in and released shortly.

Also, why would you need rate-limiting in an Electron app?

adiel:
Yeah, it's weird but in Electron node it's returning just 1,2,3 etc. It's just our integration tests that we run within Electron (it works fine for real). This is pretty neat as we have our tests, our front end JS and our node services all running in the same process in these tests.

nfriedly:
Released in `v2.8.1`

PR #49, opened 2017-07-24
Retry-After header

adiel:
Adds basic support for the `Retry-After` header

https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/429

I say "basic" as all I have done is return the windowMs in seconds. It would be preferable to actually return how long until the current window will elapse but I might leave that to you @nfriedly :)

### DIFF
--- a/lib/express-rate-limit.js
+++ b/lib/express-rate-limit.js
@@ -21,6 +21,9 @@ function RateLimit(options) {
             return false;
         },
         handler: function (req, res /*, next*/) {
+          if (options.headers) {
+            res.setHeader('Retry-After', Math.ceil(options.windowMs / 1000));
+          }
           res.format({
             html: function(){
               res.status(options.statusCode).end(options.message);
@@ -73,7 +76,6 @@ function RateLimit(options) {
 
             if (options.max && current > options.max) {
               options.onLimitReached(req, res, options);
-              res.setHeader('Retry-After', Math.ceil(options.windowMs / 1000));
               return options.handler(req, res, next);
             }
 
--- a/lib/express-rate-limit.js
+++ b/lib/express-rate-limit.js
@@ -69,6 +69,7 @@ function RateLimit(options) {
             if (options.headers) {
               res.setHeader('X-RateLimit-Limit', options.max);
               res.setHeader('X-RateLimit-Remaining', req.rateLimit.remaining);
+              res.setHeader('Retry-After', Math.ceil(options.windowMs / 1000));
             }
 
             if (options.max && current > options.max) {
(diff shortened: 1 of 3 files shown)


test/express-rate-limit.js
```diff
@@ -230,16 +229,16 @@ describe('express-rate-limit node module', function() {
...
+        createAppWith(rateLimit({windowMs: 59100}));
         goodRequest(done, function( /* err, res */ ) {
             delay = Date.now() - start;
             if (delay > 99) {
                 done(new Error("First request took too long: " + delay + "ms"));
             } else {
                 done();
             }
-        },true, 5, 4);
+        }, undefined, true, '5', '4', '60');
```
nfriedly: good catch. Not sure how this was passing before though :/

nfriedly:
Hey, thanks!

I think this will need a little work before we can ship it - the `Retry-After` header should only be sent for rejected requests, but this looks like it puts them on all requests. Perhaps you could move it to the default `handler` callback? 

I agree that setting it to `windowMs` isn't ideal, but since the behavior can change depending on the store (in-memory vs redis), anything else would require work both here and in the `rate-limit-redis` library. So I think it's acceptable.

And then, of course, the tests will have to change to expect it in `badRequest` instead of `goodRequest`.

adiel:
Yep, good point, will sort it.

lib/express-rate-limit.js
```diff
@@ -73,6 +73,7 @@ function RateLimit(options) {
 
             if (options.max && current > options.max) {
               options.onLimitReached(req, res, options);
+              res.setHeader('Retry-After', Math.ceil(options.windowMs / 1000));
```
nfriedly: This should be conditioned on `options.headers`, and I was thinking it would be inside of the default `options.handler` so that it could be overridden.

test/express-rate-limit.js
```diff
@@ -81,9 +81,12 @@ describe('express-rate-limit node module', function() {
 
         if(headerCheck){
             req
+            .expect('x-ratelimit-limit', limit)
+            .expect('x-ratelimit-remaining', remaining)
```
nfriedly: Doh. Thank you.

nfriedly:
Perfect. One more request: can you update the readme to show the correct default function for options.handler and give options.headers a more accurate description? Maybe something like 

> headers: Enable header to show request limit, current usage, and retry delay in the event of hitting the limit

nfriedly:
Sorry about all of the nit-picking, thanks for hanging in there with me! This should be released in `v2.9.0` shortly.

PR #52, opened 2017-09-08
Fix clarification comment position

felippenardi:
Fixed comment on the API that seemed to be misplaced.


nfriedly:
Yes, thank you!

PR #53, opened 2017-09-10
Pass options to "handler"

Jokero:
Current implementation:
```js
function (req, res, /*next*/) {
  if (options.headers) {
    res.setHeader('Retry-After', Math.ceil(options.windowMs / 1000));
  }
  res.format({
    html: function(){
      res.status(options.statusCode).end(options.message);
    },
    json: function(){
      res.status(options.statusCode).json({ message: options.message });
    }
  });
}
```
Ok, but what if for example I want to change it a bit:
```js
function (req, res, /*next*/) {
  if (options.headers) {
    res.setHeader('Retry-After', Math.ceil(options.windowMs / 1000));
  }
  res.status(options.statusCode).json({ message: options.message });
}
```
In this case I don't have `options`, so I will not have `options.headers`, `options.windowMs`, etc.
This PR fixes that.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 /node_modules/
 npm-debug.log
+.npmrc
 .idea/
--- a/README.md
+++ b/README.md
@@ -103,9 +103,9 @@ function (/*req, res*/) {
     return false;
 }
 ```
-* **handler**: The function to execute once the max limit is exceeded. It receives the request and the response objects. The "next" param is available if you need to pass to the next middleware. Defaults:
+* **handler**: The function to execute once the max limit is exceeded. It receives the request and the response objects, "next" function and options. The "next" param is available if you need to pass to the next middleware. Defaults:
 ```js
-function (req, res, /*next*/) {
+function (req, res, next, options) {
   if (options.headers) {
     res.setHeader('Retry-After', Math.ceil(options.windowMs / 1000));
   }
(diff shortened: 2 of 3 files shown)


nfriedly:
Hi @Jokero 

Thanks, but I'm hesitant to merge this because  I don't think it's really needed or worth the added complexity (in the sense of breaking from convention - both Express's req/res/next signature convention, and the Node.js callbacks-last convention). 

I know it's not a big change, but I think this library is already too complex as it is, and that it needs to loose a few features. 

For your specific case, though:

If you're defining the function anyways, why have it check `options` at all? Just make it do whatever behavior you need and ignore the `options` object. E.g.

```js
function (req, res, /*next*/) {
  res.setHeader('Retry-After', 5*60);
  res.status(429).json({ message: "Too many requests. Please try again later." });
}
```

Or, if there is a good reason to check `options`, just define the function such that `options` is already in scope. For example:

```js
var options = {
  // ...
};

options.handler = function(req, res, next) {
  if (options.headers) {
    res.setHeader('Retry-After', Math.ceil(options.windowMs / 1000));
  }
  res.status(options.statusCode).json({ message: options.message });
});
```

I know that's slightly more verbose, but I think it's preferable to the cognitive overhead of having an oddball function signature.

The next version of the library probably won't automatically switch formats - it'll just respond with whatever you give it, which I think will solve the root issue here.

nfriedly:
As of v3.0.0, the default handler is much simpler, just 

```js
res.status(options.statusCode).send(options.message);
```

`res.send` will automatically stringify if you pass it an object, or treat it as HTML if you pass it a string.

Issue #32, opened 2016-09-27
Rate limit reset timestamp (X-RateLimit-Reset header)


riggerthegeek:
Would anyone be interested in adding an `X-RateLimit-Reset` header?  I'd quite like this myself, but didn't want to submit the pull request if no one was interested?

It would follow the same convention as GitHub uses, returning a unix timestamp as an integer (in seconds, not ms).

If you're ok with that, I'll submit the PR one I've got it working otherwise will just use for my own purposes

nfriedly:
Yea, we talked about that on https://github.com/nfriedly/express-rate-limit/pull/26 - but the trick is that the different stores handle resetting differently. For the MemoryStore, users are grouped into buckets and the entire bucket is reset at each interval. For the Redis store, though, each user has their own reset timer. 

So, any change would have to take this into account and probably update https://www.npmjs.com/package/rate-limit-redis also.

But, that said, if you're interested in doing the work, go for it. Please include a couple of tests with it, though.

renatoargh:
hey @riggerthegeek any updates on this?

riggerthegeek:
Hi @renatoargh, I've not done anything on this - the urgency for this passed. As @nfriedly said, it would require breaking changes to the interface and clients. It's still on my list of things to look at, but there's plenty of other things above this ATM.

NolanChan:
Hi @riggerthegeek , I wouldn't mind looking at the code that you had which adds the `reset` flag, as I would be interested in having this functionality as well :)

riggerthegeek:
Hi @NolanChan, I've still not done anything on this for the reasons above

nfriedly:
@NolanChan If you want to work on adding it, I listed some ideas on https://github.com/nfriedly/express-rate-limit/issues/46 - it may be possible to add in a non-breaking manner if the lib simply doesn't set the header when rate-limit-redis is used as the store. Not sure I like that idea, though.

The alternative option is to make it a breaking change, which is what I kind of expected to do, but just haven't had the time so far.

NolanChan:
Hey @nfriedly just double checking, does https://github.com/nfriedly/express-rate-limit/commit/e5fea7799e60bf661cfd402f3db04729c550a8b5 introduce the X-RateLimit-Reset functionality with `retry-after`?

nfriedly:
[...]

PR #57, opened 2017-12-04
Adding X-RateLimit-Reset to response header

duxing:
- Added type & value check for windowMs
 - Added X-RateLimit-Reset for response header


nfriedly:
Hey, thanks for this, but it will give incorrect results when used with the Redis store. (See #32.)

I think we can release this if you add an additional condition that the MemoryStore is used. You'll want to update the readme too.

And then we'll need to figure out how to add the equivalent change for the Redis store eventually...

duxing:
good call

PR #61, opened 2017-12-14
Add memcached-store to README.md

linyows:
I created a [memcached-store](https://github.com/linyows/rate-limit-memcached) so please add it to README. Thank you for the express-rate-limit of greate software.


nfriedly:
Hey, thanks! I went ahead and re-organized that section to make it more clear that the memcached-store was meant to be used with express-rate-limit rather than as a competing option.

linyows:
Thanks!

PR #67, opened 2017-12-27
Add X-RateLimit-Reset

gndplayground:
Add X-RateLimit-Reset like GitHub API.

Sometimes we need to display reset time before the user can make the next request after they got busted. Retry-After can't help this.

### DIFF
--- a/lib/express-rate-limit.js
+++ b/lib/express-rate-limit.js
@@ -60,7 +60,7 @@ function RateLimit(options) {
 
         var key = options.keyGenerator(req, res);
 
-        options.store.incr(key, function(err, current) {
+        options.store.incr(key, function(err, current, timeReset) {
             if (err) {
               return next(err);
             }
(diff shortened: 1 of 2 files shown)


danielnitu:
👍 +1 for this. Any reason it has not been implemented so far?

lib/express-rate-limit.js
```diff
@@ -60,20 +60,24 @@ function RateLimit(options) {
...
-              remaining: Math.max(options.max - current, 0)
+              remaining: Math.max(options.max - current, 0),
+              reset: timeReset
             };
 
             if (options.headers) {
               res.setHeader('X-RateLimit-Limit', options.max);
               res.setHeader('X-RateLimit-Remaining', req.rateLimit.remaining);
+              if(req.rateLimit.reset) {
+                res.setHeader('X-RateLimit-Reset', req.rateLimit.reset);
```
nfriedly: Nice job on making this compatible with the existing stores.

lib/memory-store.js
```diff
@@ -9,7 +10,11 @@ function MemoryStore(windowMs) {
           hits[key] = 1;
       }
 
-      cb(null, hits[key]);
+      if (!timeReset[key]) {
+          timeReset[key] = parseInt((new Date().getTime() / 1000).toFixed(0), 10) + (windowMs / 1000);
```
nfriedly: This isn't quite correct. For the MemoryStore, all IPs are reset at once, so the offset should be from the last time `resetAll()` was called, not from the time of the user's first request. 

Because of this, it is unnecessary to track or even calculate the time remaining on a per-IP basis. Just record a single `var nextReset = Math.ceil((Date.now()+windowMs)/1000)`, and then update it each time `resetAll()` is called. When a new hit comes in, the reset time is always `nextReset` no matter when we first saw that IP.

(`Date.now()` is the same as `(new Date()).getTime()` but shorter and more efficient ;)

nfriedly:
I cleaned this up and merged it in. It should be released in v2.13 once CI finishes.

Milad:
Is it included in 3.0.1 too?

nfriedly:
Yes, but 3.x has some breaking changes: https://github.com/nfriedly/express-rate-limit/blob/master/README.md#v3-changes

PR #68, opened 2018-01-13
Added new callback for on violation occurred 

popkutty:
Helps developers to log and add analytics data to find out how many users affected by today and helps them to increase/decrease the limit. I badly need this. Please consider merging this. :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,7 +140,6 @@ function (req, res, options) {
   /* empty */
 }
 ```
-```
 * **onViolation**: Function to callback when request did a violation. Defaults:
 ```js
 function (req, res, options) {
--- a/README.md
+++ b/README.md
@@ -140,6 +140,14 @@ function (req, res, options) {
   /* empty */
 }
 ```
+```
+* **onViolation**: Function to callback when request did a violation. Defaults:
+```js
+function (req, res, options) {
+  /* empty */
+}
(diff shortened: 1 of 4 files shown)


popkutty:
@nfriedly will you merge it? Or I'm gonna get rid of your dependency and create by my self. 👎

nfriedly:
Hey, sorry, I've been traveling and haven't had a free minute. I'm out
without a computer right now, but for a temporary workaround, you can point
your package.json at your git branch.

nfriedly:
Ok, so I looked through your code and I see what you're trying to do now. 

However, I don't think this change is the best way to go about it. Instead of modifying the Store API (which is a breaking change), you can compute the `violation` value from the data that `onLimitReached` receives:

```js

{
  onLimitReached: function(req, res, opts) {
     if ( req.rateLimit.count === (options.max + 1) ) {
       // this is the first time the client has hit the rate limit (during this window)
       // do your onViolation() stuff here
    }
  }
}
```

popkutty:
Thanks @nfriedly So If I use my package I'm not violating any license right? Just want to make sure.

nfriedly:
Correct, no violation in using your own copy (it's MIT licensed - https://opensource.org/licenses/MIT)

PR #72, opened 2018-02-18
Added dynamicFn to options to modify options per req, res, key

ericberberich:
Added a function to dynamically change the options based on req, res, and key

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,6 +140,22 @@ function (req, res, options) {
   /* empty */
 }
 ```
+* **dynamicFn**: (Optional) Function to modify options themselves. Model:
+```js
+function (req, res, key, options) {
+  /* empty */
+}
+```
+Example:
+```js
+function (req, res, key, options) {
+  if (req.user.isAdmin) {
+    options.max = 10;
+  } else {
+    options.max = 1;
+  }
+}
+```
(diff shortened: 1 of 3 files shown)


smchung:
Hi. Is there a reason why this PR wasn't merged?

nfriedly:
Hi @smchung & @ericberberich , sorry this got overlooked. 

Looking at it, I don't think it makes sense to allow the entire configuration to be modified for each request, but I do see that there is a need to have different reate limits for different users (eg, this and #100).

What I'm thinking is that it should accept either a number or a function for the options.max, and then the function option may return either a number or a promise that resolves to a number. Would that work for your needs?

/cc @CheweyZ would that also resolve your issue?

One other relevant question: what versions of node.js are you guys using? Would dropping support for <8.x cause anyone trouble?

CheweyZ:
If max is a function/number that would also work yes

smchung:
We are using node version 8+.

A max function works for us too. Basically we're trying to limit the number of calls based on the api key.

Thanks for looking into this. Appreciate that.

nfriedly:
Support for setting options.max to a function is now released in v3.2.0. 

Please test it out and let me know how it works for you!

smchung:
Thanks @nfriedly Will check that out. Appreciate the quick turnaround 👍

PR #87, opened 2018-08-02
 Added ability to skip successful requests

busha98:
for frequent requests with some of password ( to avoid bruteforce)


nfriedly:
Hey, thanks, this looks like a good compliment to the skipFailedRequests option.

nfriedly:
should be released in v2.12.0 once CI finishes, and I think I fixed the issue that caused it to fail for your PR.

nfriedly:
Sorry, I had to fix two things to make CI pass, but now it's released on npm - v2.12.1

PR #90, opened 2018-08-15
Added getHits to memory-store

busha98:


nfriedly:
Hi, this is interesting, can I ask what the use-case is for it?

Also, I'd like to continue to support Node.js 6 as long as it's still in maintenance mode, so that means no spread operator until next year. However, `Object.assign({}, hits)` would work here.

busha98:
@nfriedly 
Hi.
For example for warnings about approaching blocking and to determine the first request after blocking.

with this also can help:
```
  this.getHit = function(key) {
    return hits[key];
  };
```
But access to this object will give more opportunities

nfriedly:
So, I don't think you need a change to enable either of those use-cases. `express-rate-limit` adds a `res.rateLimit` object with the current hits, remaining hits, and the limit: https://github.com/nfriedly/express-rate-limit/blob/master/lib/express-rate-limit.js#L71

You can then do something with `res.rateLimit.remaining` an maybe alert users if it's under 5 or whatever.

For the first request after blocking, you can check these options inside the  `handler` or `onLimitReached` callback like so:

```js
var limiter = new RateLimit({
  // ...
  onLimitReached: function(req, res, options) {
    if (req.rateLimit.current === req.rateLimit.limit + 1) {
      // this is the first blocked request within the current window
      // in the next major release, the onLimitReached method may change to *only* trigger on the first blocked request within the current window
    }
  }
});
```

(The difference between `handler` and `onLimitReached` is that handler is expected to respond to the message, whereas `onLimitReached` isn't.)

busha98:
@nfriedly 
Thanks a lot, it helped!

Issue #64, opened 2017-12-20
`onLimitReached` is misleading


SimonSchick:
The name suggests it's triggered when the limit is `reached`, not for every requests that is send after the limit is reached.

nfriedly:
Hum, good point. As I mentioned on #63, this lib is due for a major revamping with a number of breaking changes, including something here.

popkutty:
Created PR for this. @nfriedly can you please review and merge it, if you're ok with the changes.
https://github.com/nfriedly/express-rate-limit/pull/68

nfriedly:
@SimonSchick See https://github.com/nfriedly/express-rate-limit/pull/68#issuecomment-358058305 for a way to trigger for just the request that first reaches the limit

nfriedly:
As of v3.0.0 the onLimitReached callback will only be triggered once per window

PR #91, opened 2018-08-19
update: call onLimitReached only once before an IP resets

ashraful-islam:
changelog:
---

+ updated to only call onLimitReached once on exceed before IP reset
+ added test(may need review/corrections)
+ should avoid confusion mentioned here https://github.com/nfriedly/express-rate-limit/issues/64

### DIFF
--- a/test/express-rate-limit-test.js
+++ b/test/express-rate-limit-test.js
@@ -651,4 +651,36 @@ describe("express-rate-limit node module", function() {
     createAppWith(limiter, true, done, done);
     done();
   });
+
+  it("should call the custom onLimitReached only once before an IP resets", function(done) {
+    var onLimitReachedCounter = 0;
+    var noop = function() {};
+
+    var limiter = rateLimit({
+      delayMs: 0,
+      max: 1,
+      onLimitReached: function(req, res) {
+        assert.ok(req);
+        assert.ok(res);
+        // increment call count
+        ++onLimitReachedCounter;
+      }
+    });
+    createAppWith(limiter);
+    goodRequest(done);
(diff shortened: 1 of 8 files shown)


ashraful-islam:
I'd warn that, this will break existing implementation somewhere, where this(`onLimitReached`) is used in the manner it operates(called on each request after exceeding limit), say using it to log some offender or log excess calls/abuse etc.

test/express-rate-limit-test.js
```diff
@@ -651,4 +651,36 @@ describe("express-rate-limit node module", function() {
...
+      .end(noop);
+    request(app) // third call, should not trigger onLimitReached
+      .get("/")
+      .end(function(err) {
+        if (err) {
+          done(err);
+        } else if (onLimitReachedCounter > 1) {
+          done(new Error("onLimitReached called more than once."));
+        } else {
+          done();
```
nfriedly: This will also pass if onLimitReachedCounter is 0. Instead of checking for > 1 and erroring, maybe just assert that the count is 1:

```js
        if (err) {
          return done(err);
        }
        assert.equal(onLimitReachedCounter, 1, 'onLimitREachedCounter should be 1');
        done();
```

Also, I added a `return` in the `if (err)` block so that it doesn't continue executing and cause confusion.

nfriedly:
And, then, yea, this is a breaking change, so I will probably try and bundle a couple of other breaking changes into the v3 release.

nfriedly:
Also, I created a v3 branch and targeted your PR at it.

nfriedly:
This is now merged in and released in v3.0.0

PR #95, opened 2018-08-22
Fix Date response header format

### LINKED ISSUE #94: Date response header is malformed
The `Date` response header added in 598b9b41c47e6530d2020563b6dff9872bb727b6 is not in a valid format, which is causing breakage downstream for anything else that relies on the `Date` value in the response headers (e.g. https://github.com/pquerna/cachecontrol).

The offending code starts here:
https://github.com/nfriedly/express-rate-limit/blob/3b5fe600b3f88add876add420e6898b83dda0f3f/lib/express-rate-limit.js#L76

The `Date` response header should be in the following format, per the spec:

```
Tue, 15 Nov 1994 08:12:31 GMT
```

References:

https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.18
https://tools.ietf.org/html/rfc7231#section-7.1.1.1
https://tools.ietf.org/html/rfc7231#section-7.1.1.2

wKovacs64:
Closes #94.

If this is the appropriate way to resolve, it ought to be back-ported to express-rate-limit v2 as well.

### DIFF
--- a/lib/express-rate-limit.js
+++ b/lib/express-rate-limit.js
@@ -70,7 +70,6 @@ function RateLimit(options) {
         res.setHeader("X-RateLimit-Remaining", req.rateLimit.remaining);
         if (resetTime) {
           // if we have a resetTime, also provide the current date to help avoid issues with incorrect clocks
-          // format from https://stackoverflow.com/a/13219636/933879
           res.setHeader("Date", new Date().toGMTString());
           res.setHeader("X-RateLimit-Reset", resetTime);
         }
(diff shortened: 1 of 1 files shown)


nfriedly:
Arg, thank you. I'll get this released and backported shortly.

wKovacs64:
Ah good catch on the comment. Thanks.

nfriedly:
Ok, released in v3.0.3 and v2.13.1, and I think I corrected the `@latest` dist-tag so that new installs will get 3.x

wKovacs64:
Great, thanks for the quick response!

PR #98, opened 2018-09-06
Fixed express-slow-down url

eiabea:
Very minor change, but confused me while upgrading to version 3


nfriedly:
Should be released by CI in `v3.1.1`

PR #102, opened 2018-10-05
Fixed NaN remaining when max is a function

brinley:
This bug rose from https://github.com/nfriedly/express-rate-limit/releases/tag/v3.2.0 where max can now be a function (which is extremely useful btw so thanks!).

Please merge at your earliest convenience so that we can start using max as a function.
Thanks!

### DIFF
--- a/lib/express-rate-limit.js
+++ b/lib/express-rate-limit.js
@@ -67,7 +67,7 @@ function RateLimit(options) {
           req.rateLimit = {
             limit: max,
             current: current,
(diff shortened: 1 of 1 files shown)


nfriedly:
Hey, thanks and sorry I missed this. On a phone right now, but I'll try and get it released tonight or tomorrow.

nfriedly:
Should be published as v3.2.1 now. I also added a test to prevent it from breaking in the future.

PR #115, opened 2018-11-19
Make key generator a promise so you can await it

rushi:


rushi:
Accidentally created this PR, ignore me

PR #116, opened 2018-11-22
packages are updated and README.md is updated

palashg7563:


nfriedly:
Hey, thanks for trying to help out here, but I'd rather stick with package-lock.json for now, as I usually use npm.

Also, do you know why github is showing every single line of the readme as being changed? (whitespace-only, but still confusing)

Martii:
@nfriedly 

> Also, do you know why github is showing every single line of the readme as being changed? (whitespace-only, but still confusing)

Usually lock files can completely change depending on how they were created. As for the README.md depends on the line endings. \*Nix usually uses a single character and Windows uses two characters. Probably just a miss on code conformity compliance. e.g. line endings. Could be anywhere from the checkout to the editor to the push. Good catch to keep it. We use an editor subsystem called https://editorconfig.org/ to ensure that most of the time things are the way they should be. There are others too.

> ... still confusing)

https://github.com/nfriedly/express-rate-limit/pull/116/files?utf8=%E2%9C%93&diff=unified&w=1

May reveal more of what is going on. Deletion of the the package-lock.json probably in favor of yarn.lock, Some dependency updates, and some filling out of an example is what was done in this PR.

Not everyone commits the package-lock.json. It's a mixed bag that I won't get into here much. We don't do it since it can cause what it supposed to prevent.

\/2 pence

P.S. Nice project btw. :smile:

nfriedly:
thanks :)

PR #119, opened 2018-12-23
first draft for the types for express rate limiter

andreGarvin:
I think this needs to be improved more, if anyone wants to review the PR and check it out their local branch that would be great

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -5,6 +5,7 @@ export interface Options {
     message?: string;
     headers?: boolean;
     windowMs?: number;
+    store: Store | any;
     statusCode?: number;
     skipFailedRequests?: boolean;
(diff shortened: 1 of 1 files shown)


nfriedly:
This looks good to me. Is there an automated way to ensure it stays in sync with the actual JavaScript files?

santiq:
@nfriedly 
> This looks good to me. Is there an automated way to ensure it stays in sync with the actual JavaScript files?
I think there is no way to do that if the source code is not written in typescript.

Regardless that minor issue, I think this it's a nice feature for those using typescript like me, would you considerer on merging this?

nfriedly:
Fair enough, I'll go ahead and get it merged in and released shortly

santiq:
@nfriedly Thank you so much! 😄

PR #121, opened 2019-01-17
feat: setHeader function

ekoeryanto:
It allows me to use this plugin with fastify fastify/fastify-rate-limit#10

### DIFF
--- a/lib/express-rate-limit.js
+++ b/lib/express-rate-limit.js
@@ -15,6 +15,9 @@ function RateLimit(options) {
     keyGenerator: function(req /*, res*/) {
       return req.ip;
     },
+    setHeader: function(req, res) {
+      return res.setHeader
+    },
     skip: function(/*req, res*/) {
       return false;
     },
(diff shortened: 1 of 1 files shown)


nfriedly:
Hi,

I'm not quite sure I understand the goal here, does express-rate-limit not work with fastify right now?

And then I also see a couple of issues with the implementation:

* the big one is that you defined `options.setHeaders` as expecting `req, res`, but then you're calling it with two strings (?)
* also, it's failing the linter; `npx eslint . --fix` should be able to fix that automatically

ekoeryanto:
Sorry for that, I edited directly on github web ui. I'll be back when I am not AFK

nfriedly:
FYI, res.setHeader is a core Node.js thing, not something Express ads: https://nodejs.org/api/http.html#http_response_setheader_name_value ;)

nfriedly:
I'm going to go ahead and close this for now. Feel free to re-open if appropriate.

PR #133, opened 2019-05-09
Fix type of message to match Request.send

googol:
The `send` function on Express.Request is typed as accepting `any` value. Since the documentation states the `message` option will be directly passed to `send`, it makes sense to have the same type here.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -2,7 +2,7 @@ import { RequestHandler, Request, Response, NextFunction } from 'express';
 
 export interface Options {
     max?: number;
-    message?: string;
+    message?: any;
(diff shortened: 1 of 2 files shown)


nfriedly:
Ah, good call. I'm on my phone right now, but I'll get this merged in and release later on today.

googol:
I added a second commit that actually makes the type definitions available to the typescript compiler via the "types" key in package.json

nfriedly:
Sorry, yesterday ended up kind of hectic, but it should be published to npm in v3.5.1 as soon as the CI loop finishes.

googol:
Thanks!

PR #135, opened 2019-05-21
Include typescript typings in npm package

googol:
Just one more thing in the way of the typings, they aren't getting included in the package because of the files prop in package.json. This fixes that


nfriedly:
Good call, I'll get this published shortly.

PR #137, opened 2019-05-22
Makes Options.store a optional member. Fixes "Property 'store' is mis…

jason-han-nue:
…sing" error in typescript


nfriedly:
Awesome, thanks.

nfriedly:
This should be published in v4.0.1 by CI shortly. I also backported it to the 3.x branch and published that manually with a custom tag so that so that it wouldn't replace 4.x as the latest release.

Issue #138, opened 2019-05-22
TypeScript typings and tests


nfriedly:
Currently, the typescript typings are untested, which allows releases to go out with mistakes. To remedy this, either the current tests could be converted to typescript, or new tests could be written in typescript that just validate the interface.

I'm not looking to switch the library itself to typescript at this point, although I'm not against that in principle.

This relates to #65 / #119 and #136 / #137 

Update: the typescript typings were removed from v5.0.0 and moved to the [typescript](https://github.com/nfriedly/express-rate-limit/tree/typescript) branch. I'll merge them back into master and publish a new release once they're tested and fixed.

Sieabah:
Just got bit by this today, had it working on 3.4.1 but upgraded to 4.0.1 today.

`Type 'typeof import("node_modules/express-rate-limit/index")' has no compatible call signatures.`

In the meantime I've just changed my calls to `(rateLimit as any)({ ..config object.. })`

nfriedly:
Ouch, I'm sorry. Is that something that you know how to fix? I've basically never used type script, but I know a lot of people that use this library use it, so I do want it to be supported better.

Sieabah:
It's not big deal, I don't write many definition files so I'm not familiar with how to structure it. Looking at the history this error shouldn't have happened in the first place so it's probably just something up with my configuration.

RoniFinTech:
Hi, anybody found a solution?
getting this error with 4.0.1

derSoerrn95:
you have to do something like this:

`import {RateLimit} from 'express-rate-limit';`

jecsham:
Version:
4.0.2
trying running `tsc` command
```
node_modules/express-rate-limit/index.d.ts(32,1): error TS2309: An export assignment cannot be used in a module with other exported elements.
src/index.ts(17,20): error TS7009: 'new' expression, whose target lacks a construct signature, implicitly has an 'any' type.
```

ivan-rangel:
Agree with @derSoerrn95, it worked but kind of dirty solution

nfriedly:
Sorry guys. I 've never really done anything with TypeScript. I merged in the initial PR to add typings without realizing there'd be this much trouble. Now, 6 PRs later, folks are still having issues. I knew I didn't know how to use TS properly, but it seems like no one else does either :(

At this point, I'm thinking I might just remove the `index.d.ts` file, it doesn't seem to be worth the trouble.

kevinkrug:
[...]

PR #139, opened 2019-05-27
Update index.d.ts

Diluka:
According to the source code, typings shall be like this.
https://github.com/nfriedly/express-rate-limit/blob/5e123ad26ee6aa2447b11eaef370031e95980b3a/lib/express-rate-limit.js#L4
https://github.com/nfriedly/express-rate-limit/blob/5e123ad26ee6aa2447b11eaef370031e95980b3a/lib/express-rate-limit.js#L147

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -16,7 +16,6 @@ export interface Options {
 }
 
 export interface Store {
-    constructor(windowMs: number);
     hits: {
         [key: string]: number;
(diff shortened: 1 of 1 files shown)


nfriedly:
Thank you, I'll get it released shortly. 

Do you think you can get me a test or two that just validates the basic functionality? #138

PR #141, opened 2019-05-29
Update index.d.ts

derSoerrn95:
fixed ts 3.5.x error (only testet with ts 3.5.1)


nfriedly:
Looks good to me, thanks!

PR #143, opened 2019-05-30
Ensure promise execs once

tjk:


tjk:
Wow, sorry -- pressed the wrong button.

Issue #142, opened 2019-05-30
TypeScript TS7009 & TS2309 Error


jecsham:
Version:
4.0.2
Running `tsc` command
```
node_modules/express-rate-limit/index.d.ts(32,1): error TS2309: An export assignment cannot be used in a module with other exported elements.
src/index.ts(17,20): error TS7009: 'new' expression, whose target lacks a construct signature, implicitly has an 'any' type.
```

nfriedly:
Please update to 4.0.3, and let me know if it fixed that issue.

sliktrik:
can confirm the issue still exists on 4.0.3

jecsham:
The issue still.

You are not exporting RateLimit function by default so we can import to our project with
`import { RateLimit } from "express-rate-limit";`

and delete the `new` when we want instace cause `RateLimit` is not a class with constructor.

PR #145, opened 2019-06-01
Fix bug with TypeScript support

AnandChowdhary:
We the function RateLimit as the default export


AnandChowdhary:
This should also fix #142.

nfriedly:
Hey, thanks! Is there any chance you could also write some tests to keep this working in the future?

PR #146, opened 2019-06-06
:bug: fixes Ratelimit interface to use valid typescript

kevinkrug:
just updates the definitions file with a typescript syntax fix..

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -36,4 +36,8 @@ export interface Store {
   incr(key: string | any, cb: (err?: Error, hits?: number) => void): void;
 }
 
-export default function RateLimit(options?: Options): (req: Request, res: Response, next: NextFunction) => void;
+interface RateLimit {
+  (options?: Options): (req: Request, res: Response, next: NextFunction) => void;
+}
+
+export default RateLimit;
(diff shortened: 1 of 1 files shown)


kevinkrug:
@nfriedly sorry I just realized my PR is incorrect, I'm just exporting the type as default, but what I actually what is to declare the rateLimit type.. (```declare const rateLimit: RateLimit;`) fix coming up!

nfriedly:
Can you please add a test also? Just something basic that loads the lib and maybe creates an instance. But, obviously, written in typescript so that it validates the typings.

I'm not saying that your PR is wrong, I really don't know enough to say either way. But the previous 6 apparently were and I probably should have required a test before merging any of them. So, at this point, I'm not going to merge anything new for typescript without a passing test.

Let me know if you need help. I'm good with the infrastructure/automation side of things, I've just never worked with typescript very much.

nfriedly:
Ok, I re-targeted this at the typescript branch, and I'll go ahead and merge it in there. 

The npm package is now on v5 and no longer includes any typings. I'll merge the typescript branch into master and publish it to npm once it's tested and I'm confident it's not going to break things for people.

PR #152, opened 2019-07-07
Convert to typescript

googol:
So since there's been some talk about how to test if the typescript typings are valid, I thought the simplest way is to generate them from the actual code. Since the codebase is so small, I just converted it all into typescript. I don't know if you like this idea at all, but here it is as a proposal. If this is not where you want to take the project, this can at least work as a reference for the typings


Martii:
Impresssive but a few questions come to mind:

1. What sort of change in timings *(execution/build/distributing)* will this have compared to the building routine vs. static distribution?
2. Will this affect static expresses routing at all *(and variants of routing)*?
3. Will this affect statically linking  to GH in case npmjs.com is down/change/other? e.g. will we need to add an additional build step?
4. Will the README.md need to be updated to reflect any/all possible changes?
5. How will this affect all of the other stores? e.g. What about incompatibilities there?
6. With the general browser migration towards Chromium based back-ends... how fluid does it really need to be?

Basically I come from a history of CPP which is very strict on typing *(type casting for example when needed)* and JavaScript was quite a change for myself. However the benefits of being able to keep certain elements "vague" on their type have been a perk over the last decade or so. Our project currently expects a potential developer not to rely soley on a single IDE *(these environments change all the time over the decades)*. Quite honestly I prefer a plain old text editor and learning from any possible mistakes. Sure it's nice to have it pointed out during typing on the keyboard but nothing will catch everything over the decades of experience and team projects I've worked in the private and public sectors. Having CI detect errors can be good however it's not always correct either.

Anyhow... those are a few questions off the top of my head on taking any JavaScript Code out of vanilla JavaScript into another language *(Superset or not)*. I have observed several "debates" around CoffeeScript over the years which, from what I can tell, seems to do nearly or similar things as TypeScript.

googol:
Good points!

For a javascript consumer of the npm package, this change shouldn't introduce any changes. It doesn't include any logic level changes. 

As to your specific bullet points:
1. For a consumer of the package, there is no change in how it is used. No added build steps or such
2. This won't affect express at all. The exported value is still the same kind of middleware as before
3. With the current setup in this branch, installing directly from github won't work, as the repo doesn't contain the built code, but only the sources. It is definitely possible to include the built code in the repo as well if required, it just requires a bit more complexity in the package publishing process.
4. Definitely, the readme should mention how to build etc
5. This doesn't affect any other store. This middleware still expects the same kind of object as the store as before. If a store implementer wants to implement their store in typescript, they can use the `Store` interface from this branch to type their store.

I don't quite understand what you mean with your sixth point, can you explain a bit more?

So anyways, if you want to just use javascript, this won't affect you at all.

Here's the build output in a gist, if you want to compare it to the current files: https://gist.github.com/googol/d21c6b341ca3a47384d0848f17bec260

Martii:
Appreciate you taking the time to lay it out for everyone including myself.

> ... your sixth point, can you explain a bit more?

ECMAScript *(ES6+)* may change syntax but with Edge/Chrome/Fx migrating towards Chromium backends all the browsers will eventually use nearly the same JavaScript parser to CPP code. The usual discussion on forced updates comes into play as well which I'd rather avoid discussing here as it's too volatile. 

> ... With the current setup in this branch, installing directly from github won't work

This could be a potential :-1: . I've looked into build routines and there is basically none that I really like since I'm the Active Maintainer. I am however always scouring for new ones or different feature requests with existing ones... all on the lengthy TODO list. Currently we just do things *(optionally a build)* as install scripts with `npm` that we write instead of the additional abstraction of a build language parser. From what I've skimmed over TypeScript seems to be related to Microsofts IDE but since I'm a noobie there *(technically out of date)* it raises an eyebrow when our system is mostly text based and usually run under a Linux server. Transpiling from .ts to .js would definitely be a plus in the README.md under all platforms *(Linux, macOS, Windows, other)*.

One of our dependencies claimed it was renamed when it was actually a new nmpjs.com project. We would like to avoid these types of situations. That particular dep update is stalled for more critical matters so we pull directly from a commit reference on GH to maintain current usage. Obviously this might break at some point but the more time to migrate the better imho. :) In my little corner of the world wide web we would hope that this would be marked as possibly breaking with semver if the process goes forward due to the nature of eliminating a source pull with `npm`. We also use quite a bit of static routing to inject into the DOM. This project shouldn't be an issue with this part I think since it's usually handled in *node* back-end *(server side)*... but there can be DOM usage with some dependencies.

As @nfriedly stated [over there](https://github.com/nfriedly/express-rate-limit/issues/138#issue-447172005):
> I'm not looking to switch the library itself to typescript at this point, although I'm not against that in principle.

... We're not against the principle either but do appreciate the recognition of the queries. This is one of our most critical dependencies to keep the site going since we're attacked almost constantly *(we know where in the world they are although as to the why is just a shrug)*.

tsconfig.json
```diff
@@ -0,0 +1,64 @@
...
+    // "incremental": true,                   /* Enable incremental compilation */
+    "target": "es5" /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019' or 'ESNEXT'. */,
+    "module": "commonjs" /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', or 'ESNext'. */,
+    // "lib": [],                             /* Specify library files to be included in the compilation. */
+    // "allowJs": true,                       /* Allow javascript files to be compiled. */
+    // "checkJs": true,                       /* Report errors in .js files. */
+    // "jsx": "preserve",                     /* Specify JSX code generation: 'preserve', 'react-native', or 'react'. */
+    "declaration": true /* Generates corresponding '.d.ts' file. */,
+    // "declarationMap": true,                /* Generates a sourcemap for each corresponding '.d.ts' file. */
+    "sourceMap": true /* Generates corresponding '.map' file. */,
```
nfriedly: does this require the `/lib/` folder to be useful?

googol:
The `lib` directory is not needed in the npm package since the actual javascript is now in the `dist` directory. `dist` contains the javascript that is imported by consumers, the typings and the sourcemaps

nfriedly:
Right, but there's two ways to make a sourceMap - either inline the original sources, or else point to them. I think the current config points to them, but then doesn't include them in the published package, which would make the sourcemap kind of useless. Let me test it, though, because I could be wrong...

nfriedly:
I changed the target to the typescript branch; that caused a conflict in index.d.ts, sorry about that. It should be straightforward to fix, though.

I also published this to npm under the tag 'typescript'. Folks can install it for testing like so

`npm i express-rate-limit@typescript`

Can you guys test that out and confirm that it works as expected? I'll leave a comment on #138 also.

As far as this earlier comment of mine:
> I'm not looking to switch the library itself to typescript at this point, although I'm not against that in principle.

That was largely because I believed the tests-only route would be less work, although given that this is now done, I'm leaning the other way now. 

I'll probably make this a semver major release anyways, just in case.

tsconfig.json
```diff
@@ -0,0 +1,64 @@
...
+    // "allowSyntheticDefaultImports": true,  /* Allow default imports from modules with no default export. This does not affect code emit, just typechecking. */
+    "esModuleInterop": false /* Enables emit interoperability between CommonJS and ES Modules via creation of namespace objects for all imports. Implies 'allowSyntheticDefaultImports'. */
+    // "preserveSymlinks": true,              /* Do not resolve the real path of symlinks. */
+    // "allowUmdGlobalAccess": true,          /* Allow accessing UMD globals from modules. */
+
+    /* Source Map Options */
+    // "sourceRoot": "",                      /* Specify the location where debugger should locate TypeScript files instead of source locations. */
+    // "mapRoot": "",                         /* Specify the location where debugger should locate map files instead of generated locations. */
+    // "inlineSourceMap": true,               /* Emit a single file with source maps instead of having a separate file. */
+    // "inlineSources": true,                 /* Emit the source alongside the sourcemaps within a single file; requires '--inlineSourceMap' or '--sourceMap' to be set. */
```
nfriedly: I tested my earlier question about sourcemaps and confirmed that they didn't actually work unless you were working from a local copy of the code with `npm link` - when installed from npm, the source was omitted, so the sourcemap was useless. However, turning this option on causes it to inline the source into the sourcemap. So, when installing from npm and stepping through a debugger, you see the original TypeScript source, not the transpiled Javascript. It's published in `express-rate-limit@0.0.0-typescript-beta-4`

.travis.yml
```diff
@@ -17,12 +17,12 @@ jobs:
   include:
     - stage: deploy
       node_js: stable
-      script: skip
+      script: npm run build
```
nfriedly: I reverted this and created a `scripts.prepublish` in the package.json with `npm run build` there. That way a manual `npm publish` also compiles before publishing.

PR #157, opened 2019-07-12
Typescript improvements

googol:
A couple of improvements on top of the typescript conversion.

- Since you noted that the stores do not have to return the resetTime, I changed the type on the increment callback as well
- Made all of the properties on the interfaces read only, since they are handled as such anyway
- Created a helper function for getting the current request's rate limit, since it's always a bit difficult to access those added request properties in ts
- Added a bit of extra support for tests written in typescript

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -4,7 +4,10 @@
 # top-most EditorConfig file
 root = true
 
-[*.{js,json}]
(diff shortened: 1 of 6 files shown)


nfriedly:
Awesome, thank you!

PR #161, opened 2019-09-12
Detect headers sent before sending headers. 

michalzubkowicz:
fixes Error: Can't set headers after they are sent when exception inside controller method is thrown

### DIFF
--- a/lib/express-rate-limit.js
+++ b/lib/express-rate-limit.js
@@ -73,7 +73,7 @@ function RateLimit(options) {
             resetTime: resetTime
           };
 
-          if (options.headers) {
+          if (options.headers && !res.headersSent) {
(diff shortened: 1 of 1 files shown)


michalzubkowicz:
You are right, it is not very usable after header have been sent, but it
should support situation when some controller is throwing an error. For
example you can loose control of rate limiting when some backend system is
not responsive.
This error is not explaining clearly what is happening and it's better to
have it hidden, because it can lead someone to thinking that something is
wrong.

nfriedly:
Yea, that's fair, a "headers already sent" error can be confusing. I think it should throw a more clear error, rather than silently doing nothing, though. Otherwise it'd be too easy to get it wrong from the start and never know it was mis-configured.

Issue #168, opened 2020-01-17
Support for the new RateLimit standardization proposal


ioggstream:
## I wish

this plugin to support (eg. eventually via configuration) the [new ratelimit standardization proposal](https://tools.ietf.org/id/draft-polli-ratelimit-headers-01.html).

The proposal was:
-  recently presented at the [HTTP Workgroup during the IETF Meeting](http://bit.ly/2qWhwwD)
- implemented by Kong
- under implementation in 3scale

## The proposal

Very similar to your current implementation, but  header names have no `X-` and `Reset` is in `delta-seconds` instead of timestamp. The rationale for the choice is in [FAQ 5](https://ioggstream.github.io/draft-polli-ratelimit-headers/draft-polli-ratelimit-headers.html#rfc.appendix.D).

```
RateLimit-Limit: 60
RateLimit-Remaining: 50
RateLimit-Reset: 5
```

## Proposed change

The line to be changed (even via a configuration parameter) is here:
https://github.com/nfriedly/express-rate-limit/blob/f56434f417180f1c4dbaf2f04a0c150877adce8e/lib/express-rate-limit.js#L83

## Notes

Your feedback on the ratelimit spec is very welcome!

nfriedly:
Hey, that's interesting. I think we can definitely support that with a configuration flag right now. Do you want to put together a pull request with that?

Assuming this gets standardized, I think the next semver major release of rate limit should default to the standard version and perhaps offer a config flag to go back to the current behavior. (Maybe not...)

What is the process of going from a proposal to an official standard?

ioggstream:
I'll work on a PR. Where can I add the configuration parameter?

The standardisation work is in the preliminary phase and we are now gathering implementations.

During the July IETF meeting we'll present the implementors and move forward the spec. 

We can opt for an experimental rfc - with a faster track-or a standard one - slower

ioggstream:
Hi @XVincentX
 seems you're using `express-rate-limit` here: https://github.com/ExpressGateway/express-gateway/blob/8d4e9646e795538ee2abf272b190cabff0ae7568/lib/policies/rate-limit/rate-limit.js 

When this issue will be fixed can we say that `express-gateway` too supports the I-D or we need some more changes there?

XVincentX:
I think I am going to need to update the dependency but other than that, we should be good.

nfriedly:
@ioggstream Apologies, I was out for a few days.

[...]

PR #169, opened 2020-01-21
Config for RateLimit standardization headers 

sanyamdogra:
For #168 

* Added new option `draft_polli_ratelimit_headers` set to `false`.
* Added the set of new headers `RateLimit-Limit`, `RateLimit-Remaining` and `RateLimit-Reset`.
* Added *draft_polli_ratelimit_headers* in Docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,7 +141,9 @@ Defaults to `true`.
 
 ### draft_polli_ratelimit_headers
 
-This plugin is to support (eg. eventually via configuration) the new (ratelimit standardization proposal)[https://tools.ietf.org/id/draft-polli-ratelimit-headers-01.html].
+Enable headers conforming to the [ratelimit standardization proposal](https://tools.ietf.org/id/draft-polli-ratelimit-headers-01.html): `RateLimit-Limit`, `RateLimit-Remaining`, and, if the store supports it, `RateLimit-Reset`. May be used in conjunction with, or instead of the `headers` option.
+
+Defaults to `false`, but will likely be changed in the next major release.
 
 ### keyGenerator
 
--- a/README.md
+++ b/README.md
@@ -139,6 +139,10 @@ Enable headers for request limit (`X-RateLimit-Limit`) and current usage (`X-Rat
 
 Defaults to `true`.
 
+### draft_polli_ratelimit_headers
+
+This plugin is to support (eg. eventually via configuration) the new (ratelimit standardization proposal)[https://tools.ietf.org/id/draft-polli-ratelimit-headers-01.html].
+
 ### keyGenerator
 
 Function used to generate keys.
@@ -222,7 +226,6 @@ Available data stores are:
 - [rate-limit-redis](https://npmjs.com/package/rate-limit-redis): A [Redis](http://redis.io/)-backed store, more suitable for large or demanding deployments.
 - [rate-limit-memcached](https://npmjs.org/package/rate-limit-memcached): A [Memcached](https://memcached.org/)-backed store.
 - [rate-limit-mongo](https://www.npmjs.com/package/rate-limit-mongo): A [MongoDB](https://www.mongodb.com/)-backed store.
-
 
 You may also create your own store. It must implement the following in order to function:
 
--- a/lib/express-rate-limit.js
+++ b/lib/express-rate-limit.js
@@ -71,8 +71,7 @@ function RateLimit(options) {
             limit: max,
             current: current,
             remaining: Math.max(max - current, 0),
-            resetTime: resetTime,
-            deltaSeconds: Math.ceil((resetTime.getTime() - Date.now()) / 1000)
+            resetTime: resetTime
           };
 
           if (options.headers && !res.headersSent) {
@@ -91,10 +90,10 @@ function RateLimit(options) {
             res.setHeader("RateLimit-Limit", max);
             res.setHeader("RateLimit-Remaining", req.rateLimit.remaining);
             if (resetTime) {
-              res.setHeader(
-                "RateLimit-Reset",
-                Math.max(0, req.rateLimit.deltaSeconds)
+              const deltaSeconds = Math.ceil(
+                (resetTime.getTime() - Date.now()) / 1000
               );
+              res.setHeader("RateLimit-Reset", Math.max(0, deltaSeconds));
             }
           }
 
--- a/lib/express-rate-limit.js
+++ b/lib/express-rate-limit.js
@@ -71,7 +71,8 @@ function RateLimit(options) {
             limit: max,
             current: current,
             remaining: Math.max(max - current, 0),
-            resetTime: resetTime
+            resetTime: resetTime,
+            deltaSeconds: Math.ceil((resetTime.getTime() - Date.now()) / 1000)
           };
 
           if (options.headers && !res.headersSent) {
@@ -86,15 +87,13 @@ function RateLimit(options) {
               );
             }
           }
-          if (options.draft_polli_ratelimit_headers) {
+          if (options.draft_polli_ratelimit_headers && !res.headersSent) {
             res.setHeader("RateLimit-Limit", max);
             res.setHeader("RateLimit-Remaining", req.rateLimit.remaining);
-            if (resetTime instanceof Date) {
-              // if we have a resetTime, also provide the current date to help avoid issues with incorrect clocks
-              res.setHeader("Date", new Date().toGMTString());
+            if (resetTime) {
               res.setHeader(
                 "RateLimit-Reset",
-                Math.ceil((resetTime.getTime() - Date.now()) / 1000)
+                Math.max(0, req.rateLimit.deltaSeconds)
               );
             }
           }
--- a/lib/express-rate-limit.js
+++ b/lib/express-rate-limit.js
@@ -9,6 +9,7 @@ function RateLimit(options) {
       message: "Too many requests, please try again later.",
       statusCode: 429, // 429 status = Too Many Requests (RFC 6585)
       headers: true, //Send custom rate limit header with limit and remaining
+      draft_polli_ratelimit_headers: false, //Support for the new RateLimit standardization headers
       skipFailedRequests: false, // Do not count failed requests (status >= 400)
       skipSuccessfulRequests: false, // Do not count successful requests (status < 400)
       // allows to create custom keys (by default user IP is used)
@@ -82,6 +83,18 @@ function RateLimit(options) {
               res.setHeader(
                 "X-RateLimit-Reset",
                 Math.ceil(resetTime.getTime() / 1000)
+              );
+            }
+          }
+          if (options.draft_polli_ratelimit_headers) {
+            res.setHeader("RateLimit-Limit", max);
+            res.setHeader("RateLimit-Remaining", req.rateLimit.remaining);
+            if (resetTime instanceof Date) {
+              // if we have a resetTime, also provide the current date to help avoid issues with incorrect clocks
+              res.setHeader("Date", new Date().toGMTString());
+              res.setHeader(
+                "RateLimit-Reset",
+                Math.ceil((resetTime.getTime() - Date.now()) / 1000)
               );
             }
           }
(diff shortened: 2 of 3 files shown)


lib/express-rate-limit.js
```diff
@@ -85,6 +86,18 @@ function RateLimit(options) {
               );
             }
           }
+          if (options.draft_polli_ratelimit_headers) {
+            res.setHeader("RateLimit-Limit", max);
+            res.setHeader("RateLimit-Remaining", req.rateLimit.remaining);
+            if (resetTime instanceof Date) {
+              // if we have a resetTime, also provide the current date to help avoid issues with incorrect clocks
+              res.setHeader("Date", new Date().toGMTString());
```
ioggstream: I am not sure we have to provide Date, as it may be set by the application instead.

Date (and other timestamp headers) should be used only in conjunction with NTP or similar mechanism.

lib/express-rate-limit.js
```diff
@@ -85,6 +86,18 @@ function RateLimit(options) {
...
+              res.setHeader(
+                "RateLimit-Reset",
+                Math.ceil((resetTime.getTime() - Date.now()) / 1000)
```
ioggstream: I'd ensure that returns a non-negative number.

lib/express-rate-limit.js
```diff
@@ -85,6 +86,18 @@ function RateLimit(options) {
               );
             }
           }
+          if (options.draft_polli_ratelimit_headers) {
```
nfriedly: add a ` && !res.headersSent` check here to avoid errors if other middleware does something unexpected.

lib/express-rate-limit.js
nfriedly: Yea, we don't need the Date header here. The old "X-RateLimit-Reset" was set to a timestamp, so the date was provided so that endpoints could calculate the offset even if the server or client (or both) had an incorrectly set clock.

With the new "RateLimit-Reset" header, we don't need that because it's already an offset.

README.md
```diff
@@ -139,6 +139,10 @@ Enable headers for request limit (`X-RateLimit-Limit`) and current usage (`X-Rat
 
 Defaults to `true`.
 
+### draft_polli_ratelimit_headers
+
+This plugin is to support (eg. eventually via configuration) the new (ratelimit standardization proposal)[https://tools.ietf.org/id/draft-polli-ratelimit-headers-01.html].
```
nfriedly: Lets reword this to something like

> Enable headers conforming to the (ratelimit standardization proposal)[https://tools.ietf.org/id/draft-polli-ratelimit-headers-01.html]: 'RateLimit-Limit`, `RateLimit-Remaining`, and, if the store supports it, `RateLimit-Reset`. May be used in conjunction with, or instead of the `headers` option.

That also prompts the question: what should we do if the store doesn't tell us when the rate limit will reset? I'll bring that up in the original ticket.

ioggstream: The spec [does not  mandate](https://ioggstream.github.io/draft-polli-ratelimit-headers/draft-polli-ratelimit-headers.html#providing-ratelimit-headers) the usage of all headers.

```
A server MAY use one or more RateLimit response header fields defined in this document to communicate its quota policies.
```

>  what should we do if the store doesn't tell us when the rate limit will reset? 

[here](https://ioggstream.github.io/draft-polli-ratelimit-headers/draft-polli-ratelimit-headers.html#ratelimit-limit-header)

```
If the expiring-limit is not associated to a time-window, the time-window MUST either be:

- inferred by the value of RateLimit-Reset at the moment of the reset, or
- communicated out-of-band (eg. in the documentation). <--- this is the Memcache
```

lib/express-rate-limit.js
ioggstream: @nfriedly thanks for all the support you're providing for this PR! 

An advice: the requirement for a middleware to add a  `Date`  header are [here](https://tools.ietf.org/html/rfc7231#section-7.1.1.2)

```
   A recipient *with a clock* that receives a response message without a
   Date header field MUST *record the time it was received* and append a
   corresponding Date header field to the message's header section if it
   is cached or forwarded downstream.
```

lib/express-rate-limit.js
sanyamdogra: Sure, Math.abs() would return the absolute value. I'll do that.

lib/express-rate-limit.js
sanyamdogra: Okay sure!

README.md
nfriedly: Cool, I put a note on the original ticket.

Also, @sanyamdogra I tweaked the proposed text in my earlier comment a bit, and made it copy-past-able.

lib/express-rate-limit.js
ioggstream: I think something like 

```
resetTime.getTime() - Date.now() < 0 ? 0 : Math.ceil(resetTime.getTime() - Date.now()/1000) 
```

nfriedly: That almost works, but it lost the parens around `resetTime.getTime() - Date.now()`. This gets the correct result with a little less code:

```js
Math.max(0, Math.ceil((resetTime.getTime() - Date.now())/1000));
```

or, to improve readability:

```js
var deltaSeconds =  Math.ceil((resetTime.getTime() - Date.now()) / 1000);
res.setHeader("RateLimit-Reset", Math.max(0, deltaSeconds));
```

nfriedly: Also, come to think of it, we should add some tests on this stuff...

lib/express-rate-limit.js
nfriedly: That makes it sound like the server should handle the date header, not the middleware.

And, some quick testing shows that the node.js http module includes a date header by default, so I should probably remove the existing code that sets the date header also...

lib/express-rate-limit.js
ioggstream: +1 for the second;
+1 for tests.

lib/express-rate-limit.js
sanyamdogra: I'll remove from this then.

lib/express-rate-limit.js
sanyamdogra: I thought of `deltaSeconds` too, so will do that then. Let me do these changes then we can move to tests.

README.md
sanyamdogra: Alright sure!

sanyamdogra:
Hi @nfriedly @ioggstream, any leads for the tests?

lib/express-rate-limit.js
```diff
@@ -71,7 +71,8 @@ function RateLimit(options) {
             limit: max,
             current: current,
             remaining: Math.max(max - current, 0),
-            resetTime: resetTime
+            resetTime: resetTime,
+            deltaSeconds: Math.ceil((resetTime.getTime() - Date.now()) / 1000)
```
nfriedly: `resetTime` is not guaranteed to exist, so you need to check for it first

nfriedly:
@sanyamdogra there's some existing tests in https://github.com/nfriedly/express-rate-limit/blob/master/test/express-rate-limit-test.js although they're kind of messy. There are a number of `it(...)` tests that call `goodRequest()` and `badRequest()`, and those methods can check the headers and other things. You could add an additional check for the new headers there, but given how messy they already I might recommend starting over with new methods and just using the existing ones as a template.

lib/express-rate-limit.js
sanyamdogra: Oops! Yes

sanyamdogra:
@nfriedly I think the new headers have been embedded in `express-rate-limit`. Now the only thing remaining is the tests. Let me get an Idea of how does the testing work in this package. Thanks!

sanyamdogra:
@nfriedly I got an idea, I'll have to add `it( check for new headers)`. How should the function `goodRequest( )` be amended to check the new headers?

test/express-rate-limit-test.js
```diff
@@ -307,6 +307,28 @@ describe("express-rate-limit node module", function() {
     );
   });
 
+  it("should send correct ratelimit-limit and ratelimit-remaining", function(done) {
```
nfriedly: This is a good start, but it isn't actually checking the new headers, it's just checking the delay length and the old x-ratelimit-* headers (which are on by default) in the goodRequest method. 

Instead of going that route, lets skip the goodRequest method and just create our own request and do our own checks. In that case, the body would look something more like this (I haven't tested this, but it should be about right):

```js
    const limit = 5;
    const windowMs = 60 * 1000; // 60 * 1000 = 1 minute
    const app = createAppWith(
      rateLimit({ windowMs: windowMs, limit: limit, draft_polli_ratelimit_headers: true })
    );
    const expectedRemaining = 4;
    const expectedResetSeconds = 60;
    request(app).get("/");
        .expect("ratelimit-limit", limit.toString())
        .expect("ratelimit-remaining", expectedRemaining.toString())
        .expect("ratelimit-reset", expectedResetSeconds.toString())
        .expect(200, /response!/)
        .end(done);
  }
```

nfriedly: Oh, I missed your question asking about amending `goodRequest`. Sorry. I think this is probably a better route anyways.

sanyamdogra: Alright, on it then!

ioggstream:
@sanyamdogra Can you provide a docker-compose.yml with express-gateway + this plugin for testing this WIP solution?

sanyamdogra:
> @sanyamdogra Can you provide a docker-compose.yml with express-gateway + this plugin for testing this WIP solution?

Yes on it!

lib/express-rate-limit.js
```diff
@@ -85,6 +86,16 @@ function RateLimit(options) {
               );
             }
           }
+          if (options.draft_polli_ratelimit_headers && !res.headersSent) {
```
ioggstream: @nfriedly WDYT to set the header field names in the configuration?
This will:

- avoid duplicate code in L78-79 and L90-91
- concentrate changes to L82
- provide more flexibility

nfriedly: Those sections aren't duplicates, L78-79 has the `X-` prefix and L90-91 doesn't.

I like it the way it is now because:

1) it's backwards-compatible
2) it allows for users to choose one or both styles of headers (with or without `X-`)

I'm anticipating a future semver major release that deprecates and removes the `X-` headers; that should provide the simplification you're looking for. But, for now I think it's best to support both varieties and ship this as a semver minor.

ioggstream: Ok, let's support both for now! It seems a good choice!

lib/express-rate-limit.js
```diff
@@ -9,6 +9,7 @@ function RateLimit(options) {
       message: "Too many requests, please try again later.",
       statusCode: 429, // 429 status = Too Many Requests (RFC 6585)
       headers: true, //Send custom rate limit header with limit and remaining
+      draft_polli_ratelimit_headers: false, //Support for the new RateLimit standardization headers
```
XVincentX: Does not seem a great property name to me, maybe let's use the draft number instead of `polli`?

ioggstream: To be honest, I suggested `draft_chickens` :)

nfriedly: Yea, I thought about this a bit. I don't really like it for a permament name, but it isn't meant to be. Once the draft becomes an actual standard, I'm going to rename it to just `headers` and drop or rename the existing `headers` param. 

I think this name helps to both tie it to the draft and reinforce the idea that this is temporary and will change in the future.

The official id looks to be [draft-polli-ratelimit-headers-01](https://tools.ietf.org/id/draft-polli-ratelimit-headers-01.html) but I switched to underscores for ergonomics and dropped the number because I don't anticipate any significant changes. If there are some, we can add a `draft_polli_ratelimit_headers_02` param or whatever.

ioggstream:
Dear @nfriedly and @sanyamdogra, thank you very much for all your collaboration!
I would be happy to mention you in the spec https://github.com/ioggstream/draft-polli-ratelimit-headers/pull/80/files Hope it's fine...

nfriedly:
Fine with me, I think @sanyamdogra deserves most of the credit, though.

PR #181, opened 2020-03-28
IP are always same/localhost fixed

YanDevDe:
Hello everyone,

Most (for me it was always) in cases "req.ip" just returned the false IP Address such "::ffff:7f00:1". Which can cause to block all requests from all users even for new user.

With this fix it will allow us to get real IP Addresses and can block request properly.

Let me know if you have any further questions.

Regards,

Yan

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,11 +151,11 @@ Defaults to `false`. Behavior and name will likely change in future releases.
 
 Function used to generate keys.
 
-Defaults to req.ip:
+Defaults to `req.headers['x-forwarded-for'] || req.connection.remoteAddress || req.ip`:
 
 ```js
 function (req /*, res*/) {
-    return req.ip;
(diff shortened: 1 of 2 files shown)


nfriedly:
This happens when you're behind a load balancer or reverse proxy. Express has built-in support for it, but it is disabled by default for security reasons. To enable it, add this line:

```js
app.set('trust proxy', 1);
```

The reason it's disabled by default is that any server which is not behind a proxy could tricked by a malicious user setting the header to a random IP address and then having effectively unlimited requests.

See https://expressjs.com/en/guide/behind-Sroxies.html for more details.

YanDevDe:
Thank you for the clarify!
Btw, the link goes to 404. The correct link is https://expressjs.com/en/guide/behind-proxies.html

nfriedly:
Op, yes, sorry

PR #185, opened 2020-04-28
Handle exceptions

wandersonwhcr:
Allow express to handle exceptions with default error handler. Now, we can throw an exception into `handler` callback if limits reached.

@see https://expressjs.com/en/guide/error-handling.html#the-default-error-handler

### DIFF
--- a/lib/express-rate-limit.js
+++ b/lib/express-rate-limit.js
@@ -65,7 +65,6 @@ function RateLimit(options) {
         typeof options.max === "function" ? options.max(req, res) : options.max;
 
       Promise.resolve(maxResult)
-        .catch(next)
         .then((max) => {
           req.rateLimit = {
             limit: max,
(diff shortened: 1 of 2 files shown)


wandersonwhcr:
Sorry. I forgot to push tests.

nfriedly:
Good call. I think you can remove the `.catch()` on line 68 - it should fall through to the new one.

wandersonwhcr:
Done.

nfriedly:
It's now published to npm as v5.1.3, thanks!

nfriedly:
Oh, and BTW, what you did works fine, but this is how I would handle that lint error:

https://github.com/nfriedly/express-rate-limit/commit/3e4bfd9717ccb2cc54358e1200f2166159278d8b

```diff
+    // eslint-disable-next-line no-unused-vars
     app.use((err, req, res, next) => {
       res.status(err.code).send(err.message);
-      next; // ignore lint errors?
     });
```

wandersonwhcr:
Ah, great! <3

TY

PR #188, opened 2020-05-24
Typescript rebase

shilangyu:
In response to #138 ([this](https://github.com/nfriedly/express-rate-limit/issues/138#issuecomment-633145625) to be exact) I have ported the new changes from master to typescript branch. Most of them required just copy paste.

Some tests are failing (which they already have been doing on a clean install), I have not yet investigated why, thus I'm leaving the PR on a draft status for now.

TODO:
- [x] fix failing tests
- [ ] bump dependencies to match master
- [x] rebase new changes

PR #192, opened 2020-06-29
Test on Windows and MacOS

nfriedly:
Run the existing tests on Windows and MacOS, in addition to Linux.

Targeting the typescript branch thanks to @shilangyu's work in #188.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: node_js
 node_js:
   - node # latest stable Node.js release
   - lts/* # latest LTS Node.js release
-  - 8
+  - 10
 
 os:
   - linux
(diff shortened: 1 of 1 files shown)


nfriedly:
It looks like this makes the tests run on windows, but it does not make them pass on windows (on Travis CI): https://travis-ci.org/github/nfriedly/express-rate-limit/jobs/703295377

shilangyu:
I was hoping bumping mocha by 2 major versions would solve this problem but sadly the tests are unstable on my machine as well

nfriedly:
I think there may be some timing bugs in the test that just happened to pass on Windows and Mac but I'm not sure.

PR #217, opened 2021-01-06
redirect express to dependencies

m0h0n:
express is **dependencies** not a devDependencies

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,12 +31,13 @@
     "brute-force",
     "attack"
   ],
-  "dependencies": {},
+  "dependencies": {
+    "express": "^4.17.1"
+  },
   "devDependencies": {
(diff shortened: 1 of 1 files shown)


nfriedly:
Thanks, but I don't think this is correct. 

It might make sense to have express in peerDependencies (with a wider set of compatible versions), because it's designed to work with express, but the library doesn't `import` or `require` express, which is what dependencies usually implies. 

The library, by default, uses `req.ip`, `res.status()`, and `res.send()` but you could fairly trivially implement those yourself OR configure configure the library not use them by supplying a custom `keyGenerator` and `handler` functions.

Additionally, regardless of the above, express does belong in devDependencies with a more specific version because it is require'd when running the tests.

I hope that all makes sense.

PR #219, opened 2021-01-11
Await the `skip` method because it is async

rushi:


rushi:
Incorrectly created

nfriedly:
Hey, no worries. Allowing `skip` and `keyGenerator` to be async is probably a good idea, if you want to try a new PR with just those changes.

async/await is probably ok too, but probably belongs in it's own PR just to keep things straightforward.

rushi:
Ok, sure. I'll clean up things and raise a new PR.

PR #222, opened 2021-02-08
Fix - Warning: a promise was created in a handler but was not returne…

vpeshka:
…d from it when used with Bluebird.


nfriedly:
Awesome, thanks!

nfriedly:
Should be released in v5.2.5 shortly

PR #224, opened 2021-02-15
Change Date.toGMTString to Date.toUTCString due to deprecation

Hypfer:
See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toGMTString

Interestingly, .toGMTString is actually already .toUTCString in node v14
image


nfriedly:
Nice, thanks!

I'm only on my phone right now, so I'll merge it in now and publish it later. If you don't see it published within a couple of days please poke me to remind me.

nfriedly:
OK, this should be published in v5.2.6 by CI shortly.

PR #241, opened 2021-06-25
Ability to manually decide if request was successful

melya:
This PR adds ability to decide if request was successful (and as the result should be skipped) manually, based on request and/or response.

I had a use-case when comparison against `response.statusCode` wasn't enough, because application always responds with redirects to static pages (3xx) if error happened

Hope, this feature could be useful for community 🙌 
_Backward compatibility is saved_

### DIFF
--- a/test/express-rate-limit-test.js
+++ b/test/express-rate-limit-test.js
@@ -438,7 +438,7 @@ describe("express-rate-limit node module", () => {
     );
 
     await request(app).get("/").expect(200);
-    assert(store.decrement_was_called, "decrement was not called on the store");
+    assert(store.decrement_was_called, "decrement was called on the store");
   });
 
   it("should not decrement hits with success response with custom 'requestWasSuccessful' option", async () => {
@@ -455,7 +455,7 @@ describe("express-rate-limit node module", () => {
 
     await request(app).get("/bad_response_status").expect(403);
 
-    assert(!store.decrement_was_called, "decrement was called on the store");
+    assert(!store.decrement_was_called, "decrement was not called on the store");
   });
 
   it("should decrement hits with failed response and skipFailedRequests", async () => {
--- a/test/express-rate-limit-test.js
+++ b/test/express-rate-limit-test.js
@@ -425,6 +425,37 @@ describe("express-rate-limit node module", () => {
     assert(!store.decrement_was_called, "decrement was called on the store");
   });
 
+  it("should decrement hits with success response and skipSuccessfulRequests: statusCode < 400", async () => {
+    const store = new MockStore();
+    createAppWith(
+      rateLimit({
+        skipSuccessfulRequests: function (req, res) {
+          return res.statusCode < 400;
+        },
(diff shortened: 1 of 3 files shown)


nfriedly:
Hey, thanks, that's a good idea. One suggestion, though, since both `skipFailedRequests` and `skipSuccessfulRequests` are options already, why don't we make a new option named something like `requestWasSuccessful(req, res)`, and then both `skipFailedRequests` and `skipSuccessfulRequests` can use it.

melya:
Hey @nfriedly, good suggestion 👍 
Take a look, please

test/express-rate-limit-test.js
```diff
@@ -425,6 +425,39 @@ describe("express-rate-limit node module", () => {
...
+        skipSuccessfulRequests: true,
+        requestWasSuccessful: function (req, res) {
+          return res.statusCode === 200;
+        },
+        store: store,
+      })
+    );
+
+    await request(app).get("/").expect(200);
+    assert(store.decrement_was_called, "decrement was not called on the store");
```
nfriedly, line 441:
```suggestion
    assert(store.decrement_was_called, "decrement was called on the store");
```

test/express-rate-limit-test.js
```diff
@@ -425,6 +425,39 @@ describe("express-rate-limit node module", () => {
...
+        requestWasSuccessful: function (req, res) {
+          return res.statusCode === 200;
+        },
+        store: store,
+      })
+    );
+
+    await request(app).get("/bad_response_status").expect(403);
+
+    assert(!store.decrement_was_called, "decrement was called on the store");
```
nfriedly, line 458:
```suggestion
    assert(!store.decrement_was_called, "decrement was not called on the store");
```

test/express-rate-limit-test.js
```diff
@@ -425,6 +425,39 @@ describe("express-rate-limit node module", () => {
...
   });
 
+  it("should decrement hits with success response with custom 'requestWasSuccessful' option", async () => {
+    const store = new MockStore();
+    createAppWith(
+      rateLimit({
+        skipSuccessfulRequests: true,
+        requestWasSuccessful: function (req, res) {
+          return res.statusCode === 200;
+        },
```
nfriedly, line 435: I'm a little iffy about this because, for this test, the behavior is the same as the default `requestWasSuccessful` function.

What if you made it work off of the querystring? e.g. `return req.query.success === "1";` here and in the next test, and then do`.get("/?success=1").` here and `.get("/?success=0")` in the next test?

nfriedly (commented):
Nice, I'm really happy with where the code is at for this, I think it will be useful to some folks.

I'd like to see a couple of tweaks to the tests, and then I think this is good to go.

nfriedly (approved):
Nice work! I'll merge this in now and get it published to npm soon.

nfriedly:
v5.3.0 is now live on npm

PR #244, opened 2021-07-22
add option to allow  custom param name added to req

takeshirs:
Allow to rename the req.rateLimit parameter. Useful if using more than one rateLimiter with different keys.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,7 +102,7 @@ app.post("/create-account", createAccountLimiter, function(req, res) {
 
 A `req.rateLimit` property is added to all requests with the `limit`, `current`, and `remaining` number of requests and, if the store provides it, a `resetTime` Date object. These may be used in your application code to take additional actions or inform the user of their status.
 
-The property name can be configured with the configuration option `reqRateLimitParam`
+The property name can be configured with the configuration option `requestPropertyName`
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -230,6 +232,11 @@ function (/*req, res*/) {
 }
 ```
 
+### reqRateLimitParam
```
nfriedly, line 235: Let's rename this to `reqestPropertyName`. I feel like that's a little bit more clear as to what this controls.

nfriedly (changes requested):
Hey, thank you for doing this and sorry for taking so long to get back to you!

I have two changes I'd like to see before merging this in:

1. Please change the name to `reqestPropertyName`
2. Please add a couple of tests with a custom name, and with two rate-limiters, each with custom names, to confirm that the no longer clobber each other.

I might make some tweaks to the existing tests to make that second request more straightforward.

README.md
takeshirs: Yeah it's ok.

takeshirs:
Hey, thanks for your input. I changed the property name and added a test (it("should handle two rate-limiters independtly")).

nfriedly:
This is now published to npm in v5.4.0

Issue #212, opened 2020-11-27
express-rate-limit handling double number of max value requests


shilpamahajan9:
Hi,

I am using the library express-rate-limit to rate limit REST API to 5 request/second for a regular web server and followed the steps mentioned in https://www.npmjs.com/package/express-rate-limit/v/5.1.3

On localhost results are as expected i.e it handles 5 request/second but when deploy to testing server handles 10 requests/second i.e. double the value of max property, which is not expected.
Not sure why this is happening. What could be the possibilities that causes server to handle double the max value number of requests?

Thanks & Best regards,
Shilpa

nfriedly:
Hi @shilpamahajan9 

Most likely, you have multiple servers or processes (e.g. with the `cluster` module). The default Memory Store keeps state in the process, so if the limit is 5 and there are two servers or processes, then an end user gets 5 requests to each.

Switching to any of the alternate stores will remedy this, although each requires some type of data store to synchronize everything.

shilpamahajan9:
Thank you so much for your reply.
It was very helpful!

I have implemented the rate limiting but now struggling to test it. I tried below ways to test the rate limiting but not sure which one is correct as results are different for each of them. 

**Postman**: during windowMs, we are sending requests every 50 milliseconds. Results: 5 requests are successful and rest HTTP 429
**Shell script**: In the for loop sending number of requests for windowMs. This windowMs calculated like start of shell script execution + windowMs. Results: 12 to 15 requests are successful and rest HTTP 429.
**ab testing**: ab -t windowMs -c 50 -H "X-MyIO-Access: mykey" -H "X-MyIO-Account: myAccount" -H "Content-Type: application/json" -v 4 https://myurl/{id} Results: In the given windowMs, handling 20+ requests and rest HTTP 429

Not sure which test method is correct. Could you please suggest a best way to test our APIs rate limit. Any recommendation for tools. Your suggestions are welcome :)

Best regards,
Shilpa

nfriedly:
How many server processes do you have running, and what store are you using?

The other thing I should mention is that on the default Memory Store, requests are bucketed to an internal clock. So if your limit is 5 per second, that second-timer doesn't start when the first request comes in, it starts when the server starts. 

If an end-user got the timing *just so*, they might manage to get 10 requests in a row because half go into one second and half go into the next. But they'd have to wait 2 seconds to be able to do that again. (It's an optimization to reduce the amount of memory and computation the Memory Store has to do.)

[...]

Issue #207, opened 2020-11-03
Inconsitent limits


orangejuicejones:
Hi,
I'm running node + express with express-rate-limit for the first time with the below config.

```js
const limiter = rateLimit({
  windowMs: 1 * 60 * 1000, // 1 minute
  max: 5 // limit each IP to 100 requests per windowMs
});

ProductRoute.post('/data', limiter, cacheMiddleware(20), (req, res) => {
// return data
}
```

I understand that the user can make up to 5 calls in every minute. When I tested, I noticed I could make more than 5 requests in a minute if i make a request every 5 seconds. I also noticed if I make a call every second, I then get blocked on the 6th. 
I would have expected not more than 5 calls can be made in a minute irrespective.

nfriedly:
Hi @orangejuicejones,

That's because the 1 minute doesn't start at the first request. The way the built-in memory store works is that all requests during a given minute get bucketed and then after the minute is elapsed, the bucket is emptied. So, in your 1 request every 5 seconds example, the bucket was emptied after you had made <=5 requests, resetting the counter to 0 and allowing you up to 5 requests for the next bucket.

With this setup, a user could get a burst of up to 10 requests if they timed it _just right_. But then they would have to wait a full minute to get 5 more requests, or two minutes to get another burst of 10. It's still 5/minute, it's just that the server's minute doesn't necessarily align to their clock.

This is a memory/load optimization that I felt was appropriate for the default store. However, if you want precise tracking, starting a per-user clock at exactly when the first request comes in, I believe all of the custom stores, as well as the other rate limiters mentioned in the readme offer this feature.

PR #246, opened 2021-08-05
Add precision configuration option

SimonKocurek:
Currently it is possible for user/attacker to send `max * 2` requests if they send half of the requests just before `hits` map gets cleared... to mitigate this issue, requests are split into buckets (effectively dividing whole `windowMs` into multiple sub-windows).

Inspired by https://www.figma.com/blog/an-alternative-approach-to-rate-limiting

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,9 +124,9 @@ Defaults to `60000` (1 minute).
 
 Number of request buckets inside time window. Buckets group requests based on timestamp for efficient cleanup.
 
-With precision set to `1` all remembered requests are reset, when time window expires. Increasing this value too much will result in decreased performance.
+This option is currently utilized only by `memory-store`, other stores rely on databases with "infinite" per request precision. With precision set to `1` all remembered requests are reset, when time window expires. Increasing this value too much will result in decreased performance.
 
-Defaults to `4`.
+Defaults to `1`.
 
 ### message
 
--- a/README.md
+++ b/README.md
@@ -120,6 +120,14 @@ Note: with non-default stores, you may need to configure this value twice, once 
 
 Defaults to `60000` (1 minute).
 
+### precision
+
+Number of request buckets inside time window. Buckets group requests based on timestamp for efficient cleanup.
+
+With precision set to `1` all remembered requests are reset, when time window expires. Increasing this value too much will result in decreased performance.
+
+Defaults to `4`.
+
 ### message
 
 Error message sent to user when `max` is exceeded.
--- a/lib/express-rate-limit.js
+++ b/lib/express-rate-limit.js
@@ -10,6 +10,7 @@ function RateLimit(options) {
       statusCode: 429, // 429 status = Too Many Requests (RFC 6585)
       headers: true, //Send custom rate limit header with limit and remaining
       draft_polli_ratelimit_headers: false, //Support for the new RateLimit standardization headers
+      precision: 4, //Number of request buckets inside time window. Buckets serve as a performant alternative to tracking timestamp of each request and clearing them one by one.
       // ability to manually decide if request was successful. Used when `skipSuccessfulRequests` and/or `skipFailedRequests` are set to `true`
       requestWasSuccessful: function (req, res) {
         return res.statusCode < 400;
@@ -31,8 +32,13 @@ function RateLimit(options) {
     options
   );
 
+  if (options.precision <= 0) {
+    throw new Error("Rate limiter precision must be positive number");
+  }
+
   // store to use for persisting rate limit data
-  options.store = options.store || new MemoryStore(options.windowMs);
+  options.store =
+    options.store || new MemoryStore(options.windowMs, options.precision);
 
   // ensure that the store has the incr method
   if (
--- a/lib/express-rate-limit.js
+++ b/lib/express-rate-limit.js
@@ -10,7 +10,7 @@ function RateLimit(options) {
       statusCode: 429, // 429 status = Too Many Requests (RFC 6585)
       headers: true, //Send custom rate limit header with limit and remaining
       draft_polli_ratelimit_headers: false, //Support for the new RateLimit standardization headers
-      precision: 4, //Number of request buckets inside time window. Buckets serve as a performant alternative to tracking timestamp of each request and clearing them one by one.
+      precision: 1, //Number of request buckets inside time window. Buckets serve as a performant alternative to tracking timestamp of each request and clearing them one by one.
       // ability to manually decide if request was successful. Used when `skipSuccessfulRequests` and/or `skipFailedRequests` are set to `true`
       requestWasSuccessful: function (req, res) {
         return res.statusCode < 400;
--- a/lib/memory-store.js
+++ b/lib/memory-store.js
@@ -6,41 +6,60 @@ function calculateNextResetTime(windowMs) {
   return d;
 }
 
-function MemoryStore(windowMs) {
-  let hits = {};
+function MemoryStore(windowMs, precision) {
+  // Buckets divide `windowMs` into multiple subparts that can be cleared
+  // individually, allowing for better precision when clearing outdated requests.
+  // After a bucket has expired, all requests that were assigned to it's time window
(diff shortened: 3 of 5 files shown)


README.md
```diff
@@ -120,6 +120,14 @@ Note: with non-default stores, you may need to configure this value twice, once
 
 Defaults to `60000` (1 minute).
 
+### precision
+
+Number of request buckets inside time window. Buckets group requests based on timestamp for efficient cleanup.
+
+With precision set to `1` all remembered requests are reset, when time window expires. Increasing this value too much will result in decreased performance.
+
+Defaults to `4`.
```
nfriedly, line 129: This should clarify that `precision` only applies to the built-in memory store. 

(Other stores generally track the time per-user and so essentially have "infinity precision".)

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "express-rate-limit",
-  "version": "5.3.0",
+  "version": "5.4.0",
```
nfriedly, line 3: You should leave this out of the PR. After it's merged in, I'll use the `npm version minor` command which both increments the version number and makes a git tag.

lib/express-rate-limit.js
```diff
@@ -10,6 +10,7 @@ function RateLimit(options) {
       statusCode: 429, // 429 status = Too Many Requests (RFC 6585)
       headers: true, //Send custom rate limit header with limit and remaining
       draft_polli_ratelimit_headers: false, //Support for the new RateLimit standardization headers
+      precision: 4, //Number of request buckets inside time window. Buckets serve as a performant alternative to tracking timestamp of each request and clearing them one by one.
```
nfriedly, line 13: Lets make the default `1`, otherwise this could end up surprising some people. I might bump it up to 4 in the next major release, though.

lib/memory-store.js
```diff
@@ -6,41 +6,60 @@ function calculateNextResetTime(windowMs) {
   return d;
 }
```
nfriedly, line 7: `calculateNextResetTime` should also take the precision into account. It might need to look through the buckets too (?) We basically want it to give out the next time that the current user could make a request.

nfriedly (changes requested):
Hey, thanks for doing this and sorry for taking so long to review it.

I left a few requests below, but overall I really like this.

lib/memory-store.js
SimonKocurek: There is iterator which takes precision into account and moves onto next bucket when this time expires:
```js
  const bucketTime = windowMs / precision;
  const bucketResetInterval = setInterval(this.resetBucket, bucketTime);
```

`NextResetTime` is just pure `windowMs`, since it is sent via HTTP headers to inform the client when will server forget his requests. That happens after all buckets are cleared.

I might have wrong understanding of this header though. If you are not convinced I can change it to also take `precision` into account. :)

SimonKocurek: After reading [https://tools.ietf.org/id/draft-polli-ratelimit-headers-00.html#rfc.section.3.3](https://tools.ietf.org/id/draft-polli-ratelimit-headers-00.html#rfc.section.3.3) I see that I've understood it wrongly.  Thought it is used in header with time when user will be 100% safe to send new request.

test/memory-store-test.js
```diff
@@ -4,7 +4,7 @@ const MemoryStore = require("../lib/memory-store.js");
 
 describe("MemoryStore store", () => {
   it("sets the value to 1 on first incr", (done) => {
-    const store = new MemoryStore(-1);
+    const store = new MemoryStore(-1, 4);
```
nfriedly, line 7: Lets undo these changes and instead add some tests that specifically check different precision values.

```suggestion
    const store = new MemoryStore(-1);
```

test/memory-store-test.js
```diff
@@ -21,7 +21,7 @@ describe("MemoryStore store", () => {
   });
 
   it("increments the key for the store each incr", (done) => {
-    const store = new MemoryStore(-1);
+    const store = new MemoryStore(-1, 4);
```
nfriedly, line 24:
```suggestion
    const store = new MemoryStore(-1);
```

test/memory-store-test.js
```diff
@@ -40,7 +40,7 @@ describe("MemoryStore store", () => {
   });
 
   it("resets the key for the store when used with resetKey", (done) => {
-    const store = new MemoryStore(-1);
+    const store = new MemoryStore(-1, 4);
```
nfriedly, line 43:
```suggestion
    const store = new MemoryStore(-1);
```

test/memory-store-test.js
```diff
@@ -61,7 +61,7 @@ describe("MemoryStore store", () => {
   });
 
   it("resets all keys for the store when used with resetAll", (done) => {
-    const store = new MemoryStore(-1);
+    const store = new MemoryStore(-1, 4);
```
nfriedly, line 64:
```suggestion
    const store = new MemoryStore(-1);
```

test/memory-store-test.js
```diff
@@ -94,7 +94,7 @@ describe("MemoryStore store", () => {
   });
 
   it("resets all keys for the store when the timeout is reached", (done) => {
-    const store = new MemoryStore(50);
+    const store = new MemoryStore(50, 4);
```
nfriedly, line 97:
```suggestion
    const store = new MemoryStore(-1);
```

test/memory-store-test.js
```diff
@@ -147,7 +147,7 @@ describe("MemoryStore store", () => {
     });
 
     it("can run in electron where setInterval does not return a Timeout object with an unset function", (done) => {
-      const store = new MemoryStore(-1);
+      const store = new MemoryStore(-1, 4);
```
nfriedly, line 150:
```suggestion
    const store = new MemoryStore(-1);
```

test/memory-store-test.js
```diff
@@ -171,7 +171,7 @@ describe("MemoryStore store", () => {
   });
 
   it("decrements the key for the store each decrement", (done) => {
-    const store = new MemoryStore(-1);
+    const store = new MemoryStore(-1, 4);
```
nfriedly, line 174:
```suggestion
    const store = new MemoryStore(-1);
```

nfriedly (changes requested):
Sorry again for taking so long.

I left a few minor suggestions below, and then the last thing I'd like to see are a few tests exercising different values of precision to check how many hits are remaining after different amounts of time. You can model it on this one or bring a clock mocking library if you prefer: 
https://github.com/nfriedly/express-rate-limit/blob/c4f9c46ddf6d6f86861007df431d775deb5f379c/test/memory-store-test.js#L96-L133

gamemaker1:
Hi @SimonKocurek,

Thanks for taking the time to make this contribution!

Do you wish to continue working on this PR? I see that this PR has had no activity since September last year. In the time between, we have rewritten the library in Typescript and made quite a few changes to the `Store` interface too. I completely understand if you do not have the time to continue work on this, in which case I could maybe implement this myself (@nfriedly would you still be interested in seeing this merged?).

Thanks!

SimonKocurek:
@gamemaker1 I have no plans in the near future to finish this. Feel free to handle this PR. (AFAIK nobody has requested this feature, so even closing is fine by me.)

gamemaker1:
> AFAIK nobody has requested this feature

There are a couple of issues that would be solved with this increased precision feature like https://github.com/nfriedly/express-rate-limit/issues/212 and https://github.com/nfriedly/express-rate-limit/issues/207. I think this would be a good enhancement to the memory store, while still keeping it simple.

> I have no plans in the near future to finish this.

I'll close this PR for now, and start a new one from the `master` branch, to avoid dealing with the merge conflicts. Thanks for creating this PR and giving us this awesome idea :D

nfriedly:
Yeah, I'm ambivalent on this one. I don't think it's a bad idea, but my main issue is that the mechanics of the MemoryStore are already complex enough that they often confuse people, and this just dials it up a notch. Perhaps it would make more sense in an external store?

Issue #214, opened 2020-12-20
handler should have access to options, just like onLimitReached


rubenstolk:
`handler` should have access to `options`, just like `onLimitReached`

nfriedly:
Hey, sorry this slipped through the cracks. I suppose that's a fair enough request, though.  It'd have to be a 4th param after `next`. Want to send a PR with the change and a test or two?

rubenstolk:
Would love to, not sure if I'll find some time soon though...

takeshirs:
Hey,
created a pull request for this:
https://github.com/nfriedly/express-rate-limit/pull/245

nfriedly:
I believe this is possible without any changes. When the handler is called, the `this` value will point to `options`:

```js
rateLimit({
  windowMs: 60 * 60 * 1000, // 1 hour window
  max: 5, // start blocking after 5 requests
  handler: function(req, res, next) {
    const options = this;
    console.log(`max is ${options.max}`);
    res.status(options.statusCode).send(options.message);
  }
});
```

The main gotcha that you have to watch out for is that you can't use "fat arrow" (`() => {}`) functions, because they change the behavior of `this`.

NoahAndrews:
The fact that you can access it that way feels like a side effect of an implementation detail. I think passing in the options object as a parameter is much more explicit and clear.

nfriedly:
> feels like a side effect of an implementation detail.

That's true, it more or less is. 

But I really don't like adding a new param *after* the callback function. That just feels wrong. 

I also thought about trying to be clever and check if `options.handler.length >= 4` and then pass in `req, res, options, next` otherwise just `req, res, next`. That would do the right thing most of the time for most people.. but it'd be a nightmare to debug when it didn't. So I think that solution is out.

I'd also consider making a breaking API change and a major version release if I saw a lot of value in it... but, using `this` does work in the current codebase, and I'm willing to document and commit to keeping it working in future revisions. It's an awkward API, but I think it's the best option for the moment.

takeshirs:
Hm, I can understand your concerns.
But regarding:

> But I really don't like adding a new param after the callback function. That just feels wrong.

It's just like it's done for the `handler` option, so I think it would be consistent to also add the `optionsUsed` parameter to `onLimitReached`.

StephanBijzitter:
While `options` coming after `next` feels weird indeed, it would still be a non-breaking change.
[...]

PR #249, opened 2021-09-02
Change handler to not use express specific function

ceifa:
Hey!
I know the package have express on the name, but this is the only thing we need to do to work on [polka](https://github.com/lukeed/polka) too!

### DIFF
--- a/lib/express-rate-limit.js
+++ b/lib/express-rate-limit.js
@@ -24,7 +24,8 @@ function RateLimit(options) {
         return false;
       },
       handler: function (req, res /*, next*/) {
-        res.status(options.statusCode).send(options.message);
+        res.writeHead(options.statusCode);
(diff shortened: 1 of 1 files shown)


nfriedly:
Hey, thanks! 

But, that's not quite equivalent - express's `req.send()` performs a bit of "magic" under the hood, for example it automatically stringifies arrays and objects objects and set some appropriate content-type headers. I know of a few people who are using that feature, so this change would definitely break things for them :(

Maybe you should make a `polka-rate-limit` package that just wraps this one, but sets the default handler to yours?

ceifa:
Thanks for the answer, I see your point!

> Maybe you should make a `polka-rate-limit` package that just wraps this one, but sets the default handler to yours?

Yep, this is an option. Needs #214

PR #257, opened 2021-10-25
Update memory-store.js

Nebula1213:
Use an expiring dict instead of just deleting all of the data every certain amount of seconds.


nfriedly:
Your lib is linked to from the Readme, and it's published in v5.5.1 - https://www.npmjs.com/package/express-rate-limit :)

PR #263, opened 2021-11-29
Move to Typescript and ESM

### LINKED ISSUE #114: Add a `CHANGELOG`
Thanks for this project!

Would it be possible to add a `CHANGELOG` to make it easy to see new changes without going through commits?

The `README` shows major versions changes, but not minor versions.

### LINKED ISSUE #138: TypeScript typings and tests

### LINKED ISSUE #153: Provide an alternate to npm for JS compiled from TS
@Martii brought up a good point in the context of transitioning to Typescript (#138 & #152):

> >  ... With the current setup in this branch, installing directly from github won't work
>
> This could be a potential 👎

https://github.com/nfriedly/express-rate-limit/pull/152#issuecomment-509433428

I agree with this sentiment. I think github releases is probably the correct place to store the compiled code on github, but I'm not sure if npm can install from a github release, so we may need to also stick it into git in order to meet that need.

gamemaker1:
Hi @nfriedly!

I saw #138 and decided to contribute by converting the library to Typescript and ESM. Some of the changes are breaking, while others a bit opiniated, so let me know if you would like me to revert/change anything.

I have made the PR against `master`, since you mentioned you would like to merge the `typescript` branch into `master` at some point [[#138 (comment)](https://github.com/nfriedly/express-rate-limit/issues/138#issuecomment-510668579)], but let me know if I should make it against the `typescript` branch instead.

## Changes

- Converted the library itself and tests to Typescript. [closes #138]
- Converted the library to an ES module.
- Upgrade all dependencies to latest versions.
- Ported newer features from the JS branch (e.g., added the `requestWasSuccessful` option, etc.).
- Allows `keyGenerator` to be an async function.
- Renamed `draft_polli_ratelimit_headers` => `useStandardizedHeaders`
- Added a changelog [closes #114]
- Publish a Github Release (via the `Publish`) workflow with the compiled JS code post `npm publish`. [closes #153]
- Renamed some files and folders, e.g. `lib/` => `source/`, `LICENSE` => `license.md`, etc.
- Use `jest` to test instead of `mocha` (why: easier mocks, built in coverage).
- Use `xo` to lint code (why: combined linter and formatter, no config needed).
- Use `lint-staged` + `husky` instead of `pretty-quick` to lint staged files.


gamemaker1:
Hi @nfriedly,

Eager to know if you got a chance to review the PR?

Thanks

nfriedly:
Shoot, no, sorry. I feel like I've been underwater lately. I have some time off next week, that should help me get caught back up.

gamemaker1:
No worries, thanks for the prompt reply :)

reslear:
@gamemaker1 good job 👍💪
@nfriedly wait you return ))

readme.md
```diff
@@ -0,0 +1,417 @@
+# Express Rate Limit
```
nfriedly, line 1: Would you mind undoing the rename for *this* PR? (You can still do it in a followup if you want, it's just that github doesn't recognize that it's the same file, and so it's hard to see what actually changed.)

source/lib.ts
```diff
@@ -0,0 +1,271 @@
...
+const parseOptions = (passedOptions: Partial<Options>): Options => {
+	const {
+		// Default window of time is 1 minute
+		windowMs = 60 * 1000,
+		// The default storage provider used is a {@link MemoryStore}
+		store = new MemoryStore(windowMs),
+	} = passedOptions
+
+	// Now add the defaults for the other options
+	const options: Options = {
```
nfriedly, line 25: Why is this two separate blocks?

source/lib.ts
```diff
@@ -0,0 +1,271 @@
...
+			const augmentedRequest = request as AugmentedRequest
+
+			// Get a unique key for the client
+			const key = await Promise.resolve(options.keyGenerator(request, response))
+			// Increment the client's hit counter by one
+			const { totalHits, resetTime } = await new Promise<{
+				totalHits: number
+				resetTime: Date | undefined
+			}>((resolve, reject) => {
+				options.store.increment(
```
nfriedly, line 148: I like the rename to `increment`, but we also need to support `options.store.incr()` here for backwards compatibility with all of the third-party stores.

source/types.ts
```diff
@@ -0,0 +1,247 @@
...
+ * An interface that all hit counter stores must implement.
+ */
+export interface Store {
+	/**
+	 * Method to increment a client's hit counter.
+	 *
+	 * @param key {string} - The identifier for a client
+	 * @param callback {IncrementCallback} - The callback to call once the counter is incremented
+	 */
+	increment: (key: string, callback: IncrementCallback) => void
```
nfriedly, line 75: As I mentioned above, we need to also support legacy stores with the `incr()` method. 

Is there a way of marking that as deprecated, or telling typescript that it needs to have one or the other of `incr`, `increment` (but not both)?  Maybe just make them both option and explain it in the documentation otherwise.

nfriedly (changes requested):
Thank you! In general, this looks really great! I have a few changes I requested below, and I'm not finished looking through things yet, but I wanted to post what I have so far.

nfriedly:
I just had one thought: maybe instead of explicitly supporting `incr()` in the `Store` interface, we provide a "legacy-to-modern" adapter thing, that takes a store with `incr()` API and wraps it to return a new store that matches the new `increment()` API. I feel like that would be cleaner, we'd just need to document how to use it in the readme and call it out in the changelog.

.editorconfig
```diff
@@ -1,10 +0,0 @@
-# EditorConfig: automatically set indentation and such
-# http://EditorConfig.org
-
-# top-most EditorConfig file
-root = true
-
-[*.{js,json}]
-charset = utf-8
-indent_style = space
-indent_size = 2
```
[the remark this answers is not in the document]
gamemaker1: I prefer keeping the style configuration in one place (prettier's config), and `husky` and `lint-staged` do a good job at keeping everything formatted before pushing changes.

I could still restore this file, since most editors respect `.editorconfig`s without any configuration.

source/types.ts
gamemaker1: We could mark `incr` as deprecated but still support it if it exists. Something like this:

```ts
/**
 * An interface that all hit counter stores must implement.
 */
export interface Store {
	/**
	 * (Deprecated) Method to increment a client's hit counter.
	 *
	 * @param key {string} - The identifier for a client
	 * @param callback {IncrementCallback} - The callback to call once the counter is incremented
	 *
	 * @deprecated 6.x - Implement the `increment` function instead.
	 */
	incr?: (key: string, callback: IncrementCallback) => void

	/**
	 * Method to increment a client's hit counter.
	 *
	 * @param key {string} - The identifier for a client
	 * @param callback {IncrementCallback} - The callback to call once the counter is incremented
	 */
	increment: (key: string, callback: IncrementCallback) => void

	// ...
}
```

And then replace [this](https://github.com/gamemaker1/express-rate-limit/blob/opiniated-improvements/source/lib.ts#L148) with:

```ts
(options.store.increment ?? options.store.incr)(
	...
)
```

Does that sound good?

gamemaker1:
> We provide a "legacy-to-modern" adapter ...

That sounds great - we could even make the 'newer' `Store` functions async instead of using callbacks.

source/lib.ts
gamemaker1: Not sure... will combine it with the second block

nfriedly:
> That sounds great - we could even make the 'newer' Store functions async instead of using callbacks.

Yea, I like that. 

Going one step further, we can probably automatically detect and wrap stores using the legacy API. So end-users can upgrade with no code changes.

source/lib.ts
```diff
@@ -0,0 +1,264 @@
...
+const parseOptions = (passedOptions: Partial<Options>): Options => {
+	// Now add the defaults for the other options
+	const options: Options = {
+		windowMs: 60 * 1000,
+		store: new MemoryStore(passedOptions.windowMs ?? 60 * 1000),
+		max: 5,
+		message: 'Too many requests, please try again later.',
+		statusCode: 429,
+		headers: true,
+		useStandardizedHeaders: false,
```
nfriedly, line 25: Ditto above comments about renaming these. I think I like `legacyHeaders` and `standardHeaders` - but I think we should continue defaulting to legacy on and standard off for backwards compatibility. (As you've done here.)

source/lib.ts
```diff
@@ -0,0 +1,264 @@
...
+	>
+	for (const option of deprecatedOptions) {
+		// This doesn't trigger if any value is set to a falsy value (e.g., 0),
+		// because that essentially disables them
+		if (passedOptions[option]) {
+			throw new Error(
+				`The \`${option}\` option is deprecated and will likely be removed from the \`express-rate-limit\` package in the future.`,
+			)
+		}
+	}
```
nfriedly, line 70: I think we can get rid of these checks - but we're going to want to have similar code to check for the old headers names and recommend the new ones.

source/lib.ts
```diff
@@ -0,0 +1,264 @@
...
+	if (
+		typeof options.store.increment !== 'function' ||
+		typeof options.store.resetKey !== 'function' ||
+		(options.skipFailedRequests &&
+			typeof options.store.decrement !== 'function')
+	) {
+		throw new Error(
+			'An invalid store was passed. Please ensure that the store is a class that implements the `Store` interface.',
+		)
+	}
```
nfriedly, line 56: Oh, and we'll have to make sure this does the wrapping thing if store.incr is a function, rather than just throwing an error.

source/lib.ts
```diff
@@ -0,0 +1,264 @@
...
+ *
+ * @param options {Options} - The options the user specifies
+ *
+ * @returns {Options} - A complete configuration object
+ */
+const parseOptions = (passedOptions: Partial<Options>): Options => {
+	// Now add the defaults for the other options
+	const options: Options = {
+		windowMs: 60 * 1000,
+		store: new MemoryStore(passedOptions.windowMs ?? 60 * 1000),
```
nfriedly, line 20: I think we should stop passing anything into the constructor, and instead add an `init(config, limiter)` method to the store API - that way third-party stores would have an official way of getting the config instead of the current solution of requiring end-users to pass in the values twice. We'd want to make the init method optional and only call it if it exists.

source/lib.ts
```diff
@@ -0,0 +1,264 @@
...
+				}
+
+				options.handler(request, response, options)
+				return
+			}
+
+			next()
+		},
+	)
+
```
nfriedly, line 254: Probably add that init call right here. Something like this:

```suggestion
if (typeof options.store.init === 'function') {
  options.store.init(options, middleware);
}
```

source/memory-store.ts
```diff
@@ -0,0 +1,111 @@
...
+		this.hits = {}
+		this.resetTime = calculateNextResetTime(windowMs)
+
+		// Reset hit counts for ALL clients every windowMs
+		const interval = setInterval(() => {
+			this.resetAll()
+		}, windowMs)
+		if (interval.unref) {
+			interval.unref()
+		}
```
nfriedly, line 59: ... and then move the `windowMs` bits at the beginning and end here to the new `init()` method

source/types.ts
nfriedly: Yea, I'm OK with either this approach or the wrapper. I kind of like the wrapper better, I could live with either one.

nfriedly (changes requested):

source/types.ts
gamemaker1: > Going one step further, we can probably automatically detect and wrap stores using the legacy API. So end-users can upgrade with no code changes.

I went ahead and implemented a basic `promisifyStore` function to wrap up legacy stores. Will add the `store.init` part now.

reslear:
@gamemaker1 and see my pr )

gamemaker1:
Thanks for the PR - it has some conflicts, so I will rebase and merge it.

> Rename `changelog` file to caps `CHANGELOG.md`

I prefer not to shout in filenames - I renamed the readme file so that @nfriedly would be able to review the changes made there easily and will revert that change when the PR is ready to be merged.

README.md
```diff
@@ -1,349 +1,422 @@
...
 ```
 
+Replace `{version}` with the version of the package that you want to your, e.g.:
+`6.0.0`.
+
+> **This library (v6.0.0 or greater) is now pure ESM**. CommonJS packages will
+> not be able to synchronously import this package, and it is recommended to use
+> the package with NodeJS version 12 or greater. Take a look at this
+> [article](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c)
+> for more details.
```
nfriedly, line 54: Isn't there a way to ship both via a small wrapper? So that users doing `const rateLimit = require("express-rate-limit");` still get the same behavior instead of having to change to `import` or `require(...).default` ?

README.md
```diff
@@ -196,23 +196,23 @@ HTTP status code returned when `max` is exceeded.
...
+### `standardHeaders`
 
 Enable headers conforming to the
 [ratelimit standardization proposal](https://tools.ietf.org/id/draft-polli-ratelimit-headers-01.html):
 `RateLimit-Limit`, `RateLimit-Remaining`, and, if the store supports it,
 `RateLimit-Reset`. May be used in conjunction with, or instead of the `headers`
 option.
 
-Defaults to `false`. Behavior and name will likely change in future releases.
+Defaults to `false` (for backward compatibility), but recommended to use.
```
reslear, line 215: Then you need to specify the `true` in all the examples that new users would use by default `standardHeaders`

README.md
nfriedly: Or, if not a wapper, just run the type script compiler twice: once in CJS mode, and once in ESM mode, and then ship both outputs. 

Basically, I want the majority of existing users to be able to upgrade to the 6.0 release without having to make changes to their code.

source/lib.ts
```diff
@@ -0,0 +1,323 @@
...
+const parseOptions = (passedOptions: Partial<Options>): Options => {
+	// Now add the defaults for the other options
+	const options: Options = {
+		windowMs: 60 * 1000,
+		store: new MemoryStore(passedOptions.windowMs ?? 60 * 1000),
+		max: 5,
+		message: 'Too many requests, please try again later.',
+		statusCode: 429,
+		legacyHeaders: true,
+		standardHeaders: false,
```
nfriedly, line 92:
```suggestion
		// support the v5.x names (headers and draft_polli_ratelimit_headers)
		// if the v6.x name is also set, it will override the v5.x names
		legacyHeaders: passedOptions.headers ?? true,
		standardHeaders: passedOptions.draft_polli_ratelimit_headers ?? false,
```

I'm 80% sure this will do the right thing.

source/lib.ts
```diff
@@ -0,0 +1,323 @@
...
+		'headers',
+		'draft_polli_ratelimit_headers',
+	] as Array<keyof Options>
+	for (const option of deprecatedOptions) {
+		if (typeof passedOptions[option] !== 'undefined') {
+			throw new TypeError(
+				`The \`${option}\` option is deprecated/renamed and will likely be removed from the \`express-rate-limit\` package in the future.`,
+			)
+		}
+	}
```
nfriedly, line 143: All of this can go - the global and delay options haven't been supported in years, and the above change should handle the headers options.
```suggestion

```

nfriedly (commented):
Apologies for the thrashing on this, but the more I think about it, the more I think we should just "do the right thing" if the v5.x param names are used for the legacy and standard header options. 

I haven't tested it but I think this pair of changes will do the trick.

README.md
gamemaker1: > Or, if not a wapper, just run the type script compiler twice: once in CJS mode, and once in ESM mode, and then ship both outputs.

Sure, we could do that. Maybe something like this in package.json:

```
"module": "dist/esm/index.js",
"main": "dist/cjs/index.js",
"exports": {
    ".": {
        "import": "./dist/esm/index.js",
        "require": "./dist/cjs/index.js"
    }
},
```

> Basically, I want the majority of existing users to be able to upgrade to the 6.0 release without having to make changes to their code.

👍🏾

source/lib.ts
gamemaker1: Doing this will effectively mean that both the old (`headers`, `draft_polli_ratelimit_headers`) and new (`legacyHeaders`, `standardHeaders`) options are valid and we will have two sets of options that do the same thing.

Since this is anyways a major version bump, I think it should be ok if we mark those as deprecated and throw an error if they are passed instead of still supporting them.

source/lib.ts
gamemaker1: I could remove the `global`, `delayMs` and `delayAfter` options and add a custom deprecation message for the header options.

source/lib.ts
gamemaker1: I'm curious, how would the stores use the middleware passed as the `limiter` param in the `init` function?

source/lib.ts
reslear: yep, we write to README, migration guide like breaking changes

source/lib.ts
reslear: yep, we write to README, migration guide like breaking changes

nfriedly: > Doing this will effectively mean that both the old (headers, draft_polli_ratelimit_headers) and new (legacyHeaders, standardHeaders) options are valid and we will have two sets of options that do the same thing.

Yea, exactly so. And, while that's not great, I do think it's preferable to making a breaking change that doesn't need to be a breaking change. It's about empathy for the end users: we're the ones who decided to rename these things - everyone else just installed the library and followed the directions correctly. (Or, in many cases, someone who doesn't even work there any more installed and configured it, so the new person would be starting from scratch.)

Also, while I like clean code, it's literally a 2-line change to continue supporting the old names in addition to the new ones. This thing gets downloaded 300k times per week - think about how much time it takes for us to not break this, vs the time it would take for everyone else to change their code just to work with our new release.

We'll call it out in the changelog as a non-breaking change, and update the readme to document the new names and mark the old ones as deprecated.

gamemaker1: > Think about how much time it takes for us to not break this, vs the time it would take for everyone else to change their code just to work with our new release

I agree 👍🏾 

I'll mark the old ones as deprecated, but still support them as suggested.

.editorconfig
```diff
@@ -1,10 +1,14 @@
-# EditorConfig: automatically set indentation and such
-# http://EditorConfig.org
+# /.editorconfig
+# Tells most editors what our style preferences are
```
nfriedly, line 2: Lets keep the url here.
```suggestion
# Tells most editors what our style preferences are
# http://EditorConfig.org
```

changelog.md
```diff
@@ -0,0 +1,84 @@
...
+
+- Rename the `draft_polli_ratelimit_headers` option to `standardHeaders`.
+- Rename the `headers` option to `legacyHeaders`.
+- Change the way custom stores are defined.
+  - Add the `init` method for stores to set themselves up using options passed
+    to the middleware.
+  - Rename the `incr` method to `increment`.
+  - Allow the `increment`, `decrement`, `resetKey` and `resetAll` methods to
+    return a promise.
+  - Old stores will automatically promisified and used.
```
nfriedly, line 28:
```suggestion
  - Old stores will automatically be promisified and used.
```

config/tsconfig/cjs.json
```diff
@@ -0,0 +1,10 @@
+// /config/tsconfig/cjs.json
+// Typescript compiler configuration for building a CommonJS module
```
nfriedly, line 2: I don't think comments are allowed in .json files - github is flagging them as red. 

I sometimes add a "description" field to the json object. Or we could change the file from .json to .ts and export the config.

gamemaker1: This is actually a `tsconfig.json` file, which allows comments - which is why the builds still work. However, GitHub does not realise that as it thinks (mostly correctly) that only files named `tsconfig.json` are TS compiler config files.

> I sometimes add a "description" field to the json object. Or we could change the file from .json to .ts and export the config.

The configuration for the TS compiler has to be a JSON file IIRC. If needed, I could remove the comments instead.

source/lib.ts
```diff
@@ -0,0 +1,326 @@
...
+
+/**
+ * Type guard to check if a store is legacy store.
+ *
+ * @param store {LegacyStore | Store} - The store to check
+ *
+ * @return {boolean} - Whether the store is a legacy store
+ */
+const isLegacyStore = (store: LegacyStore | Store): store is LegacyStore =>
+	typeof (store as LegacyStore).incr === 'function'
```
nfriedly, line 24: perhaps we should only consider it as legacy if it has `incr` and it *doesn't* have `increment` - that way store authors can keep `incr` around for backwards-compatibility, while also using the new API. 

Or, perhaps we just add a `apiVersion` field to the `Store` interface - if it's unset or <2, then do the upgrade. If it's >= 2, don't. I don't know.. maybe I'm overthinking this one...

gamemaker1: > Perhaps we should only consider it as legacy if it has incr and it doesn't have increment - that way store authors can keep incr around for backwards-compatibility, while also using the new API.

Good point, I'll add this.

> Or, perhaps we just add a apiVersion field to the Store interface - if it's unset or <2, then do the upgrade. If it's >= 2, don't. I don't know.. maybe I'm overthinking this one...

A version number sounds a bit overkill, but it might be a good idea if there are a lot of changes to the `Store` interface in the future...

README.md
```diff
@@ -1,349 +1,449 @@
...
 // see https://expressjs.com/en/guide/behind-proxies.html
 // app.set('trust proxy', 1);
 
 const limiter = rateLimit({
-  windowMs: 15 * 60 * 1000, // 15 minutes
-  max: 100 // limit each IP to 100 requests per windowMs
-});
+	windowMs: 15 * 60 * 1000, // 15 minutes
+	max: 100, // Limit each IP to 100 requests per `window` (here, per 15 minutes)
+	standardHeaders: true, // Return rate limit info in the `RateLimit-*` headers
```
nfriedly, line 70: Lets put both `standardHeaders` and `legacyHeaders` options in here. Just enabling the standard headers without disabling the legacy headers will mean both get sent.

```suggestion
	standardHeaders: true, // Return rate limit info in the `RateLimit-*` headers
	legacyHeaders: false, // Disable the `X-RateLimit-*` headers
```

README.md
```diff
@@ -1,349 +1,449 @@
...
 // see https://expressjs.com/en/guide/behind-proxies.html
 // app.set('trust proxy', 1);
 
 const apiLimiter = rateLimit({
-  windowMs: 15 * 60 * 1000, // 15 minutes
-  max: 100
-});
+	windowMs: 15 * 60 * 1000, // 15 minutes
+	max: 100, // Limit each IP to 100 requests per `window` (here, per 15 minutes)
+	standardHeaders: true, // Return rate limit info in the `RateLimit-*` headers
```
nfriedly, line 90:
```suggestion
	standardHeaders: true, // Return rate limit info in the `RateLimit-*` headers
	legacyHeaders: false, // Disable the `X-RateLimit-*` headers
```

README.md
```diff
@@ -1,349 +1,449 @@
...
 // app.set('trust proxy', 1);
 
 const apiLimiter = rateLimit({
-  windowMs: 15 * 60 * 1000, // 15 minutes
-  max: 100
-});
-app.use("/api/", apiLimiter);
+	windowMs: 15 * 60 * 1000, // 15 minutes
+	max: 100, // Limit each IP to 100 requests per `window` (here, per 15 minutes)
+	standardHeaders: true, // Return rate limit info in the `RateLimit-*` headers
```
nfriedly, line 109:
```suggestion
	standardHeaders: true, // Return rate limit info in the `RateLimit-*` headers
	legacyHeaders: false, // Disable the `X-RateLimit-*` headers
```

README.md
```diff
@@ -1,349 +1,449 @@
...
-    "Too many accounts created from this IP, please try again after an hour"
-});
-app.post("/create-account", createAccountLimiter, function(req, res) {
-  //...
-});
+	windowMs: 60 * 60 * 1000, // 1 hour
+	max: 5, // Limit each IP to 5 create account requests per `window` (here, per hour)
+	message:
+		'Too many accounts created from this IP, please try again after an hour',
+	standardHeaders: true, // Return rate limit info in the `RateLimit-*` headers
```
nfriedly, line 119:
```suggestion
	standardHeaders: true, // Return rate limit info in the `RateLimit-*` headers
	legacyHeaders: false, // Disable the `X-RateLimit-*` headers
```

README.md
```diff
@@ -1,349 +1,449 @@
...
-
-//  apply to all requests
-app.use(limiter);
+	windowMs: 15 * 60 * 1000, // 15 minutes
+	// `max` could also be an async function or return a promise
+	max: (request, response) => {
+		if (isPremium(request)) return 10
+		else return 5
+	},
+	standardHeaders: true, // Return rate limit info in the `RateLimit-*` headers
```
nfriedly, line 180: Lets omit the other config options here, the point of this is to provide an example of using the `max` option.

```suggestion
	// `max` could also be an async function or return a promise
	max: (request, response) => {
		if (isPremium(request)) return 10
		else return 5
	},
	// ...
```

nfriedly (changes requested):

config/tsconfig/cjs.json
nfriedly: Ok, then, yeah, I think we should 'nix the comments. Even if TypeScript allows it, there are other JSON parsers that would choke here.

gamemaker1: Ok, will do

source/lib.ts
```diff
@@ -0,0 +1,312 @@
...
+					) => {
+						if (error) reject(error)
+						resolve({ totalHits, resetTime })
+					},
+				)
+			})
+		}
+
+		async decrement(key: string): Promise<void> {
+			return Promise.resolve(legacyStore.decrement(key))
```
nfriedly, line 63: I *think* that marking it as `async` is enough, you don't have to make it explicitly return a promise. I could be wrong though. We should test this with a couple of external stores just to be safe.

gamemaker1: Yes, marking it `async` is enough, I usually mark it explicitly `async` so if we change the code in the future and remove the `return Promise...`, it still does return a `Promise`.

source/lib.ts
```diff
@@ -0,0 +1,312 @@
...
+			if (!request.ip) {
+				console.error(
+					'WARN | `express-rate-limit` | `request.ip` is undefined. You can avoid this by providing a custom `keyGenerator` function, but it may be indicative of a larger issue.',
+				)
+			}
+
+			return request.ip
+		},
+		skip: () => false,
+		handler: (_request, response, _optionsUsed: Options) =>
```
nfriedly, line 110:
```suggestion
		handler: (_request, response, _next, options) =>
```

The third argument is a `next()` function, although it isn't used here. The 4th is the options. (It was a late addition, so it was tacked onto the end for back-compat.)

Also, might as well use the passed-in options now that we have them. Make it easier for someone else to copy-paste this method as the basis for their custom version.

source/lib.ts
```diff
@@ -0,0 +1,312 @@
...
+		max: 5,
+		message: 'Too many requests, please try again later.',
+		statusCode: 429,
+		legacyHeaders: passedOptions.headers ?? true,
+		standardHeaders: passedOptions.draft_polli_ratelimit_headers ?? false,
+		requestPropertyName: 'rateLimit',
+		skipFailedRequests: false,
+		skipSuccessfulRequests: false,
+		requestWasSuccessful: (_request, response) => response.statusCode < 400,
+		keyGenerator: (request) => {
```
nfriedly, line 100:
```suggestion
		keyGenerator: (request, _response) => {
```

README.md
```diff
@@ -1,349 +1,456 @@
...
-The`req.rateLimit` object has `limit`, `current`, and `remaining` number of requests and, if the store provides it, a `resetTime` Date object.
+The `request.rateLimit` object has `limit`, `current`, and `remaining` number of
+requests and, if the store provides it, a `resetTime` Date object.
 
 Defaults to:
 
 ```js
-function (req, res, next, options) {
-    res.status(options.statusCode).send(options.message);
+const handler = (request, response, options) => {
```
nfriedly, line 284:
```suggestion
const handler = (request, response, _next, options) => {
```

source/lib.ts
```diff
@@ -0,0 +1,312 @@
...
+			}
+
+			// If the client has exceeded their rate limit, set the Retry-After
+			// header and call the {@link Options.handler} function
+			if (maxHits && totalHits > maxHits) {
+				if (options.legacyHeaders && !response.headersSent) {
+					response.setHeader('Retry-After', Math.ceil(options.windowMs / 1000))
+				}
+
+				options.handler(request, response, options)
```
nfriedly, line 295: We should keep the `next` param here. It allows end-users to decide *not* to handle the request. The param ordering is a bit of a wart, but I don't think it's worth breaking backwards compatibility over. 
```suggestion
				options.handler(request, response, next, options)
```

source/types.ts
```diff
@@ -0,0 +1,274 @@
...
+	 * Express request handler that sends back a response when a client is
+	 * rate-limited.
+	 */
+	readonly handler: RateLimitEventCallback
+
+	/**
+	 * Express request handler that sends back a response when a client has
+	 * reached their rate limit, and will be rate limited on their next request.
+	 */
+	readonly onLimitReached: RateLimitEventCallback
```
nfriedly, line 233: I think we'll have to split these into different interfaces to keep `next` in the handler callback. Sorry :/

nfriedly (changes requested):
A couple of things here:

* Add unused params as `_whatever` in all of the default methods.
* Restore the `next()` param in the handler function - sorry :(

source/types.ts
gamemaker1: No problem :D, will do

gamemaker1: Should we also pass the `next` param to the `onLimitReached` callback? For example, the user might want to skip setting a particular header if the `verbose` query param is `false`.

nfriedly: No, `onLimitReached` is kind of a weird function. 

If your limit is 3 and an end-user makes 10 requests, `handler` will be called 7 times, but `onLimitReached` will only be called once. 

It's intended for logging that a user hit the limit, without logging every single request after the limit was hit.

gamemaker1: > If your limit is 3 and an end-user makes 10 requests, handler will be called 7 times, but onLimitReached will only be called once.

Yes, I understand that.

The use case I was describing was the server returning a field warning the client that their rate limit has been reached in the response, e.g.:

```jsonc
// -- response body

{
	"meta": {
		"status": 200,
		"rateLimitReached": true,
	},
	"data": {
		// ...
	}
}
```

source/lib.ts
```diff
@@ -0,0 +1,312 @@
...
+			// Increment the client's hit counter by one
+			const { totalHits, resetTime } = await options.store.increment(key)
+
+			// Get the quota (max number of hits) for each client
+			const retrieveQuota =
+				typeof options.max === 'function'
+					? options.max(request, response)
+					: options.max
+
+			const maxHits = await Promise.resolve(retrieveQuota)
```
nfriedly, line 206: Do we still need `Promise.resolve(..)` here, or could it just be `await retrieveQuota`?

gamemaker1: I just checked and it could be just `await retrieveQuota`... I was surprised that worked. I have used `Promise.resolve` in all places just in case there is some weird edge though.

source/lib.ts
```diff
@@ -0,0 +1,312 @@
...
+					}
+				}
+
+				if (options.skipFailedRequests) {
+					response.on('finish', async () => {
+						if (!options.requestWasSuccessful(request, response))
+							await decrementKey()
+					})
+					response.on('close', async () => {
+						if (!response.writableEnded) await decrementKey()
```
nfriedly, line 267: I believe this was previously 

`if (!res.finished) {`

Why did it change to `writeEnabled`?

gamemaker1: It was deprecated in Node 13.4.0 - https://nodejs.org/api/http.html#http_request_finished

source/lib.ts
```diff
@@ -0,0 +1,312 @@
...
+			// Call the {@link Options.onLimitReached} callback if
+			// the client has reached their rate limit.
+			if (maxHits && totalHits === maxHits + 1) {
+				options.onLimitReached(request, response, options)
+			}
+
+			// If the client has exceeded their rate limit, set the Retry-After
+			// header and call the {@link Options.handler} function
+			if (maxHits && totalHits > maxHits) {
+				if (options.legacyHeaders && !response.headersSent) {
```
nfriedly, line 291: I think this should trigger for either legacy or standard headers. 
```suggestion
				if ((options.legacyHeaders || options.standardHeaders) && !response.headersSent) {
```

We should also call this out in the changelog and the readme...

changelog.md
```diff
@@ -0,0 +1,84 @@
...
+
+- Better Typescript support (the library was rewritten in Typescript).
+- Export the package as both ESM and CJS.
+- Publish the built package (`.tgz` file) on GitHub releases as well as the npm
+  registry.
+
+### Changed
+
+- Rename the `draft_polli_ratelimit_headers` option to `standardHeaders`.
+- Rename the `headers` option to `legacyHeaders`.
```
nfriedly, line 21:
```suggestion
- Rename the `headers` option to `legacyHeaders`.
- `Retry-After` header is now sent if either `legacyHeaders` or `standardHeaders` is set.
```

README.md
```diff
@@ -1,349 +1,488 @@
...
 
-### draft_polli_ratelimit_headers
+### `standardHeaders`
 
-Enable headers conforming to the [ratelimit standardization proposal](https://tools.ietf.org/id/draft-polli-ratelimit-headers-01.html): `RateLimit-Limit`, `RateLimit-Remaining`, and, if the store supports it, `RateLimit-Reset`. May be used in conjunction with, or instead of the `headers` option.
+Enable headers conforming to the
+[ratelimit standardization draft](https://github.com/ietf-wg-httpapi/ratelimit-headers/blob/main/draft-ietf-httpapi-ratelimit-headers.md)
+adopted by the IETF: `RateLimit-Limit`, `RateLimit-Remaining`, and, if the store
+supports it, `RateLimit-Reset`. May be used in conjunction with, or instead of
+the `legacyHeaders` option.
```
nfriedly, line 254:
```suggestion
the `legacyHeaders` option.

This setting also enables the `Retry-After` header when `max` is exceeded.
```

nfriedly (changes requested):

source/lib.ts
```diff
@@ -0,0 +1,312 @@
...
+				if (options.skipSuccessfulRequests) {
+					response.on('finish', async () => {
+						if (options.requestWasSuccessful(request, response))
+							await decrementKey()
+					})
+				}
+			}
+
+			// Call the {@link Options.onLimitReached} callback if
+			// the client has reached their rate limit.
```
nfriedly, line 283:
```suggestion
			// Call the {@link Options.onLimitReached} callback on
			// the first request where client exceeds their rate limit.
```

source/types.ts
```diff
@@ -0,0 +1,291 @@
...
+export type RateLimitExceededEventHandler = (
+	request: Express.Request,
+	response: Express.Response,
+	next: Express.NextFunction,
+	optionsUsed: Options,
+) => void
+
+/**
+ * Express request handler that sends back a response when a client has reached
+ * their rate limit, and will be rate limited on their next request.
```
nfriedly, line 51:
```suggestion
 * Event callback that is triggered on a client's first request that exceeds the limit
 * but not for subsequent requests. May be used for logging, etc. Should *not*
 * send a response.
```

source/lib.ts
nfriedly: Aah, good catch. Thanks!

package.json
```diff
@@ -1,53 +1,140 @@
...
+	"type": "module",
+	"files": [
+		"dist/",
+		"tsconfig.json",
+		"package.json",
+		"package-lock.json",
+		"readme.md",
+		"license.md",
+		"changelog.md"
+	],
```
nfriedly, line 48: `request.writeEnabled` was added in node 12.9 - https://nodejs.org/api/http.html#requestwritableended (see also https://docs.npmjs.com/cli/v6/configuring-npm/package-json#engines )

```suggestion
	],
	"engines" : { "node" : ">=12.9.0" },
```

test/headers-test.ts
```diff
@@ -0,0 +1,60 @@
...
+			rateLimit({
+				windowMs: 60 * 1000,
+				max: 5,
+				legacyHeaders: true,
+			}),
+		)
+		const expectedResetTimestamp = Math.ceil(
+			(Date.now() + 60 * 1000) / 1000,
+		).toString()
+		const resetRegexp = new RegExp(`^\\d+${expectedResetTimestamp.slice(-2)}$`)
```
nfriedly, line 22: Previously this had `\d\d` at the end, now it has `\d+` at the beginning - why did that change?

The intention was to match the timestamp, but ignore the last two digits to allow for a bit of wiggle room.

source/types.ts
nfriedly: OK, let me think about that a bit. I feel like something won't work, but I can't quite put my finger on it.

test/headers-test.ts
nfriedly: Oh, right, slice with a negative 2 gets just the last two characters:

```js
'1234567890'.slice(-2)
// > "90"
```

So, kind of the opposite of the goal here.

```suggestion
		const resetRegexp = new RegExp(`^${expectedResetTimestamp.substr(0, expectedResetTimestamp.length - 2)}\\d\\d$`)
```

nfriedly:
Op, sorry, I think I introduced a lint error somewhere in my suggestions.

nfriedly:
Alright, I need to step away for a little bit. I still haven't finished reviewing everything, but I think we're really close.

package.json
gamemaker1: We are using `response.writableEnded`, not `request.writeEnabled`

gamemaker1: `response.writableEnded` was added in Node 12.9 too 😆

source/types.ts
```diff
@@ -0,0 +1,292 @@
...
+	init?: (options: Options) => void
+
+	/**
+	 * Method to increment a client's hit counter.
+	 *
+	 * @param key {string} - The identifier for a client
+	 *
+	 * @returns {IncrementResponse} - The number of hits and reset time for that client
+	 */
+	increment: (key: string) => Promise<IncrementResponse> | IncrementResponse
```
nfriedly, line 140: I'm wondering if we should just make this and the other APIs async instead of having two return types. The MemoryStore is the only one that will ever be synchronous. What do you think?
```suggestion
	async increment: (key: string) => IncrementResponse
```

gamemaker1: The current type declaration (`Promise<T> | T`) allows the function to be both synchronous and asynchronous. We could remove the `| T` part to make it only asynchronous.

Also, we cannot add the `async` keyword to types - instead, an async function is declared by wrapping the return type with `Promise<...>` - https://stackoverflow.com/a/55890828.

nfriedly: Aah, ok, I hadn't realized that. Let's just make it always return a promise then.

gamemaker1: Ok, will do

gamemaker1: Done!

test/memory-store-test.ts
```diff
@@ -0,0 +1,111 @@
...
+		await store.increment(keyTwo)
+
+		setTimeout(async () => {
+			const { totalHits: totalHitsOne } = await store.increment(keyOne)
+			const { totalHits: totalHitsTwo } = await store.increment(keyTwo)
+
+			expect(totalHitsOne).toEqual(1)
+			expect(totalHitsTwo).toEqual(1)
+		}, 60)
+	})
```
nfriedly, line 74: Without the `done` callback, or wrapping it in a promise and `await`ing it, the code in the setTimeout callback will execute after the test completes. So, if it fails, the test will already have been marked as passing and Jest won't know which test to attribute the failure to. 

Making setTimeout play nice with async/await is kind of a PITA, but fortunately Jest provides a clock-mock'ing mechanism. This is actually better in every sense - it makes the test simpler, faster, and more reliable. 

The only downside is that the setup and teardown should be outside of the test, so I created a new `describe` block to wrap it all. 

```suggestion
	describe('reset time', () => {
		beforeAll(() => {
			jest.useFakeTimers('modern')
		  })
		  afterAll(() => {
			jest.useRealTimers()
		  })
		  it('resets the count for all the keys in the store when the timeout is reached', async () => {
			const store = new MemoryStore()
			store.init({ windowMs: 50 } as Options)
			const keyOne = 'test-store-one'
			const keyTwo = 'test-store-two'
	
			await store.increment(keyOne)
			await store.increment(keyTwo)
	
			jest.advanceTimersByTime(60)

			const { totalHits: totalHitsOne } = await store.increment(keyOne)
			const { totalHits: totalHitsTwo } = await store.increment(keyTwo)

			expect(totalHitsOne).toEqual(1)
			expect(totalHitsTwo).toEqual(1)
		})
	})
```

test/memory-store-test.ts
```diff
@@ -0,0 +1,111 @@
...
+		const key = 'test-store'
+
+		try {
+			const { totalHits } = await store.increment(key)
+			expect(totalHits).toEqual(1)
+		} finally {
+			// @ts-expect-error `realTimeoutId` is already set in the `spyOn` call
+			clearTimeout(realTimeoutId)
+		}
+	})
```
nfriedly, line 117: Does Jest automatically undo `spyOn` mocks at the end of the test? I think you needed to call `jest.restoreAllMocks()` - but it should ideally be outside of the test, in an after* method so that it will run even if the test fails. I suppose in this case we could just do `const spy = jest.spyOn(global, 'setTimeout')...` and then do `spy.mockRestore()` in the `finally` block, although that makes me a bit nervous that something could go wrong outside of the `try`.

Also, having that `clearTimeout` inside the test also makes me nervous, although, come to think of it, none of the other tests even attempt to clean up after their memory stores...

test/memory-store-test.ts
gamemaker1: Done, good catch!

test/memory-store-test.ts
gamemaker1: Good catch, we should be calling `restoreAllMocks` in the `finally` block.

gamemaker1: > none of the other tests even attempt to clean up after their memory stores

How would we go about that?

test/middleware-test.ts
```diff
@@ -0,0 +1,710 @@
+// /test/middleware-test.ts
+// Tests the rate limiting middleware
+
+import { jest } from '@jest/globals'
+import Express from 'express'
+import Sinon, { SinonFakeTimers } from 'sinon'
```
nfriedly: We should be able to replace this with the Jest equivalents - https://jestjs.io/docs/jest-object#mock-timers 

Although, I suppose if this working, maybe we should just leave it for now.

gamemaker1: I just changed that as well as part of https://github.com/nfriedly/express-rate-limit/pull/263#discussion_r774300344

test/middleware-test.ts
```diff
@@ -0,0 +1,710 @@
...
+const sandbox = Sinon.createSandbox()
+
+describe('middleware test', () => {
+	let clock: SinonFakeTimers
+	beforeEach(() => {
+		clock = Sinon.useFakeTimers()
+	})
+	afterEach(() => {
+		clock.restore()
+		sandbox.restore()
```
nfriedly: since you're doing `jest.spyOn` instead of `sandbox.stub` in some places below, we need to add a `jest.restoreAllMocks()` here.
```suggestion
		sandbox.restore()
		jest.restoreAllMocks()
```

gamemaker1: Should I replace the `spy.mockRestore` call in the `finally` block with this instead?

test/memory-store-test.ts
nfriedly: > How would we go about that?

That's a good question... there's not really a good way to do it right now, although I suppose using fake timers for every test and then resetting things to real timers would do the trick.

test/middleware-test.ts
nfriedly: Yeah, I'd kind of prefer that.

nfriedly:
Ok, I think I am spent for tonight. Still gotta finish reading through the tests, I think I'll be able to do that tomorrow. 

Thanks for all the effort you're putting into this, have a great evening!

source/types.ts
gamemaker1: I just realised that if we force all the methods (especially `decrement`, `resetKey` and `resetAll`) to return a `Promise`, then other stores won't be able to maintain backwards compatibility - all the current functions are sync and return just `void`.

gamemaker1:
Note to self:

- add test for `promisifyStore` function
- fix the (currently commented) hanging middleware test

source/types.ts
nfriedly: aah, shucks. You're right.

There might be a way to make it work, 5.x doesn't care about the return type, let me think about it a bit...

gamemaker1:
Testing other stores:

- [ ] [rate-limit-redis](https://github.com/wyattjoh/rate-limit-redis): Tried it out, but encountered a few type errors. Need to try again.
- [ ] [rate-limit-memcached](https://github.com/linyows/rate-limit-memcached): --
- [ ] [rate-limit-mongo](https://github.com/2do2go/rate-limit-mongo): --
- [ ] [precise-memory-rate-limit](https://github.com/Nebulizer1213/precise-memory-rate-limit): --

test/middleware-test.ts
```diff
@@ -0,0 +1,696 @@
...
+		expect(store.resetKeyWasCalled).toEqual(true)
+	})
+
+	describe('timer tests', () => {
+		beforeEach(() => {
+			jest.useFakeTimers('modern')
+		})
+		afterEach(() => {
+			jest.useRealTimers()
+		})
```
nfriedly, line 114: I think we might as well move these to the top-level `describe` block - the clock was mocked for every test in this file previously, and while it doesn't make any difference for the ones with the MockStore, it prevents leaking `setInterval`s for the ones that use the "real" MemoryStore.

nfriedly: (We should probably also do that for the MemoryStore tests...)

test/middleware-test.ts
```diff
@@ -0,0 +1,696 @@
...
+
+		expect(store.initWasCalled).toEqual(true)
+	})
+
+	it('should print an error if `request.ip` is undefined', async () => {
+		jest.spyOn(global.console, 'error').mockImplementation(() => {})
+
+		await Promise.resolve(
+			rateLimit()(
+				{ ip: undefined } as any as Express.Request,
```
nfriedly, line 70: I didn't know you could do `as` twice in a row. Cool.

gamemaker1: It's usually used when Typescript puts it foot down and says 'you may not!' 😆

gamemaker1: I usually just suppress the errors with a `@ts-expect-error` comment if its a test though

gamemaker1:
> There might be a way to make it work, 5.x doesn't care about the return type, let me think about it a bit...

I've restored the dual return type while I test it with the other stores, let me know what you think

test/middleware-test.ts
gamemaker1: Done

nfriedly:
> I've restored the dual return type while I test it with the other stores, let me know what you think

Yea, we should probably document it, but I think that's the right call. I'm thinking maybe I should go and make an external store, maybe for sqlite or the cluster module, to exercise the API a bit and provide a good example for how to build a store that's compatible with 5.x and 6.x

test/middleware-test.ts
```diff
@@ -0,0 +1,696 @@
...
+			}),
+		)
+
+		await request(app).get('/').expect(200)
+		await request(app).get('/').expect(200)
+		await request(app).get('/').expect(429).expect(message)
+	})
+
+	it('should allow the error status code to be customized', async () => {
+		const statusCode = 420
```
nfriedly, line 174: lol

changelog.md
```diff
@@ -0,0 +1,87 @@
...
+  - Rename the `incr` method to `increment`.
+  - Allow the `increment`, `decrement`, `resetKey` and `resetAll` methods to
+    return a promise.
+  - Old stores will automatically be promisified and used.
+- The package can now only be used with NodeJS version 12.9.0 or greater.
+
+### Removed
+
+- Remove the deprecated `limiter.resetIp` method.
+- Remove the deprecated options `delayMs`, `delayAfter` and `global`.
```
nfriedly, line 36:
```suggestion
- Remove the deprecated `limiter.resetIp` method. (Use `limiter.resetKey` instead.)
- Remove the deprecated options `delayMs`, `delayAfter` and `global`. (Use the `express-slow-down` package for delay, and a custom `keyGenerator` that always returns the same key in place of the `global` option.)
```

test/middleware-test.ts
```diff
@@ -0,0 +1,696 @@
...
+		await request(app).get('/').expect(200)
+
+		expect(store.decrementWasCalled).toEqual(false)
+	})
+
+	// FIXME: Started hanging after I added the `useFakeTimers` call to `beforeAll`
+	/*
+	it('should decrement hits when response closes and `skipFailedRequests` is set to true', async () => {
+		jest.setTimeout(50_000)
+
```
nfriedly, line 437: This one requires "real" timers- I re-enabled them in the JS version:

https://github.com/nfriedly/express-rate-limit/blob/b9e0e59d1b65621d92827ec50208420936114106/test/express-rate-limit-test.js#L486-L487

nfriedly: (It's OK to call `jest.useRealTimers()` twice)

nfriedly:
Ok, I have finally looked through everything. I noticed that a lot of the comments I had added to JS unit tests didn't make it through to the TS versions - it's not the end of the world, but some of them would be nice to have. 

Beyond that, I think we just need to finish up the manual testing and then this should be good to go.

Thank you for all of your work!

gamemaker1:
> I noticed that a lot of the comments I had added to JS unit tests didn't make it through to the TS versions - it's not the end of the world, but some of them would be nice to have.

Hmmm let me take a look at the js versions again and try to add them back - comments are always good to have

gamemaker1:
Also, what do you think about this? - https://github.com/nfriedly/express-rate-limit/pull/263#discussion_r773957618

source/types.ts
nfriedly: I figured out what was throwing me off: It sounds like you're wanting to give extra info on the last request that's allowed through - the 3rd request if the limit is 3.  Is that right?

However, `onLimitReached` is triggered on the 4th request. So, it's too late for a warning, the user is already blocked at that point. They won't get the `data`, just whatever the express-rate-limit `handler` function returns.

I think your use-case would be better served by checking the `req.rateLimit` property after `express-rate-limit` has done it's thing. 

I think `onLimitReached` predates `req.rateLimit` and, honestly, it should probably just be deprecated. It's very confusing and not particularly helpful. Everything that can be done with `onLimitReached` can also be done with a custom `handler` and an `if` statement.

Or, were you saying the 4th request should actually be allowed through, just with a warning? In that case, you could achieve the same effect by doing it in `handler` but bumping the limit up to `4`.

(To be clear, I think we should keep it in there with the same behavior for now, but just mark it as deprecated try to discourage people from using it.) 

A better replacement might be to return an EventEmitter and have some kind of event for this, but I'm not convinced it's actually needed at all. (Also, I'm not entirely sure that a express middleware style function can also be an EventEmitter...)

gamemaker1: > I figured out what was throwing me off: It sounds like you're wanting to give extra info on the last request that's allowed through - the 3rd request if the limit is 3. Is that right?

Yes.

> However, onLimitReached is triggered on the 4th request

Ohk, I didn't realise that 🤦🏾 

> I think your use-case would be better served by checking the req.rateLimit property after express-rate-limit has done it's thing.

👍🏾 

> A better replacement might be to return an EventEmitter and have some kind of event for this, but I'm not convinced it's actually needed at all. (Also, I'm not entirely sure that a express middleware style function can also be an EventEmitter...)

An `EventEmmitter` would make it reallyy complicated 😅 , and as you mentioned, this kind of use case is already covered in the `handler` function. 

> I think we should keep it in there with the same behavior for now, but just mark it as deprecated try to discourage people from using it

So I'll add a `@deprecated` tag to the `onLimitReached` method but the middleware will still continue to support it.

gamemaker1:
Noooo... why did this fail only on this run? - https://github.com/nfriedly/express-rate-limit/runs/4623999171?check_suite_focus=true

nfriedly:
> Noooo... why did this fail only on this run? - https://github.com/nfriedly/express-rate-limit/runs/4623999171?check_suite_focus=true

Ugh, that test. That's the one that we can't mock the clock because we have to wait for underlying things to actually time out. We could try something like `jest.setTimeout(3 * 60 * 1000)` to give it a bigger window. 

I think I'm done for tonight, though. I'll talk to you tomorrow.

gamemaker1:
Created https://github.com/gamemaker1/express-rate-limit-store-tests to test the changes we make here against the external stores listed in the readme.

gamemaker1:
Added a contributing guide as well as issue/pr templates

nfriedly (approved):
Nice work. This is frankly amazing.  I think this is good to go - is there anything else you wanted to add before I merge it?

nfriedly:
Ugh, that one test. It seems to only fail on macOS; maybe we should just disable it there?

nfriedly:
Also, I feel like this is as much your code as it is mine at this point - would you like to be a maintainer here and have commit access to the repo?

gamemaker1:
> Is there anything else you wanted to add before I merge it?

No, this looks great!

>  It seems to only fail on macOS; maybe we should just disable it there?

Oh that's weird, but I guess we can disable it for now. Not sure how to do that, will look it up.

> Would you like to be a maintainer here and have commit access to the repo?

Yes, I would love that. It's been great working on this with you, looking forward to more <3

gamemaker1:
Also, should I make some follow up PRs to other external stores to help them migrate to the newer `Store` interface?

[They all still work](https://github.com/nfriedly/express-rate-limit/pull/263#issuecomment-1000357635), but it will be easier for users to use external stores that use the `init` method.

nfriedly:
> > It seems to only fail on macOS; maybe we should just disable it there?
> 
> Oh that's weird, but I guess we can disable it for now. Not sure how to do that, will look it up.

In the test, check if [os.platform()](https://nodejs.org/dist/latest-v16.x/docs/api/os.html#osplatform) is `darwin` and return early. Add a comment about it being flakey on macOS.

nfriedly:
Ugh, and now that test is failing on windows. I'm just going to go back to having it commented out for now.

nfriedly:
Oh, and normally I don't do the package.json version change in the PR - I merge it into master first and then use the [npm version](https://docs.npmjs.com/cli/v6/commands/npm-version) command on master to both update package.json and create a git tag in one go.  But I didn't notice that before merging, so I just created the git tag manually this time. If the CI publish works, it should be live on npm shortly!

gamemaker1:
> Then use the npm version command on master to both update package.json and create a git tag in one go.

I usually do that too, I think I changed it when I was testing the external stores though.

> If the CI publish works, it should be live on npm shortly!

Yay!

nfriedly:
It's live! https://www.npmjs.com/package/express-rate-limit 

Thanks again!

Issue #269, opened 2021-12-25
Error: Cannot find module 'express-rate-limit'


andreidubov:
### Description

After upgrading to a new version, I get the following error:
`Error: Cannot find module 'express-rate-limit'`. 

Here is the import:
`import rateLimit, { Options, RateLimitRequestHandler } from "express-rate-limit";`

### Library version

6.0.1

### Node version

v12.15.0

### Typescript version (if you are using it)

4.2.3

### Module system

ESM

Issue #268, opened 2021-12-25
How to import this into Typescript


krishnaTORQUE:
### Description

How to import this into typescript?

None of those below working

`import rateLimit from express-rate-limit''`

`const rateLimit = require('express-rate-limit')`

`const rateLimit = (await import('express-rate-limit')).default`

### Library version

6.0.0

### Node version

16.13.1

### Typescript version (if you are using it)

4.5.4

### Module system

CommonJS

gamemaker1:
Hi @krishnaTORQUE,

Could you please share the error you are getting when you compile the code? (run `npx tsc`)

The following should work while using the library with Typescript:

```
import rateLimit from 'express-rate-limit'
```

krishnaTORQUE:
`Error [ERR_REQUIRE_ESM]: require() of ES Module`

gamemaker1:
Could you try the following instead:

```ts
const rateLimit = require('express-rate-limit/dist/cjs/index.js')

// OR

import rateLimit from 'express-rate-limit/dist/cjs/index.js'
```

gamemaker1:
Also, could you please paste your `tsconfig.json`?

krishnaTORQUE:
```json
{
  "extends": "ts-node/node16/tsconfig.json",
  "ts-node": {
    "transpileOnly": false,
    "files": true,
    "compilerOptions": {
      "module": "commonjs"
    }
  },
  "compilerOptions": {
    "target": "es5",
    "module": "commonjs",
    "lib": ["esnext"],
    "jsx": "react",
    "moduleResolution": "node",
    "allowJs": true,
    "checkJs": true,
    "noImplicitAny": true,
    "noImplicitReturns": true,
    "strictNullChecks": true,
    "strictFunctionTypes": true,
    "noUnusedLocals": true,
    "noUnusedParameters": true,
    "removeComments": true,
    "strict": true,
    "noFallthroughCasesInSwitch": true,
    "allowSyntheticDefaultImports": true,
    "esModuleInterop": true,
    "skipLibCheck": true,
    "forceConsistentCasingInFileNames": true,
    "resolveJsonModule": true,
    "isolatedModules": true,
    "importHelpers": true,
    "outDir": "buildServer",
    "rootDirs": ["src"]
  },
  "exclude": ["node_modules"]
}

```

krishnaTORQUE:
> Could you please try the following instead:
> 
> ```ts
[...]

Issue #267, opened 2021-12-25
'express-rate-limit' does not provide an export named 'default' when running with jest with --experimental-vm-modules


soruly:
### Description

```
$ npm run jest

> trace.moe-api@1.0.0 jest
> NODE_NO_WARNINGS=1 NODE_OPTIONS=--experimental-vm-modules jest --runInBand

Determining test suites to run...
 FAIL  src/get-status.test.js
  ● Test suite failed to run

    SyntaxError: The requested module 'express-rate-limit' does not provide an export named 'default'

      at Runtime.linkAndEvaluateModule (node_modules/jest-runtime/build/index.js:779:5)
```

This has been working for express-rate-limit@5.5.1 with nodejs and jest, and working with 6.0.0 with nodejs, but not with jest

Maybe jest (27.4.5) doesn't fully support ESM yet. Running with node is perfectly fine.

This test are written in ESM and requires `--experimental-vm-modules` to run with jest, because it doesn't transpile anything with babel.

The test case: https://github.com/soruly/trace.moe-api/blob/master/src/app.js

### Library version

6.0.0

### Node version

v16.13.0

### Module system

ESM

gamemaker1:
Hi @soruly,

This is [a known issue](https://github.com/facebook/jest/issues/9771) with Jest (it does not support the `exports` field in `package.json`). A similar issue exists on the Jest repo - https://github.com/facebook/jest/issues/11742).

There is [a PR to fix it](https://github.com/facebook/jest/pull/11961) that [might be part of Jest v28](https://github.com/facebook/jest/pull/11961#discussion_r758162104), but until then you can use the workaround described [here](https://github.com/facebook/jest/pull/11961#discussion_r748903736) (basically, just delete the `main` field from `node_modules/express-rate-limit/package.json` before running the test and restore it after).

Hope this helps.

soruly:
Thanks, I'll stay in old version until then ;)

gamemaker1:
Hi @soruly,

Could you try running these tests again with v6.0.1 - they seem to run fine in [this example project](https://github.com/gamemaker1/test/tree/express-rate-limit-hotfix/repro/ts-esm).

Thanks!

soruly:
Hi @gamemaker1

Confirmed working with v6.0.1 without any modifications, thanks :D

PR #272, opened 2021-12-28
meta: use esbuild for builds and add external tests

gamemaker1:
## Related Issues

#270, #269, #268, #267

### Build process

- Use [`esbuild`](https://esbuild.github.io/) to generate ESM and CJS output.
- Use [`dts-bundle-generator`](https://github.com/timocov/dts-bundle-generator) to generate Typescript types.

### Tests

- Add additional tests that test:
	- importing the library in `js-cjs`, `js-esm`, `ts-cjs`, `ts-esm` environments
	- usage of the library with external stores

## Checklist

      added for the same.

---

@nfriedly what do you think?


scripts/make.js
```diff
@@ -0,0 +1,146 @@
+#!/usr/bin/env node
+
+// /scripts/make
+// ShellJS 'makefile' to run various tasks
+
+import Process from 'node:process'
+
+import 'shelljs/make.js'
+import Shell from 'shelljs'
+import Chalk from 'chalk'
```
nfriedly, line 10: What does this gain us over `npm run ...` scripts? I see that there's logging, but are there any other benefits?

One thing I like about the original `make` command is that it understood dependencies, and so it would automatically skip recompiling things when the source hadn't changed, but I don't think we're getting that here (?)

(Also, this thing should compile fast enough that it wouldn't really matter here...)

.github/workflows/test.yaml
```diff
@@ -1,36 +1,115 @@
...
+    branches:
+      - 'master'
+  pull_request:
 
 jobs:
-  test:
-    name: Build, Lint and Test
+  build_library:
+    name: Build
+    runs-on: ubuntu-latest
```
nfriedly, line 15: I like this idea of just building once for all of the tests!

.github/workflows/test.yaml
```diff
@@ -1,36 +1,115 @@
...
+        uses: actions/checkout@v2
+      - name: Use Node ${{ matrix.node-version }}
+        uses: actions/setup-node@v1
+        with:
+          node-version: ${{ matrix.node-version }}
+      - name: Test various imports with the library
+        run: |
+          git checkout origin/tests
+          cd test/import-tests
+          bash run-all-tests
```
nfriedly, line 79: Why not include these tests in the main branch?

.github/workflows/release.yaml
```diff
@@ -35,9 +33,10 @@ jobs:
       - name: Upload coverage report
         uses: actions/upload-artifact@v2
         with:
-          name: Coverage Report
+          name:
+            Coverage Report (Node ${{ matrix.node-version }} on ${{ matrix.os
+            }})
```
nfriedly, line 38: Kind of a nitpick, but I'd prefer this to be all on one line:

```suggestion
          name: Coverage Report (Node ${{ matrix.node-version }} on ${{ matrix.os }})
```

nfriedly (commented):
I'm somewhat inclined to just give this a 👍 since it fixes the issue, but there's a few things that seem overly complex, or at least I don't understand the reasoning...

scripts/make.js
gamemaker1: You are right, it is a bit of a complication. The reason I introduced it was because I was appending stuff to the built files and it was getting a bit messy - this is largely fixed with the usage of esbuild.

Reverted to good old npm scripts!

gamemaker1: > (Also, this thing should compile fast enough that it wouldn't really matter here...)

ESBuild is indeed blazingly fast - it takes just 2 ms to build the library! 🎉

.github/workflows/test.yaml
gamemaker1: Sure, I'll work on that

gamemaker1: Done, but need to make the external store tests use jest as a test runner. Also, switched the tests to run on only ubuntu, as running docker services is not supported on windows and macos 😕

gamemaker1: All tests use jest now! 🚀

gamemaker1:
@nfriedly I made some changes, let me know what you think

gamemaker1:
Also, I have been labelling issues and closing stale ones, feel free to reopen or change the labels if you feel I made a mistake.

test/external/run-all-tests
```diff
@@ -0,0 +1,36 @@
...
+set -e 
+set -o pipefail
+
+# Tests all the variations (JS-TS and CJS-ESM) with the `express-rate-limit` package.
+cd imports/
+for dir in `ls -d */`; do
+	cd $dir
+
+	rm -rf node_modules
+	npm install ../../../../*.tgz
```
nfriedly, line 14: What is this?

I assume it's the node_modules folder.. but why is it in a .tgz file? Maybe add a comment clarifying that?

.github/workflows/test.yaml
```diff
@@ -1,36 +1,38 @@
...
-    name: Build, Lint and Test
+    name: Test
     strategy:
       matrix:
-        node-version: [12.x, 14.x, 15.x, 16.x, 17.x]
-        os: [ubuntu-latest, macos-latest, windows-latest]
-    runs-on: ${{ matrix.os }}
-
+        node-version: [14.x, 15.x, 16.x, 17.x]
+    runs-on: ubuntu-latest
```
nfriedly, line 14: Do you think we could run at least `npm run test:lib` on mac and windows? Maybe just on node 16.*. I think there's a way to add extra jobs to the matrix, but I don't remember what it is off the top of my head.

It'd be really nice if we could just define this once and then make the release.yml depend on the test.yml, but I don't think github supports that...

nfriedly (commented):
Thanks! I haven't finished looking through this but I left a couple of questions below.

test/external/run-all-tests
gamemaker1: Basically all of the tests in the `test/external` folder are example 'projects', that have `express-rate-limit` as a dependency. When we run `npm pack`, it builds a `.tgz` file that we install as a dependency in all the example projects, using `npm install <path-to-tgz>`. The file name contains the version number, so to allow the script to work no matter what the version, i use a `*.tgz` glob. I guess I could make it more specific and change it to `express-rate-limit-*.tgz`, and add this as a comment in that script.

.github/workflows/test.yaml
gamemaker1: Maybe we can have only one workflow file, `ci.yaml`, with the following jobs:

- `build` - runs `npm run compile`, makes sure that the code is valid typescript on node 14 to 17, and on linux, windows and macos
- `lint` - run `npm run lint`, makes sure there are no linting or formatting errors on node 14 to 17, and on linux, windows and macos
- `test_library`: run `npm run test:lib`, runs the tests for the library on node 14 to 17, and on linux, windows and macos
- `test_external`: run `npm run test: ext`, runs the tests that checks if the imports and external stores work on node 14 to 17, and on linux only (because the external store tests need docker containers running, this is supported by github actions only on linux runners)
- `release` - runs only if a tag is pushed (`if: startsWith(github.ref, 'refs/tags/v')`), needs the all the above jobs to succeed, and publishes the `npm pack`ed package to the registry. Also creates a new release.

nfriedly (approved):
On second thought, I think we should merge the in as-is, ship the fix, and then come back to CI stuff after that.

nfriedly:
Oh, and make sure to remove the `.default` stuff from the readme now.

gamemaker1:
Sure, will do

gamemaker1:
Merged 🎉

gamemaker1:
Oops - I rebased and merged instead of squash and merge 🤦🏾

gamemaker1:
Ok, the CI went mad on a mistake in the release workflow - I'm fixing that bug, manually merging this, and then publishing 6.0.2. Sorry about this 😢

.github/workflows/test.yaml
gamemaker1: Added this in 32202bae817ef4e23d3294c96f4d4e6c61f22d6d 🚀

test/external/run-all-tests
gamemaker1: Done in b548dc8

PR #279, opened 2022-01-02
meta: add `main` field to `package.json` and run eslint tests along with import tests

### LINKED ISSUE #277: ESLint not able to resolve import - `Error: unable to resolve path to module "express-rate-limit"`
### Description

I installed it with:

```
npm install express-rate-limit
```

and imported it as:

```js
const rateLimit = require('express-rate-limit');
```

I'm using ESLint and it errors on the fact that there is no "/node_modules/express-rate-limit/dist/index" module. ESLint probably looks for the plain javascript format but there is no `index.js`, only `index.cjs`, `index.d.ts` and `index.mjs`.

image

### Library version

6.0.3

### Node version

14.13.0

_No response_

### Module system

CommonJS

gamemaker1:
## Related Issues

- Fixes #277

### Tests

- Run `eslint` as part of the import tests too.

### Build

- Add the `main` and `module` fields to `package.json` so ESLint does not complain

## Checklist

      added for the same.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,13 +29,10 @@
 	],
 	"type": "module",
 	"exports": {
-		".": [
-			{
-				"import": "./dist/index.mjs",
-				"require": "./dist/index.cjs"
-			},
-			"./dist/index.cjs"
-		]
+		".": {
+			"import": "./dist/index.mjs",
+			"require": "./dist/index.cjs"
+		}
 	},
 	"main": "./dist/index.cjs",
 	"module": "./dist/index.mjs",
--- a/test/external/imports/js-cjs/package.json
+++ b/test/external/imports/js-cjs/package.json
@@ -4,6 +4,7 @@
 	"description": "A minimal example (JS-CJS) of a project using the express-rate-limit package.",
 	"scripts": {
 		"start": "node source/index.js",
+		"lint": "eslint source/**/*.js",
 		"test": "jest"
 	},
 	"dependencies": {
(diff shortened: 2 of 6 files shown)


gamemaker1:
Hmmm... the ESLint errors for the `ts-cjs` and `ts-esm` import tests saying it can't find the `@typescript-eslint/eslint-plugin` package

gamemaker1:
That's weird - `@typescript-eslint/eslint-plugin` is listed as a dev dependency in both the projects 🤔

nfriedly:
> That's weird - @typescript-eslint/eslint-plugin is listed as a dev dependency in both the projects 🤔

I think I see why: we're doing `npm install` for the external stores tests:

https://github.com/nfriedly/express-rate-limit/blob/5a16de18f1b2642c73384c7492dd641588187dcb/test/external/run-all-tests#L45

But we're not for the the ts-js and cjs-esm tests:

https://github.com/nfriedly/express-rate-limit/blob/5a16de18f1b2642c73384c7492dd641588187dcb/test/external/run-all-tests#L14-L18

For those, we do [`npm ci` and `npm pack` on the top level](https://github.com/nfriedly/express-rate-limit/blob/2333f9847170a5d11382f89340fdae4d72086298/.github/workflows/ci.yaml#L74-L75) (where @typescript-eslint/eslint-plugin is [*not* a dev dependency](https://github.com/nfriedly/express-rate-limit/blob/5a16de18f1b2642c73384c7492dd641588187dcb/package.json#L75-L94)) and just copy that down. Since @typescript-eslint/eslint-plugin is not a top-level dependency, it doesn't get installed anywhere in that process.

I'm going to try adding an `npm install` step to the ts/js/cjs/esm tests and see if that fixes it.

nfriedly:
Yep, that did the trick. We should probably figure out how to use `npm ci` in there instead of `npm install`, but I think this is good enough for now. Lets ship 6.0.4.

package.json
```diff
@@ -29,13 +29,10 @@
 	],
 	"type": "module",
 	"exports": {
-		".": [
-			{
-				"import": "./dist/index.mjs",
-				"require": "./dist/index.cjs"
-			},
-			"./dist/index.cjs"
-		]
```
nfriedly, line 38: Oh, which variation of this were folks in #277 testing?

gamemaker1: Yes they tested it with the array but it seems that doesn't have any effect if we already explicitly mention `import` and `require`

nfriedly: Eh, I guess it shouldn't matter if the entire issue was that eslint was ignoring this field.

PR #282, opened 2022-01-05
fix: use named imports for express types

### LINKED ISSUE #281: index.d.ts: import Express from 'express'; error
### Description

After upgrading from v5 to v6, during TS compilation I've this error coming from index.d.ts of express-rate-limit (that seems autogenerated):

```
node_modules/express-rate-limit/dist/index.d.ts:3:8 - error TS1259: Module '"/home/dave/Job/myproject/node_modules/@types/express/index"' can only be default-imported using the 'esModuleInterop' flag

3 import Express from 'express';
         ~~~~~~~

  node_modules/@types/express/index.d.ts:133:1
    133 export = e;
        ~~~~~~~~~~~
    This module is declared with using 'export =', and can only be used with a default import when using the 'esModuleInterop' flag.


Found 1 error.
```

First lines of index.d.ts:
```
// Generated by dts-bundle-generator v6.2.0

import Express from 'express';
```

### Library version

6.0.4

### Node version

16.13.1

4.3.5

### Module system

CommonJS

gamemaker1:
## Related Issues

- Fixes #281

### Changed

- Use named imports for Express types.

### Removed

- Remove additional options from `tsconfig.json`s in the import tests.

## Caveats/Problems

- Still had to keep the `esModuleInterop` flag in the tests as the `express` import doesn't work otherwise.

## Checklist

      added for the same.


test/external/imports/ts-cjs/tsconfig.json
```diff
@@ -4,12 +4,11 @@
 	"compilerOptions": {
 		"target": "esnext",
 		"module": "commonjs",
-		"declaration": true,
+		"declaration": false,
```
nfriedly, line 7: `false` is the default, right? so, we can probably just omit this line now.

test/external/imports/ts-cjs/tsconfig.json
```diff
@@ -4,12 +4,11 @@
...
 		"sourceMap": false,
-		"outDir": "build/",
 		"strict": true,
-		"moduleResolution": "node",
 		"esModuleInterop": true,
```
nfriedly, line 10: what happens if we turn off `esModuleInterop`?

test/external/imports/ts-cjs/tsconfig.json
```diff
@@ -4,12 +4,11 @@
...
-		"skipLibCheck": true
+		"moduleResolution": "node",
```
nfriedly, line 11: this is the default also, right?

nfriedly (approved):
It seems like there's a few different things going on here, but if this fixes #281 then I think it's good.

package.json
```diff
@@ -73,19 +73,19 @@
 		"express": "^4"
 	},
 	"devDependencies": {
-		"@jest/globals": "^27.4.2",
+		"@jest/globals": "^27.4.6",
 		"@types/express": "^4.17.13",
-		"@types/jest": "^27.0.3",
-		"@types/node": "^16.11.17",
+		"@types/jest": "^27.4.0",
+		"@types/node": "^17.0.8",
```
DavideViolante, line 79: I would not use a non LTS node version here

test/external/imports/ts-cjs/tsconfig.json
```diff
@@ -4,12 +4,11 @@
 	"compilerOptions": {
 		"target": "esnext",
 		"module": "commonjs",
-		"declaration": true,
+		"declaration": false,
 		"sourceMap": false,
```
DavideViolante, line 8: Default is [false](https://www.typescriptlang.org/tsconfig#sourceMap), this line can be deleted

test/external/imports/ts-cjs/tsconfig.json
```diff
@@ -4,12 +4,11 @@
...
-		"outDir": "build/",
```
DavideViolante, line 9: Idk if this is correct, generated files now won't go inside this dir.

test/external/imports/ts-esm/tsconfig.json
```diff
@@ -4,12 +4,11 @@
 	"compilerOptions": {
 		"target": "esnext",
 		"module": "esnext",
-		"declaration": true,
+		"declaration": false,
 		"sourceMap": false,
```
DavideViolante, line 8: As above, this is false by default, can be removed.

test/external/stores/tsconfig.json
```diff
@@ -4,12 +4,10 @@
 	"compilerOptions": {
 		"target": "esnext",
 		"module": "esnext",
-		"declaration": true,
+		"declaration": false,
 		"sourceMap": false,
```
DavideViolante, line 8: Same

test/external/imports/ts-cjs/tsconfig.json
gamemaker1: `classic` is the default - https://www.typescriptlang.org/tsconfig#moduleResolution

package.json
gamemaker1: Arghhh that's the second time that's happened - I bump all the deps and this gets bumped too 🤦🏾 Maybe I should pin that to 16.x instead, until the next LTS version.

test/external/imports/ts-cjs/tsconfig.json
gamemaker1: Will do

test/external/imports/ts-cjs/tsconfig.json
gamemaker1: We get the following when running the `ts-cjs` test:

```
    test/server-test.ts:4:8 - error TS1259: Module '"/data/code/contrib/express-rate-limit/core/test/external/imports/ts-cjs/node_modules/@types/supertest/index"' can only be default-imported using the 'esModuleInterop' flag

    4 import request from 'supertest'
             ~~~~~~~

      node_modules/@types/supertest/index.d.ts:10:1
        10 export = supertest;
           ~~~~~~~~~~~~~~~~~~~
        This module is declared with using 'export =', and can only be used with a default import when using the 'esModuleInterop' flag.
    test/server-test.ts:4:8 - error TS1259: Module '"/data/code/contrib/express-rate-limit/core/test/external/imports/ts-cjs/node_modules/@types/supertest/index"' can only be default-imported using the 'esModuleInterop' flag

    4 import request from 'supertest'
             ~~~~~~~

      node_modules/@types/supertest/index.d.ts:10:1
        10 export = supertest;
           ~~~~~~~~~~~~~~~~~~~
        This module is declared with using 'export =', and can only be used with a default import when using the 'esModuleInterop' flag.
```

We could use `const request = require('supertest')` instead in those files... that seems to fix the problem

DavideViolante:
@gamemaker1 to fix `Unsafe call of an any typed value. ` linting errors replace supertest imports with:
`import { agent as request } from 'supertest'`
Then run `npm run lint` before committing to ensure there are no errors.

gamemaker1:
> `import { agent as request } from 'supertest'`

Thanks, that worked!

test/external/imports/ts-cjs/tsconfig.json
gamemaker1: Fixed - https://github.com/nfriedly/express-rate-limit/pull/282#issuecomment-1006397135 🎉

gamemaker1: Nope not fixed 😕

gamemaker1: We need this so that we can do 

```ts
import createServer from 'express'

const app = createServer()
```

A possible fix is to use `import createServer = require('express')`

gamemaker1: Even `import createServer = require('express')` doesn't work :/

tsconfig.json
```diff
@@ -7,6 +7,7 @@
 		"noUnusedLocals": true,
 		"noImplicitReturns": true,
 		"noFallthroughCasesInSwitch": true,
+		"esModuleInterop": true,
```
DavideViolante, line 10: This can be safely deleted, everything else LGTM. :+1:

gamemaker1: Unfortunately it can't - https://github.com/nfriedly/express-rate-limit/pull/282#discussion_r779435279

DavideViolante (approved):

tsconfig.json
gamemaker1: I think we can let this remain as now we don't need users to use the `esModuleInterop` flag, only the library needs to use it to make sure we can run the tests.

gamemaker1:
@nfriedly if these changes look good, either of us could make a patch release?

test/external/imports/js-cjs/package.json
```diff
@@ -9,7 +9,7 @@
 	},
 	"dependencies": {
 		"express": "^4.17.2",
-		"express-rate-limit": "file:../../../../express-rate-limit-6.0.3.tgz"
+		"express-rate-limit": "file:../../../../express-rate-limit-6.0.4.tgz"
```
nfriedly, line 12: Is this going to work once I bump it to 6.0.5 for this next release?

nfriedly: Looking through the history, it looks like no, although I'm not sure I understand why...

nfriedly (approved):
@gamemaker1 

> if these changes look good, either of us could make a patch release?

Yea, I had one question about the .tgz dependency, but it doesn't look like it blocks the release, so lets ship this as 6.0.5. Would you like to do the honors?

test/external/imports/js-cjs/package.json
gamemaker1: Hmmm this has started to bug me too :/

> Looking through the history, it looks like no, although I'm not sure I understand why...

It runs `npm install ../../../express-rate-limit-*.tgz` every time the test runs on CI/locally, so the version number is bumped - but not committed to git. When we run it locally though, the version number is bumped and if we do a `git add .`, it commits that change too.

I think we can link the package to the folder instead of a .tgz - that should remove this version bump thingie.

gamemaker1:
> Would you like to do the honors?

Yes, hooray!

test/external/imports/js-cjs/package.json
gamemaker1: Releasing 6.0.5 first, then will try out this fix

gamemaker1:
Yay 6.0.5 is live!

test/external/imports/js-cjs/package.json
DavideViolante: Why can't we just specify the version and use the package from npm here? Just to know.

gamemaker1: Because we want to run the tests with the changes made on a branch on the repo before publishing to npm - there's no point in testing after publishing the module 😅

gamemaker1: Opened #283 for this

DavideViolante: I think we can create a GitHub Action that runs the test on that branch (any new branch), and if tests dont pass, publish wont happen.

gamemaker1: That is how it works currently - https://github.com/nfriedly/express-rate-limit/blob/master/.github/workflows/ci.yaml

PR #284, opened 2022-01-07
fix: remove --footer option from esbuild to fix "default is not a function" error

### LINKED ISSUE #280: TypeError: (0 , express_rate_limit_1.default) is not a function
### Description

my version is 6.0.2, but when i use in nest, it occur an error:

(node:5062) UnhandledPromiseRejectionWarning: TypeError: (0 , express_rate_limit_1.default) is not a function

### Library version

6.0.2

### Node version

14.17.6

_No response_

### Module system

ESM

DavideViolante:
Fixes #280

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
 	},
 	"scripts": {
 		"clean": "del-cli dist/ coverage/ *.log *.tmp *.bak *.tgz",
-		"build:cjs": "esbuild --bundle --format=cjs --outfile=dist/index.cjs --footer:js=\"module.exports = rateLimit;\" source/index.ts",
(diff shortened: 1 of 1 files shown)


gamemaker1:
Hi @DavideViolante,

Thanks for taking the time to make this PR!

However, the `--footer` option is required to ensure the library is default-exported for commonjs modules via the 'commonjs style', i.e, `module.exports = rateLimit` since ESBuild, while compiling the CJS, exports it as `module.exports.default = rateLimit`.

gamemaker1:
The import tests fail with this change - https://github.com/nfriedly/express-rate-limit/runs/4738367789?check_suite_focus=true

DavideViolante:
Thanks, didn't see that that test was failing. I will try to open another PR if I find out how to fix it properly.

PR #288, opened 2022-01-18
fix: deprecate `onLimitReached`, reword documentation

### LINKED ISSUE #287: `onLimitReached` is supposed to be deprecated
### Description

I was taking a look at the [release notes for v6.0.0](https://github.com/nfriedly/express-rate-limit/releases/v6.0.0) so that I could copy the changes to the `Store` interface while sending messages to maintainers of external stores and realised that it mentions [the `onLimitReached` function being deprecated](https://github.com/nfriedly/express-rate-limit/blob/master/changelog.md?plain=1#L103-L104). However, the [option is still respected and the callback is still called](https://github.com/nfriedly/express-rate-limit/blob/master/source/lib.ts#L297-L301) when the rate limit is reached for the first time.

I remember having a conversation around `onLimitReached` in #263, but I cannot seem to find it. I think I might have left it in by mistake.

### Library version

6.0.0 - 6.1.0

### Node version

Any version

Any version

### Module system

Any module system

gamemaker1:
## Related Issues

- Fixes #287

### Changed

- Deprecates `onLimitReached`, as v6.0.0 was supposed to do.
- Bumps development dependencies (minor/patch changes only).
- Fixes inconsistencies between the readme and `source/types.ts`.

## Checklist

      added for the same.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -433,7 +433,9 @@ Here is a list of external stores:
 | [`rate-limit-mongo`](https://www.npmjs.com/package/rate-limit-mongo)                   | A [MongoDB](https://www.mongodb.com/)-backed store.                                                   | Legacy              |
 | [`precise-memory-rate-limit`](https://www.npmjs.com/package/precise-memory-rate-limit) | A memory store similar to the built-in one, except that it stores a distinct timestamp for each key.  | Legacy              |
 
-Take a look at [this guide]() if you wish to create your own store.
+Take a look at
+[this guide](https://github.com/nfriedly/express-rate-limit/wiki/Creating-Your-Own-Store)
+if you wish to create your own store.
 
 ## Request API
 
--- a/readme.md
+++ b/readme.md
@@ -258,8 +258,10 @@ Defaults to `429` (HTTP 429 Too Many Requests - RFC 6585).
 > `boolean`
 
 Whether to send the legacy rate limit headers for the limit
-(`X-RateLimit-Limit`), current usage (`X-RateLimit-Remaining`) and time to wait
-before retrying (`Retry-After`) on all responses.
+(`X-RateLimit-Limit`), current usage (`X-RateLimit-Remaining`) and reset time
+(if the store provides it) (`X-RateLimit-Reset`) on all responses. If set to
+`true`, the middleware also sends the `Retry-After` header on all blocked
+requests.
 
 Defaults to `true` (for backward compatibility).
 
@@ -272,8 +274,9 @@ Defaults to `true` (for backward compatibility).
 Whether to enable support for headers conforming to the
 [ratelimit standardization draft](https://github.com/ietf-wg-httpapi/ratelimit-headers/blob/main/draft-ietf-httpapi-ratelimit-headers.md)
 adopted by the IETF (`RateLimit-Limit`, `RateLimit-Remaining`, and, if the store
-supports it, `RateLimit-Reset`). May be used in conjunction with, or instead of
-the `legacyHeaders` option.
+supports it, `RateLimit-Reset`). If set to `true`, the middleware also sends the
+`Retry-After` header on all blocked requests. May be used in conjunction with,
+or instead of the `legacyHeaders` option.
 
 Defaults to `false` (for backward compatibility, but its use is recommended).
 
--- a/readme.md
+++ b/readme.md
@@ -191,52 +191,46 @@ For more information about the `trust proxy` setting, take a look at the
 
 ## Configuration
 
-### `windowMs`
+#### `windowMs`
+
+> `number`
 
 Time frame for which requests are checked/remembered. Also used in the
 `Retry-After` header when the limit is reached.
 
-Note: with non-default stores, you may need to configure this value twice, once
-here and once on the store. In some cases the units also differ (e.g. seconds vs
-miliseconds)
+Note: with stores that do not implement the `init` function (all stores except
+the default, built-in `MemoryStore`), you may need to configure this value
+twice, once here and once on the store. In some cases the units also differ
+(e.g. seconds vs miliseconds).
 
 Defaults to `60000` ms (= 1 minute).
 
-### `max`
+#### `max`
 
-Max number of connections during `windowMs` milliseconds before sending a 429
-response.
+> `number | function`
 
-May be a number, or a function that returns a number or a promise. If `max` is a
-function, it will be called with `request` and `response` params.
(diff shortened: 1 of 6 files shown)


readme.md
```diff
@@ -246,123 +240,157 @@ Defaults to `'Too many requests, please try again later.'`
...
-### `requestWasSuccessful`
+### `onLimitReached`
 
-Function that is called when `skipFailedRequests` and/or
-`skipSuccessfulRequests` are set to `true`. May be overridden if, for example, a
-service sends out a 200 status code on errors.
+> `function`
 
-Defaults to
+Express request handler that sends back a response when a client has reached
```
nfriedly, line 346: I don't like calling it a request handler, because that really wasn't the intention (and tends to lead to people confusing it with the `handler` option). 

Also, I think it happens on the first limited request, not the one before the it.

But, meh, it's deprecated because it was so confusing....

readme.md
```diff
@@ -246,123 +240,157 @@ Defaults to `'Too many requests, please try again later.'`
...
 
-When set to `true`, failed requests won't be counted. Request considered failed
-when:
+> `function`
 
-- response status >= 400
-- requests that were cancelled before last chunk of data was sent (response
-  `close` event triggered)
-- response `error` event was triggered by response
+Method (in the form of middleware) to determine whether or not this request
```
nfriedly, line 357: Again, I don't like calling it middleware, because it's supposed to return a value.

Also, we should mention that it may be async / return a promise

readme.md
```diff
@@ -390,7 +418,7 @@ import rateLimit, {
 } from 'express-rate-limit'
 
 /**
- * A {@link Store} that stores the hit count for each client.
+ * A `Store` that stores the hit count for each client.
```
nfriedly, line 421: I wonder if we should move this example Store to a wiki or a separate file or something. I feel like it takes up a lot of space and it's not very relevant to the vast majority of the people who use this library. What do you think?

nfriedly (commented):
I left some nit-picky comments, but mostly this is really fantastic, and I appreciate the effort. 

One thought for the future: it might be better to do the re-ordering in a separate PR from the rewording, because it gets harder to tell what's changed when the both get mixed in together.

source/types.ts
```diff
@@ -210,48 +220,64 @@ export interface Options {
...
+	 *
+	 * By default, sends back the `statusCode` and `message` set via the options.
 	 */
 	readonly handler: RateLimitExceededEventHandler
 
 	/**
 	 * Express request handler that sends back a response when a client has
 	 * reached their rate limit, and will be rate limited on their next request.
+	 *
+	 * @deprecated 6.x - Please use a custom `handler` that checks the number of
```
nfriedly, line 255: 👍

gamemaker1:
> One thought for the future: it might be better to do the re-ordering in a separate PR from the rewording, because it gets harder to tell what's changed when the both get mixed in together.

Sorry about that - I was already working on these rewording changes for a couple of days when I noticed `onLimitReached` was supposed to be deprecated. Will separate it out next time :)

readme.md
```diff
@@ -193,50 +193,44 @@ For more information about the `trust proxy` setting, take a look at the
...
+Defaults to `5`. Set it to `0` to disable the rate limiter.
 
-```ts
-import rateLimit from 'express-rate-limit'
+An example of using a function:
 
-const isPremium = (request) => {
-	// ...
+```ts
+const max = async (request, response) => {
```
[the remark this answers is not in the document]
gamemaker1: All the other examples (for `skip`, `requestWasSuccessful`, etc.) only show the function, so I decided to make `max` a bit more consistent with the others.

We could change all of them to keep the `rateLimit` call though.

readme.md
```diff
@@ -246,123 +240,157 @@ Defaults to `'Too many requests, please try again later.'`
...
 
 ### `legacyHeaders`
 
-Enable headers for request limit (`X-RateLimit-Limit`) and current usage
-(`X-RateLimit-Remaining`) on all responses and time to wait before retrying
-(`Retry-After`) when `max` is exceeded.
+> `boolean`
 
-Defaults to `true`.
+Whether to send `X-RateLimit-*` headers with the rate limit and the number of
```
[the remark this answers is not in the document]
gamemaker1: Ok, will do

readme.md
gamemaker1: Agree, will change this

readme.md
gamemaker1: I agree, it does feel a little out of place in the readme. We could move this to a `Guides` section of the wiki, in a page `Implementing your own store`.

readme.md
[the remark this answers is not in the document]
nfriedly: Yeah, that's a good point. I can see the argument for consistency. I think that perhaps my thinking was to have one fleshed out example and then let folks interpolate from there to use the other options. Probably not a big deal either way, though.

readme.md
nfriedly: Yeah, I like that. And we can link to it from here.

gamemaker1: Ok, I'll do that then 👍🏾

readme.md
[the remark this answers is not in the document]
gamemaker1: I'll change them to keep the `rateLimit` call as well so users know where the option should go.

readme.md
gamemaker1: Hmm..  I don't see the `Wiki` tab on Github - could check if wikis are enabled for this repository?

nfriedly: Oh, whoops. It's enabled now, and you should have edit permissions.

gamemaker1: Thanks <3

readme.md
```diff
@@ -246,240 +247,195 @@ Defaults to `'Too many requests, please try again later.'`
...
 
-Enable headers for request limit (`X-RateLimit-Limit`) and current usage
-(`X-RateLimit-Remaining`) on all responses and time to wait before retrying
-(`Retry-After`) when `max` is exceeded.
+> `boolean`
 
-Defaults to `true`.
+Whether to send the legacy rate limit headers for the limit
+(`X-RateLimit-Limit`), current usage (`X-RateLimit-Remaining`) and time to wait
+before retrying (`Retry-After`) on all responses.
```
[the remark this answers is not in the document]
gamemaker1: Done

gamemaker1:
Looks good to go?

nfriedly (approved):
Yep, I think so. :shipit:

PR #292, opened 2022-01-21
feat: add named export for memory store

### LINKED ISSUE #289: Expose `MemoryStore` as a named export
### Description

I want to be able to toggle between using a custom third-party store and the "default" memory one. 
Based on https://github.com/nfriedly/express-rate-limit#store you can write your own. I don't want to do that. I want to use the in-memory "default" one from this package. 
E.g. 
```js
export default rateLimit({
   ...
  store: REDIS_URL
    ? new RedisStore(REDIS_STUFF)
    : new Memory(),
})

```
You can't do
```js
import rateLimit, {MemoryStore} from 'express-rate-limit'
```
because the only thing exported is the `rateLimit` default function. 

My only way to do this now is to copy-n-paste the source code from https://github.com/nfriedly/express-rate-limit/blob/master/source/memory-store.ts into my project. 

### Why

To avoid the copy-n-paste and then have to rewrite it if the API changes. 

### Alternatives

Mentioned above .

gamemaker1:
## Related Issues

- Closes #289

### Added

- Added a named export for the `MemoryStore` class.

## Checklist

      added for the same.

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
 	},
 	"scripts": {
 		"clean": "del-cli dist/ coverage/ *.log *.tmp *.bak *.tgz",
-		"build:cjs": "esbuild --bundle --format=cjs --outfile=dist/index.cjs --footer:js=\"module.exports = rateLimit; module.exports.default = rateLimit; module.exports.rateLimit = rateLimit;\" source/index.ts",
+		"build:cjs": "esbuild --bundle --format=cjs --outfile=dist/index.cjs --footer:js=\"module.exports = rateLimit; module.exports.default = rateLimit; module.exports.rateLimit = rateLimit; module.exports.MemoryStore = MemoryStore;\" source/index.ts",
 		"build:esm": "esbuild --bundle --format=esm --outfile=dist/index.mjs source/index.ts",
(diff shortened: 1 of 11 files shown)


nfriedly (approved):
LGTM!

That esbuild `--footer` is getting a little hairy, but I don't think there's a better option right now.

gamemaker1:
> That esbuild --footer is getting a little hairy, but I don't think there's a better option right now.

That's bugging me too 😬. Maybe we can use  esbuild's `__toCommonJSExport` function there instead of the `module.export.{default, rateLimit,Memory store}`?

nfriedly:
Yeah...  ideally I'd like to see a new flag in ES build that controls the default export behavior for cjs, rather than us having to continue hacking around it.

gamemaker1:
> Maybe we can use ESBuild's `__toCommonJSExport` function there instead of the `module.export.{default, rateLimit,Memory store}`

That didn't work - Jest could not resolve named imports in the JS-ESM import tests when I did that. Merging with the big footer for now.

gamemaker1:
Should I release v6.2.0 - it will include this as well as the deprecate `onLimitReached` fixes.

nfriedly:
Yeah, go ahead :)

Issue #132, opened 2019-05-02
Options default set in passed argument object


Alirun:
I considered following code to be working
```
const RateLimit = require('express-rate-limit')

const options = {
  windowMs: 5 * 1000, // 5 seconds
  max: 1
}

const ordersLimit = RateLimit(options)
const usersLimit = RateLimit(options)
```

But requests to `ordersLimit` middleware were blocked by limit of `usersLimit` middleware and vice versa.

I did expect that options object is not directly used and modified within constructor, but it is.

So option.store was set in first constructor call and used as provided in second call.

If it's expected behaviour, please at least add a note in README. If not, I could make a PR to fix this by creating new object in constructor.

nfriedly:
Aah, that's tricky. I never intended for that behavior, but I have a bit of a concern that changing it now could break some existing code. 

I'll go ahead and change it from `defaults(options, {/*default opts*/})` to `Object.assign({}, {/*default opts*/}, options)` and drop the `defaults` dependency. And then I'll mark it as a major version bump when publishing it.

PR #294, opened 2022-02-07
fix: use default value for option when `undefined` is passed

### LINKED ISSUE #290: Passing option with `undefined` does not use default value for option
### Description

If I set `store` to `undefined`, I expect the library to use the default `MemoryStore`. Instead, it treats `undefined` as the store.

Related to #289.

### Library version

6.1.0

### Node version

Any version

Any version

### Module system

Any module system

zkldi:
## Related Issues

Closes #290

## What Does This PR Do?

This PR changes how options are merged with the defaults. At the moment, passing
```js
{ windowMs: undefined }
```

as an option, will result in `windowMs` being set to undefined, rather than falling back to the default of `60000ms`. This is a bit counterintuitive, as most APIs in JS do not differentiate between a property not existing and a property being undefined. This is a consequence of the spread operator being used, though.

### Added

Added a new private function that omits the options, and is called inside `parseOptions`.

## Caveats/Problems/Issues

TypeScript **really** sucks at dealing with object mutation like this. I have had to add a `@ts-expect-error` and `eslint-disable-next-line` inside the function that mutates the object. I spent about 15 minutes trying to come up with a neater way to mutate the object that didn't require this, but I couldn't figure it out.

I ran `autofix`, so that should make this code fit in with the repo a bit better.q

## Checklist

- [X] The issues that this PR fixes/closes have been mentioned above.
- [X] What this PR adds/changes/removes has been explained.
- [X] All tests (`npm test`) pass.
- [X] All added/modified code has been commented, and
- [X] If a new feature has been added or a bug has been fixed, tests have been
      added for the same.

### DIFF
--- a/source/lib.ts
+++ b/source/lib.ts
@@ -100,72 +100,65 @@ const parseOptions = (passedOptions: Partial<Options>): Configuration => {
 
 	// See ./types.ts#Options for a detailed description of the options and their
 	// defaults.
-	const config: Omit<Configuration, 'store'> & { store: Store | LegacyStore } =
-		{
-			windowMs: 60 * 1000,
-			// It's possible that this field might have a LegacyConfig if overrode by passedOptions.
-			// However, LegacyStores are converted to Stores later.
-			store: new MemoryStore(),
-			max: 5,
-			message: 'Too many requests, please try again later.',
-			statusCode: 429,
-			legacyHeaders: passedOptions.headers ?? true,
-			standardHeaders: passedOptions.draft_polli_ratelimit_headers ?? false,
-			requestPropertyName: 'rateLimit',
-			skipFailedRequests: false,
-			skipSuccessfulRequests: false,
-			requestWasSuccessful: (_request: Request, response: Response): boolean =>
-				response.statusCode < 400,
-			skip: (_request: Request, _response: Response): boolean => false,
-			keyGenerator: (request: Request, _response: Response): string => {
-				if (!request.ip) {
-					console.error(
-						'WARN | `express-rate-limit` | `request.ip` is undefined. You can avoid this by providing a custom `keyGenerator` function, but it may be indicative of a larger issue.',
-					)
-				}
+	const config: Configuration = {
+		windowMs: 60 * 1000,
+		max: 5,
+		message: 'Too many requests, please try again later.',
+		statusCode: 429,
+		legacyHeaders: passedOptions.headers ?? true,
+		standardHeaders: passedOptions.draft_polli_ratelimit_headers ?? false,
+		requestPropertyName: 'rateLimit',
+		skipFailedRequests: false,
+		skipSuccessfulRequests: false,
+		requestWasSuccessful: (_request: Request, response: Response): boolean =>
+			response.statusCode < 400,
+		skip: (_request: Request, _response: Response): boolean => false,
+		keyGenerator: (request: Request, _response: Response): string => {
+			if (!request.ip) {
+				console.error(
+					'WARN | `express-rate-limit` | `request.ip` is undefined. You can avoid this by providing a custom `keyGenerator` function, but it may be indicative of a larger issue.',
+				)
+			}
 
-				return request.ip
-			},
-			handler: (
-				_request: Request,
-				response: Response,
-				_next: NextFunction,
-				_optionsUsed: Options,
-			): void => {
-				response.status(config.statusCode).send(config.message)
-			},
-			onLimitReached: (
-				_request: Request,
-				_response: Response,
-				_optionsUsed: Options,
-			): void => {},
-			// Allow the options object to be overriden by the options passed to the middleware.
-			...notUndefinedOptions,
-		}
+			return request.ip
+		},
+		handler: (
+			_request: Request,
+			response: Response,
+			_next: NextFunction,
+			_optionsUsed: Options,
+		): void => {
+			response.status(config.statusCode).send(config.message)
+		},
+		onLimitReached: (
+			_request: Request,
+			_response: Response,
+			_optionsUsed: Options,
+		): void => {},
+		// Allow the options object to be overriden by the options passed to the middleware.
+		...notUndefinedOptions,
+		// Note that this field is declared after the user's options are spread in,
+		// so that this field doesn't get overriden with an un-promisified store!
+		store: promisifyStore(notUndefinedOptions.store ?? new MemoryStore()),
+	}
 
 	// Ensure that the store passed implements the either the `Store` or `LegacyStore`
 	// interface
 	if (
-		(typeof (config.store as LegacyStore).incr !== 'function' &&
-			typeof (config.store as Store).increment !== 'function') ||
+		typeof config.store.increment !== 'function' ||
 		typeof config.store.decrement !== 'function' ||
 		typeof config.store.resetKey !== 'function' ||
 		(typeof config.store.resetAll !== 'undefined' &&
 			typeof config.store.resetAll !== 'function') ||
-		(typeof (config.store as Store).init !== 'undefined' &&
-			typeof (config.store as Store).init !== 'function')
+		(typeof config.store.init !== 'undefined' &&
+			typeof config.store.init !== 'function')
 	) {
 		throw new TypeError(
 			'An invalid store was passed. Please ensure that the store is a class that implements the `Store` interface.',
 		)
 	}
 
-	// Promisify the store, if it is not already
-	// This also turns a LegacyStore into a Store, if needs be.
-	config.store = promisifyStore(config.store)
-
-	// Return the 'clean' options
-	return config as Configuration
+	return config
 }
 
 /**
--- a/source/lib.ts
+++ b/source/lib.ts
@@ -142,8 +142,7 @@ const parseOptions = (passedOptions: Partial<Options>): Configuration => {
 		store: promisifyStore(notUndefinedOptions.store ?? new MemoryStore()),
 	}
 
-	// Ensure that the store passed implements the either the `Store` or `LegacyStore`
-	// interface
+	// Ensure that the store passed implements the `Store` interface
 	if (
 		typeof config.store.increment !== 'function' ||
 		typeof config.store.decrement !== 'function' ||
--- a/source/lib.ts
+++ b/source/lib.ts
@@ -11,6 +11,9 @@ import {
 	LegacyStore,
 	Store,
 	IncrementResponse,
+	ValueDeterminingMiddleware,
+	RateLimitExceededEventHandler,
+	RateLimitReachedEventHandler,
 } from './types.js'
 
 /**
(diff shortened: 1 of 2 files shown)


gamemaker1:
Hi @zkldi,

Thanks for the PR! It looks good at first glance, but it's really late right now - I'll review it in detail first thing tomorrow :)

zkldi:
Thank you! appreciate your time.

source/lib.ts
```diff
@@ -76,18 +76,23 @@ const promisifyStore = (passedStore: LegacyStore | Store): Store => {
 	return new PromisifiedStore()
 }
 
+type OptionsWithLegacyStore = Omit<Partial<Options>, 'store'> & {
+	store?: Store | LegacyStore
+}
```
gamemaker1, line 81: What if we made two different interfaces:

1. `Options`: this is represents the options the user can pass to the rate limiter. The `store` field in the `Options` interface should then be changed to be of type `Store | LegacyStore`. All fields should remain `readonly`, as they currently are.

2. `Configuration`: this represents the parsed configuration that we derive from the options passed. This can exclude deprecated fields like `headers` and `draft_polli_ratelimit_headers`, the `store` field should be of type `Store` and all fields can be read-write.

source/lib.ts
```diff
@@ -76,18 +76,23 @@ const promisifyStore = (passedStore: LegacyStore | Store): Store => {
...
-const parseOptions = (
-	passedOptions: Omit<Partial<Options>, 'store'> & {
-		store?: Store | LegacyStore
-	},
-): Options => {
+const parseOptions = (passedOptions: OptionsWithLegacyStore): Options => {
+	// Passing undefined should be equivalent to not passing an option at all, so we'll
+	// omit all fields where their value is undefined.
+	const omittedOptions: OptionsWithLegacyStore =
+		omitUndefinedOptions(passedOptions)
```
gamemaker1, line 94: Maybe we could put the `omitUndefinedOptions` code here itself instead of declaring a new function? It is just a loop after all.

And the `ommittedOptions` variable could be of type `Partial<Options>` once we implement the change above.

gamemaker1: Also, could we rename `ommittedOptions` to something like `notUndefinedOptions`?

source/lib.ts
```diff
@@ -76,18 +76,23 @@ const promisifyStore = (passedStore: LegacyStore | Store): Store => {
...
  * @param options {Options} - The options the user specifies.
  *
  * @returns {Options} - A complete configuration object.
  */
-const parseOptions = (
-	passedOptions: Omit<Partial<Options>, 'store'> & {
-		store?: Store | LegacyStore
-	},
-): Options => {
+const parseOptions = (passedOptions: OptionsWithLegacyStore): Options => {
```
gamemaker1, line 90: If we add the `Configuration` interface, this function should return it:

```suggestion
const parseOptions = (passedOptions: OptionsWithLegacyStore): Configuration => {
```

source/lib.ts
```diff
@@ -76,18 +76,23 @@ const promisifyStore = (passedStore: LegacyStore | Store): Store => {
...
+type OptionsWithLegacyStore = Omit<Partial<Options>, 'store'> & {
+	store?: Store | LegacyStore
+}
+
 /**
  * Type-checks and adds the defaults for options the user has not specified.
  *
  * @param options {Options} - The options the user specifies.
  *
  * @returns {Options} - A complete configuration object.
```
gamemaker1, line 88:
```suggestion
 * @returns {Configuration} - A complete configuration object.
```

source/lib.ts
```diff
@@ -76,18 +76,23 @@ const promisifyStore = (passedStore: LegacyStore | Store): Store => {
...
-): Options => {
+const parseOptions = (passedOptions: OptionsWithLegacyStore): Options => {
+	// Passing undefined should be equivalent to not passing an option at all, so we'll
+	// omit all fields where their value is undefined.
+	const omittedOptions: OptionsWithLegacyStore =
+		omitUndefinedOptions(passedOptions)
+
 	// See ./types.ts#Options for a detailed description of the options and their
 	// defaults.
 	const options = {
```
gamemaker1, line 98: With the new `Configuration` interface, this could be made:

```suggestion
	const config: Configuration = {
```

source/lib.ts
```diff
@@ -126,8 +131,8 @@ const parseOptions = (
 			_response: Response,
 			_optionsUsed: Options,
 		): void => {},
-		// Allow the above to be overriden by the options passed to the middleware
-		...passedOptions,
+		// Allow the options object to be overrode by the options passed to the middleware.
```
gamemaker1, line 134:
```suggestion
		// Allow the default values to be overriden by the options passed to the middleware.
```

I think overriden (the past participle) fits here

gamemaker1 (commented):
Suggested a couple of changes - most notably the introduction of a `Configuration` type that will enable us to get rid of the `@ts-expect-error` comments while parsing the options.

source/lib.ts
zkldi: Renamed the variable -- however, I moved the omitting function to a separate function because it's not very clean. It's only used once so it's fine to inline, but it's not great :s

source/lib.ts
zkldi: Added the Configuration type. I'm not familiar with what fields are deprecated in this library so I just omitted the ones you mentioned.

gamemaker1: If we make `Configuration` a separate interface, whose fields do not have the `readonly` modifier, we could get rid of the `@ts-expect-error` directive.

source/lib.ts
```diff
@@ -76,87 +76,96 @@ const promisifyStore = (passedStore: LegacyStore | Store): Store => {
...
-				console.error(
-					'WARN | `express-rate-limit` | `request.ip` is undefined. You can avoid this by providing a custom `keyGenerator` function, but it may be indicative of a larger issue.',
-				)
-			}
+	const config: Omit<Configuration, 'store'> & { store: Store | LegacyStore } =
+		{
+			windowMs: 60 * 1000,
+			// It's possible that this field might have a LegacyConfig if overrode by passedOptions.
+			// However, LegacyStores are converted to Stores later.
+			store: new MemoryStore(),
```
gamemaker1, line 108: I don't like the `Omit` stuff we are doing [here](https://github.com/nfriedly/express-rate-limit/pull/294/files/6d2632fbc6f5f5d236e2725b4002be38d4649ce0..7eb1e6b39a2ff1dfdb92d65729bff9168f2b41aa#diff-5c2516e9822679491492c832b9ace79b6945bdd891586e07b0232ebef1c7e0b1R103) - we could remove the need for that by changing this line to:

```suggestion
			store: promisifyStore(notUndefinedOptions.store ?? new MemoryStore()),
```

And removing the `promisifyStore` call [here](https://github.com/nfriedly/express-rate-limit/pull/294/files/6d2632fbc6f5f5d236e2725b4002be38d4649ce0..7eb1e6b39a2ff1dfdb92d65729bff9168f2b41aa#diff-5c2516e9822679491492c832b9ace79b6945bdd891586e07b0232ebef1c7e0b1R165).

source/lib.ts
```diff
@@ -100,72 +100,65 @@ const parseOptions = (passedOptions: Partial<Options>): Configuration => {
...
+		},
+		onLimitReached: (
+			_request: Request,
+			_response: Response,
+			_optionsUsed: Options,
+		): void => {},
+		// Allow the options object to be overriden by the options passed to the middleware.
+		...notUndefinedOptions,
+		// Note that this field is declared after the user's options are spread in,
+		// so that this field doesn't get overriden with an un-promisified store!
```
gamemaker1, line 141: Good catch 🚀

source/lib.ts
```diff
@@ -119,39 +128,37 @@ const parseOptions = (
...
-		...passedOptions,
+		// Allow the options object to be overriden by the options passed to the middleware.
+		...notUndefinedOptions,
+		// Note that this field is declared after the user's options are spread in,
+		// so that this field doesn't get overriden with an un-promisified store!
+		store: promisifyStore(notUndefinedOptions.store ?? new MemoryStore()),
 	}
 
 	// Ensure that the store passed implements the either the `Store` or `LegacyStore`
 	// interface
```
nfriedly, line 146: This comment is now outdated

```suggestion
	// Ensure that the store passed implements the either the `Store` interface
```

(unless we want to move this check to earlier, so that it can validate a legacy store before it's promisified)

zkldi: Resolved

nfriedly (approved):
Thanks @zkldi - I really like this overall.

I se one minor point - Because the validation happens after the `promisifyStore` step here, this results in almost no validation of legacy stores... but given that any new store would likely use the new API, that's probably OK.

@gamemaker1: feel free to merge and publish once you're happy with this PR.

source/lib.ts
gamemaker1: @zkldi could you implement this (create a separate interface for `Configuration`) change?

gamemaker1:
Apart from [this](https://github.com/nfriedly/express-rate-limit/pull/294#discussion_r803259250), I think this is really great - thanks a lot @zkldi :)

Maybe in a future PR we could add some manual type checking for all the options (I'm thinking looping through the passed options and if an option is set, compare the type of the passed option with the type of the default value).

gamemaker1 (changes requested):
Apart from [this](https://github.com/nfriedly/express-rate-limit/pull/294#discussion_r803259250), I think this is really great - thanks a lot @zkldi :)

gamemaker1:
Also maybe in a future PR we could add some manual type checking for all the options (I'm thinking looping through the passed options and if an option is set, compare the type of the passed option with the type of the default value).

source/lib.ts
zkldi: Sure, I take it i can just copy over the implementation from Options?

zkldi: No worries, I should have a commit up in the hour.

gamemaker1: Great, thanks a lot :D

zkldi: It seems a bit redundant to copy the implementation twice. Is there any actual reason Options is marked as readonly?

zkldi: Yuck. Changing this just gives us a different TS error -- ` Unsafe assignment of an any value.  @typescript-eslint/no-unsafe-assignment`. As mentioned before TS is pretty poor when it comes to performing object manipulation.

Changing to a Configuration type doesn't appear to let us remove the ts-expect-error directive.

gamemaker1: The `Options` interface's properties are marked readonly to ensure we don't change them anyhow - see issue #132

gamemaker1: The error you describe is a linter error, and you can suppress that separately. I usually try to suppress only lint errors if I have to, not compiler errors.

zkldi: Oh yeah whoops - my bad. I'll submit the commit now then.

gamemaker1 (approved):
LGTM! Merging, and will release a new version soon.

Thanks for the great work @zkldi

zkldi:
Thank you!

gamemaker1:
Released as part of [`v6.2.1`](https://github.com/nfriedly/express-rate-limit/releases/tag/v6.2.1) 🎉

PR #296, opened 2022-02-24
docs: custom middleware with standardHeaders

reslear:
## Checklist

- [ ] What this PR adds/changes/removes has been explained.
- [ ] All tests (`npm test`) pass.
      added for the same.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -136,6 +136,23 @@ app.post('/create-account', createAccountLimiter, (request, response) => {
 })
 ```
 
+
+To use a custom middleware with standardHeaders:
+
+```ts
+const apiLimiter: typeof rateLimit = (options = {}) => {
+  return rateLimit({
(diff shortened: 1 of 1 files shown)


gamemaker1:
Hi @reslear,

Thanks for the PR! However, all of the examples in the readme already specifically set the `standardHeaders` option to true and the `legacyHeaders` option to false - I don't think we need to add another example just for that.

nfriedly:
Hey @reslear 

Thank you for trying to help make our documentation better!  That said, I agree with @gamemaker1 that this example probably isn't a good fit here. I feel like it's probably more likely to confuse a new user, whereas a more seasoned user ought to be able to come up with something similar based on the existing examples. 

So, I'm going to go ahead and close this without merging it in.

Sorry for not using your work this time, but please don't let it deter you from contributing in the future - we do really value folks like yourself who take the time to help out!

PR #304, opened 2022-04-23
meta: add express 5 as peer dependency

InSantoshMahto:
## What Does This PR Do?
This PR will resolve the peer dependency error fix for express v5

## Checklist

- [ ] What this PR adds/changes/removes has been explained.
- [ ] All tests (`npm test`) pass.
      added for the same.

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,7 +67,7 @@
 		"prepare": "run-s compile && husky install config/husky"
 	},
 	"peerDependencies": {
-		"express": "^4"
+		"express": "^4 || ^5"
 	},
 	"devDependencies": {
 		"@jest/globals": "27.4.6",
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -52,19 +52,6 @@ jobs:
         os: [ubuntu-latest]
     runs-on: ${{ matrix.os }}
     continue-on-error: true
-    services:
-      rate-limit-redis:
-        image: redis
-        ports:
-          - 6379:6379
-      rate-limit-mongo:
-        image: mongo
-        ports:
-          - 27017:27017
-      rate-limit-memcached:
-        image: memcached
-        ports:
-          - 11211:11211
     steps:
       - name: Checkout the repository
         uses: actions/checkout@v2
--- a/package-lock.json
+++ b/package-lock.json
@@ -1829,9 +1829,9 @@
 			"dev": true
 		},
 		"node_modules/array-flatten": {
-			"version": "1.1.1",
-			"resolved": "https://registry.npmjs.org/array-flatten/-/array-flatten-1.1.1.tgz",
-			"integrity": "sha1-ml9pkFGx5wczKPKgCJaLZOopVdI=",
+			"version": "3.0.0",
+			"resolved": "https://registry.npmjs.org/array-flatten/-/array-flatten-3.0.0.tgz",
+			"integrity": "sha512-zPMVc3ZYlGLNk4mpK1NzP2wg0ml9t7fUgDsayR5Y5rSzxQilzR9FGu/EH2jQOcKSAeAfWeylyW8juy3OkWRvNA==",
 			"dev": true
 		},
 		"node_modules/array-includes": {
@@ -2025,7 +2025,7 @@
 				"type-is": "~1.6.18"
 			},
 			"engines": {
-				"node": ">= 0.8"
+				"node": ">= 0.10"
 			}
 		},
 		"node_modules/brace-expansion": {
@@ -4591,7 +4591,7 @@
 				"content-type": "~1.0.4",
 				"cookie": "0.4.2",
 				"cookie-signature": "1.0.6",
-				"debug": "2.6.9",
+				"debug": "3.1.0",
 				"depd": "~1.1.2",
 				"encodeurl": "~1.0.2",
 				"escape-html": "~1.0.3",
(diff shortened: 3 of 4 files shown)


gamemaker1:
Hi @InSantoshMahto,

Thanks for taking the time to make the PR :)

However, Express v5 is still in beta - I think it would make more sense to update the peer dependency once v5.0.0 is officially released and tested with this library.

gamemaker1:
Also, why does this change involve 57 files? Did you need to change something to make it work with Express v5?

InSantoshMahto:
> Also, why does this change involve 57 files? Did you need to change something to make it work with Express v5?

I just rename the source directory to src to look the same as other projects. and remove some comments

package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "express-rate-limit",
-	"version": "6.3.0",
+	"version": "6.3.1",
```
gamemaker1, line 3: Please don't bump the version, we will do that as part of the release process later on

InSantoshMahto:
> Hi @InSantoshMahto,
> 
> Thanks for taking the time to make the PR :)
> 
> However, Express v5 is still in beta - I think it would make more sense to update the peer dependency once v5.0.0 is officially released and tested with this library.

Can we make it available express-rate-limit@next so that we can use with express@next for V5

gamemaker1:
> Can we make it available express-rate-limit@next so that we can use with express@next for V5

I think we could do that. AFAIK usage with Express v5 should be possible even if the peer dependency is not updated though - npm only warns if peer dependencies are not met. Have you tested the library and checked if it works with the newer version without any changes?

gamemaker1:
> I just rename the source directory to src to look the same as other projects. and remove some comments

Could you please undo those changes? The comments are helpful for new contributors and I prefer keeping full form directory names in projects. Also, the tests are failing because of the incomplete rename.

InSantoshMahto:
> > I just rename the source directory to src to look the same as other projects. and remove some comments
> 
> Could you please undo those changes? The comments are helpful for new contributors and I prefer keeping full form directory names in projects. Also, the tests are failing because of the incomplete rename.

sure

InSantoshMahto:
> > Can we make it available express-rate-limit@next so that we can use with express@next for V5
> 
> I think we could do that. AFAIK usage with Express v5 should be possible even if the peer dependency is not updated though - npm only warns if peer dependencies are not met. Have you tested the library and checked if it works with the newer version without any changes?

Yes. work perfectly with waring while using npm. But while using pnpm I faced the issue. it break my docker build.

package.json
InSantoshMahto: Sure

gamemaker1:
> But while using pnpm I faced the issue. it break my docker build.

What error do you receive?

InSantoshMahto:
> > But while using pnpm I faced the issue. it break my docker build.
> 
> What error do you receive?
> 
> In a test project, running `pnpm install express@5.0.0-beta.1` gives me the following output:
> 
> ```
> Packages: +13 -12
> +++++++++++++------------
> 
> Progress: resolved 295, reused 268, downloaded 8, added 13, done
> 
> dependencies:
> - express 4.17.3
> + express 5.0.0-beta.1
> - express-rate-limit 6.3.0
> + express-rate-limit 6.3.0
> 
>  WARN  Issues with peer dependencies found
> .
> └─┬ express-rate-limit
>   └── ✕ unmet peer express@^4: found 5.0.0-beta.1
> ```

yes same error I also received

gamemaker1:
This is just a warning, and shouldn't result in a failure during installation.

Anyways, I don't see an issue in publishing a new version under the `next` tag. @nfriedly what do you think?

nfriedly:
I think express 5.x compatibility is fine to declare in a regular semver minor release.

I would like to see us run our automated tests against both 4.x and 5.x then - but I think that could be as simple as adding a few lines to [test/external/run-all-tests](https://github.com/nfriedly/express-rate-limit/blob/master/test/external/run-all-tests) - basically, after the two `npm test` lines, do `npm install express@5` and then `npm test` again.

gamemaker1:
@InSantoshMahto could you please check the "allow maintainer edits" box on the right hand side on this page so I can push a commit updating the tests? Thanks!

InSantoshMahto:
> @InSantoshMahto could you please check the "allow maintainer edits" box on the right hand side on this page so I can push a commit updating the tests? Thanks!
Screenshot 2022-04-24 at 1 55 21 PM
Hey @gamemaker1 Already checked

gamemaker1:
Ohk, thank you

gamemaker1 (approved):
Looks good to me!

gamemaker1:
@nfriedly Do take a look at the changes, the tests now run with express 4 and 5

nfriedly (approved):
I like it!

it might be good to add a couple of `echo "running test $foo"; npm ls express` sort of comments to `run-all-tests`, but that's more cosmetic than anything else.

gamemaker1:
Sure, will do

package.json
```diff
@@ -79,7 +79,7 @@
 		"del-cli": "^4.0.1",
 		"dts-bundle-generator": "^6.4.0",
 		"esbuild": "^0.14.12",
-		"express": "^4.17.1",
```
nfriedly, line 82: oh, I think this should probably stay on 4.x if it's going to be the starting point for tests that first run against 4.x, then5.x

gamemaker1: This version is actually not changed during the tests - its only the version of express used by the projects that changes. This express dependency is only for type definitions, and AFAIK type defs for 5.0.0 haven't been released (or maybe they are the same as 4.x?)

gamemaker1: Perhaps this should stay at 4.17 then

nfriedly: yeah, I guess it doesn't matter too much, but let's keep it at 4.x for now.

gamemaker1:
Added some colors 🎆

gamemaker1:
Merging after fixing a merge conflict, will release a minor version by tomorrow :D

gamemaker1:
Released as part of [v6.4.0](https://github.com/nfriedly/express-rate-limit/releases/tag/v6.4.0)

PR #305, opened 2022-04-24
meta: run ci on node 18

gamemaker1:
## Related Issues

--

### Added

- Runs Lint and Publish steps of CI on Node 18.
- Added Node 18 to version matrix for Test steps too.

## Caveats/Problems/Issues

--

## Checklist

      added for the same.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -10,7 +10,7 @@ jobs:
     name: Lint
     strategy:
       matrix:
-        node-version: [18.x]
+        node-version: [18]
         os: [ubuntu-latest]
     runs-on: ${{ matrix.os }}
     continue-on-error: true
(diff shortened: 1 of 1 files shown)


gamemaker1:
@nfriedly should we stop running tests on Node 12? (EOL for Node 12 is [April 30th](https://github.com/nodejs/Release#release-schedule), 6 days from now)

nfriedly:
I like adding 18 to the list, but I'd prefer to keep 12 unless there's some incompatibility with a feature that we (or, more likely, one of our dependencies) needs.

I think we could drop 15 and 17, though. Neither are LTS releases, and I find it unlikely that we'd hit any issue in on of those that isn't also in 14, 16, or 18. I generally like to test against a variety of LTS releases, + whatever the current release is.

gamemaker1:
Sounds good, I'll keep 12 but drop 15 and 17 then

nfriedly (approved):
Approving now based on our conversation; I'm going to be out most of the rest of the day.

gamemaker1:
Thanks, have a good day :D

PR #306, opened 2022-04-24
chore: bump all dependencies

gamemaker1:
## Related Issues

--

### Changed

- Bumped dependencies
- Removed `^` and pinned them so that the npm doesn't automatically bump versions unless we ask it too.

## Caveats/Problems/Issues

~The library test on MacOS (Node 12) is failing because of a 409 error while fetching packages from the npm registry.~ Re-running the CI worked the second time, must be a network issue 🤷🏾 

## Checklist

      added for the same.


.npmrc
```diff
@@ -0,0 +1,6 @@
+# .npmrc
+# Configuration for npm and pnpm
+
+# Uses the exact version instead of any within-patch-range version of an
+# installed package
+save-exact=true
```
nfriedly, line 6: What's your thinking here? (We're already doing `npm ci` for our tests, which installs exact dependencies and sub-dependencies from package-lock.json.)

nfriedly: oh, actually, we're doing it for the top-level, but not the external tests. But still, we could change that if e wanted to.

gamemaker1: `npm ci` is only for the tests, however, when someone new who wants to contribute runs `npm install`, npm will install any version within the patch range instead of exactly that version, and the person will commit the version bump. To prevent this, the `save-exact` removes the `^` from `package.json`.

A couple of tests fail on node 12 on some version of jest and typescript when I accidently bumped deps like this a week ago, hence I put this in to prevent it from happening in the future.

nfriedly: Eh, I use `npm ci` locally, but maybe I'm just weird 🙃 

Does `save-exact` prevent changes to `package-lock.json` when running `npm install`?

Fair enough, though. And, if current versions of jest or whatever else we depend on no longer work in node.js 12, then perhaps we should drop it from CI.

nfriedly (approved):

.npmrc
gamemaker1: > Does save-exact prevent changes to package-lock.json when running npm install?

I think it does, take a look at some of the answers [here](https://stackoverflow.com/questions/45022048/why-does-npm-install-rewrite-package-lock-json)

> Fair enough, though. And, if current versions of jest or whatever else we depend on no longer work in node.js 12, then perhaps we should drop it from CI.

I haven't spend much time on it so I'm not exactly sure what is causing the trouble, will look into it and let you know

PR #311, opened 2022-07-20
feat: allow the `message` option to be a function

colderry:
Allow for manual response to ratelimit by function message.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -136,7 +136,7 @@ app.post('/create-account', createAccountLimiter, (request, response) => {
 })
 ```
 
-To respond by function message:
+To send a response via function message:
 
 ```ts
 import rateLimit from 'express-rate-limit'
@@ -145,6 +145,7 @@ const apiLimiter = rateLimit({
 	windowMs: 60 * 60 * 1000, // 1 hour
 	max: 5, // Limit each IP to 5 create account requests per `window` (here, per hour)
 	message: (request, response) => {
+		// response.send('Too many accounts created from this IP, please try again after an hour')
 		return 'Too many accounts created from this IP, please try again after an hour'
 	},
 	standardHeaders: true, // Return rate limit info in the `RateLimit-*` headers
--- a/readme.md
+++ b/readme.md
@@ -136,27 +136,6 @@ app.post('/create-account', createAccountLimiter, (request, response) => {
 })
 ```
 
-To send a response via function message:
-
-```ts
-import rateLimit from 'express-rate-limit'
-
-const apiLimiter = rateLimit({
-	windowMs: 60 * 60 * 1000, // 1 hour
-	max: 5, // Limit each IP to 5 create account requests per `window` (here, per hour)
-	message: (request, response) => {
-		// response.send('Too many accounts created from this IP, please try again after an hour')
-		return 'Too many accounts created from this IP, please try again after an hour'
-	},
-	standardHeaders: true, // Return rate limit info in the `RateLimit-*` headers
-	legacyHeaders: false, // Disable the `X-RateLimit-*` headers
-})
-
-app.post('/create-account', createAccountLimiter, (request, response) => {
-	//...
-})
-```
-
 To use a custom store:
 
 ```ts
@@ -259,11 +238,30 @@ const limiter = rateLimit({
 
 The response body to send back when a client is rate limited.
 
-May be a `string`, JSON object, function, or any other value that Express's
-[response.send](https://expressjs.com/en/4x/api.html#response.send) method
-supports.
+May be a `string`, JSON object, or any other value that Express's
+[`response.send`](https://expressjs.com/en/4x/api.html#res.send) method
+supports. It can also be a (sync/async) function that accepts the Express
+request and response objects and then returns a `string`, JSON object or any
+other value the Express `response.send` function accepts.
 
 Defaults to `'Too many requests, please try again later.'`
+
+An example of using a function:
+
+```ts
+const isPremium = async (user) => {
+	// ...
+}
+
+const limiter = rateLimit({
+	// ...
+	message: async (request, response) => {
+		if (await isPremium(request.user))
+			return 'You can only make 10 requests every hour.'
+		else return 'You can only make 5 requests every hour.'
+	},
+})
+```
 
 ### `statusCode`
 
--- a/readme.md
+++ b/readme.md
@@ -367,7 +367,8 @@ const limiter = rateLimit({
 Express request handler that sends back a response when a client is
 rate-limited.
 
-By default, sends back the `statusCode` and `message` set via the options:
+By default, sends back the `statusCode` and `message` set via the `options`,
+similar to this:
 
 ```ts
 const limiter = rateLimit({
--- a/readme.md
+++ b/readme.md
@@ -145,11 +145,7 @@ const apiLimiter = rateLimit({
 	windowMs: 60 * 60 * 1000, // 1 hour
 	max: 5, // Limit each IP to 5 create account requests per `window` (here, per hour)
 	message: (request, response) => {
-		response
-			.status(429)
-			.send(
-				'Too many accounts created from this IP, please try again after an hour',
-			)
+		return 'Too many accounts created from this IP, please try again after an hour'
 	},
 	standardHeaders: true, // Return rate limit info in the `RateLimit-*` headers
 	legacyHeaders: false, // Disable the `X-RateLimit-*` headers
--- a/readme.md
+++ b/readme.md
@@ -136,6 +136,30 @@ app.post('/create-account', createAccountLimiter, (request, response) => {
 })
 ```
 
+To respond by function message:
+
+```ts
+import rateLimit from 'express-rate-limit'
+
+const apiLimiter = rateLimit({
+	windowMs: 60 * 60 * 1000, // 1 hour
+	max: 5, // Limit each IP to 5 create account requests per `window` (here, per hour)
+	message: (request, response) => {
+		response
+			.status(429)
+			.send(
+				'Too many accounts created from this IP, please try again after an hour',
+			)
+	},
+	standardHeaders: true, // Return rate limit info in the `RateLimit-*` headers
+	legacyHeaders: false, // Disable the `X-RateLimit-*` headers
+})
+
+app.post('/create-account', createAccountLimiter, (request, response) => {
+	//...
+})
+```
+
 To use a custom store:
 
 ```ts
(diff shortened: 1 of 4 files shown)


gamemaker1:
Hi @colderry,

Thanks for taking the time to make this PR!

Could you describe a use case that would benefit from making message a function? The demonstrated use case that you have added is already possible by setting the `statusCode` and `message` options.

gamemaker1:
Also, I love how you planned and split the changes into three commits :]

colderry:
> Hi @colderry,
> 
> Thanks for taking the time to make this PR!
> 
> Could you please describe a use case that would benefit from making message a function? The use case that you have added to the readme is already possible by setting the `statusCode` and `message` options.

It can be beneficial for someone who wants to check who is making these requests on their app and ban them. Another use case is if someone wants to log these requests, they can do it.

nfriedly:
Hey @colderry, thanks for taking the time to do this, and for the thorough job documenting and testing it!

This seems seems very similar to the [`handler` option](https://github.com/nfriedly/express-rate-limit#handler), so I don't think it would make sense to merge in it's current form.

That said, I like the idea allowing `message` to be a function - I can see value there, and it fits with the way most other options that are allowed to be functions or values.

However, I think that it would make more sense and be more consistent with the rest of the API to allow `message` to return an arbitrary value, and then use that value as the response rather than expect to `message` sent a response itself. 

Also, `await` instead of passing in `next` because we do want to allow for async `message` functions, but not in a way that bypasses the rest of express-rate-limit. (In the rare case a user actually needs to call `next`, they can supply a custom `handler` option.)

So something like this:

```ts
const message: any = (typeof options.message === 'function') ? await options.message(request, response) : message;
```

But, we should probably also guard against crashing if someone does it your way  - so make the handler check if a response was sent before sending message. Something like this:

```ts
if (!response.headersSent) {
    res.send(message);
}
```

(`headersSent` isn't technically a guarantee that a response was sent, but it is a good proxy. There's probably some other property that would be better to check, but I don't remember what off the top of my head.)

Finally, we should probably set the status code before calling `message`, in case `message()` sends a response without setting the status. So the final handler might look something like this:

```ts
res.status(options.status);
const message: any = (typeof options.message === 'function') ? await options.message(request, response) : message;
if (!response.headersSent) {
    res.send(message);
}
```

I haven't tested any of this, and I'm pretty sleep-deprived right now, so I may be missing something dumb. But that should at least give you the gist of what I'm thinking.

This adds a bit of internal complexity over your solution, and might merit a couple more tests, but I think it will make for a better and more consistent API for our users.

nfriedly:
> It can be beneficial for someone who wants to check who is making these requests on their app and ban them. Another use case is if someone wants to log these requests, they can do it.

I think `handler` could manage those cases as easily as `message`, but an EventEmitter might be an even better fit. I remember looking at that once before, but perhaps it's time to look at it again.

colderry:
Hey @nfriedly, first of all, thank you for your time.

I have made the changes according to your message. Please check it.

gamemaker1:
> It can be beneficial for someone who wants to check who is making these requests on their app and ban them. Another use case is if someone wants to log these requests, they can do it.

For preventing someone from making requests at all, you could set `max` to 1 based on the request:

```ts
const bannedIps = ['192.168.0.10', '192.168.0.23']

const limiter = rateLimit({
	// ...
	max: async (request, response) => {
		if (bannedIps.includes(request.ip)) return 1
		else return 5
	},
})
```

To log requests that are rate-limited, you could use the handler function instead:

```ts
const limiter = rateLimit({
	// ...
	handler: async (request, response) => {
		console.log(`http: ${request.ip} was rate limited`)
	},
})
```

gamemaker1:
However I do agree with @nfriedly, `message` being a function could be beneficial to our users.

readme.md
```diff
@@ -238,7 +259,7 @@ const limiter = rateLimit({
 
 The response body to send back when a client is rate limited.
 
-May be a `string`, JSON object, or any other value that Express's
+May be a `string`, JSON object, function, or any other value that Express's
 [response.send](https://expressjs.com/en/4x/api.html#response.send) method
 supports.
```
gamemaker1, line 264:
```suggestion
May be a `string`, JSON object, or any other value that Express's
[`response.send`](https://expressjs.com/en/4x/api.html#res.send) method
supports. It can also be a (sync/async) function that accepts the Express
request and response objects and then returns a `string`, JSON object or
any other value the Express `response.send` function accepts.
```

Also, could you please change the `any` above this line to `any | function`?

readme.md
```diff
@@ -238,7 +259,7 @@ const limiter = rateLimit({
...
```
gamemaker1, line 265: Below the `Defaults to ...`, it would be helpful to add an example of using the function, similar to the `max` option above:

---

An example of using a function:

````ts
const isPremium = async (user) => {
	// ...
}

const limiter = rateLimit({
	// ...
	message: async (request, response) => {
		if (await isPremium(request.user)) return 'You can only make 10 requests every hour.'
		else return 'You can only make 5 requests every hour.'
	},
})
```

source/lib.ts
```diff
@@ -144,13 +144,24 @@ const parseOptions = (passedOptions: Partial<Options>): Configuration => {
...
+			let { message } = config as { message: unknown }
+			// Set the response status code
+			response.status(config.statusCode)
+			// Call the `message` if it is a function.
+			if (typeof message === 'function') {
+				message = await message(_request, response)
+			}
+
+			// Send the response if headers are not sent.
+			if (!response.headersSent) {
```
gamemaker1, line 162: Could you please use `!response.writableEnded` here instead?

source/lib.ts
```diff
@@ -144,13 +144,24 @@ const parseOptions = (passedOptions: Partial<Options>): Configuration => {
...
-		handler(
+		async handler(
 			_request: Request,
 			response: Response,
 			_next: NextFunction,
 			_optionsUsed: Options,
-		): void {
-			response.status(config.statusCode).send(config.message)
+		): Promise<void> {
+			let { message } = config as { message: unknown }
```
gamemaker1, line 153: Instead of doing this, could you please change the type of `message` in the `types.ts` file [here](https://github.com/nfriedly/express-rate-limit/blob/master/source/types.ts#L189) to `any | ValueDeterminingMiddleware<any>`?

gamemaker1 (changes requested):

colderry:
Hey @gamemaker1, I have made the changes according to your requested changes.

source/lib.ts
nfriedly: I figured there was something better than `headersSent`, Thank you!

source/lib.ts
```diff
@@ -144,13 +144,26 @@ const parseOptions = (passedOptions: Partial<Options>): Configuration => {
...
+			const message: unknown =
+				typeof config.message === 'function'
+					? await (
+							config.message as (request: Request, response: Response) => any
+					  )(request, response)
+					: config.message
+
+			// Send the response if writable.
+			if (!response.writableEnded) {
+				response.send(message || 'Too many requests, please try again later.')
```
nfriedly, line 165: Good call putting a default value in here.

test/library/middleware-test.ts
```diff
@@ -253,6 +253,23 @@ describe('middleware test', () => {
...
+
+		const app = createServer(
+			rateLimit({
+				message,
+				max: 1,
+			}),
+		)
+
+		await request(app).get('/').expect(200, 'Hi there!')
+		await request(app).get('/').expect(429, _message)
```
nfriedly, line 270: This is just a nitpick, but I would probably make this a hard-coded string and get rid of the `_message` variable entirely.

source/lib.ts
gamemaker1: Yes, that is a good catch! Although, could you please use the `??` operator here instead of the `||` operator?

source/lib.ts
```diff
@@ -144,13 +144,26 @@ const parseOptions = (passedOptions: Partial<Options>): Configuration => {
...
-		): void {
-			response.status(config.statusCode).send(config.message)
+		): Promise<void> {
+			// Set the response status code
+			response.status(config.statusCode)
+			// Call the `message` if it is a function.
+			const message: unknown =
+				typeof config.message === 'function'
+					? await (
+							config.message as (request: Request, response: Response) => any
```
gamemaker1, line 159:
```suggestion
							config.message as ValueDeterminingMiddleware<any>
```

nfriedly (approved):
Thank you so much @colderry!

I have a couple of nitpicks below, but I think this looks good overall, so I'm going to go ahead and approve it now. 

1) In the readme, the [description for `handler`](https://github.com/nfriedly/express-rate-limit/blob/91b312eeda103bdf905c4f9ca9fea44fc616b7f0/readme.md#handler) has some example code that no longer matches what the default `handler` actually does. I think that's fine, but perhaps we should change the text to make it more clear that the example isn't actually the default.

```diff
- By default, sends back the `statusCode` and `message` set via the `options`:
+ By default, sends back the `statusCode` and `message` set via the `options`, similar to this:
```

2) The suggestion for `_message` in the test below

@gamemaker1 if it looks good to you, feel free to merge it in and do the `npm version minor` to trigger a release.

colderry:
> Thank you so much @colderry!
> 
> I have a couple of nitpicks below, but I think this looks good overall, so I'm going to go ahead and approve it now.
> 
> 1. In the readme, the [description for `handler`](https://github.com/nfriedly/express-rate-limit/blob/91b312eeda103bdf905c4f9ca9fea44fc616b7f0/readme.md#handler) has some example code that no longer matches what the default `handler` actually does. I think that's fine, but perhaps we should change the text to make it more clear that the example isn't actually the default.
> 
> ```diff
> - By default, sends back the `statusCode` and `message` set via the `options`:
> + By default, sends back the `statusCode` and `message` set via the `options`, similar to this:
> ```
> 
> 2. The suggestion for `_message` in the test below
> 
> @gamemaker1 if it looks good to you, feel free to merge it in and do the `npm version minor` to trigger a release.

You're welcome, @nfriedly.

Also, I have just now made the changes to get rid of the nitpicks you mentioned above.

gamemaker1 (approved):
This looks good to go! Thank you so much @colderry for all the great work :]

gamemaker1:
These external tests take sooooo long to complete :sad:

I'll try playing around with cache settings for npm so it covers the node modules in the `tests/external` folder.

gamemaker1:
I tried creating a release, but npm errored out saying it needed an OTP. @nfriedly could you please update the npm token in GitHub secrets with an automation token, and then push a release? Thanks!

nfriedly:
Op, sorry. Will do.

Issue #326, opened 2022-09-26
Memory leak with jest when adding multiple limiters


RobbeCl:
### Description

When adding **2** rate limiters to express and then executing a jest test with `--detect-leaks` it gives an error that there is a memory leak. Even when resetting the store with the new shutdown method which was added due to [issue 320](https://github.com/nfriedly/express-rate-limit/issues/320).

### Library version

6.6.0

### Node version

v16.10.0

### Module system

CommonJS

RobbeCl:
After some digging in, I found that when sharing an in memory store the reset function is called in the constructor: https://github.com/nfriedly/express-rate-limit/blob/master/source/lib.ts#L236 which causes the previous interval to be overwritten and hence open for memory leaks: https://github.com/nfriedly/express-rate-limit/blob/master/source/memory-store.ts#L65-L67. 

Can I make a PR that edits the init function of memory-store and checks the interval for being used. If so it will cleare it with clearInterval?

nfriedly:
Oh, sorry I never responded to this. For some reason I thought I had.

> Can I make a PR that edits the init function of memory-store and checks the interval for being used. If so it will cleare it with clearInterval?

Yes, that's a good idea.

gamemaker1:
This should be [fixed](https://github.com/express-rate-limit/express-rate-limit/pull/378#discussion_r1303213581) in v7.

PR #327, opened 2022-09-27
add clear interval when using multiple inits

RobbeCl:
## Related Issues
Fixes https://github.com/nfriedly/express-rate-limit/issues/326

### Changed
Added check to memory store that when initialized multiple times, the previous interval is properly cleared. 

## Checklist

      added for the same.

### DIFF
--- a/test/library/memory-store-test.ts
+++ b/test/library/memory-store-test.ts
@@ -132,4 +132,16 @@ describe('memory store test', () => {
 			clearTimeout(realTimeoutId)
 		}
 	})
+
+	it('When calling init it should reset interval when already initialized', async () => {
+		const shutdownSpy = jest.spyOn(MemoryStore.prototype, 'shutdown')
+
+		const store = new MemoryStore()
+
+		store.init({} as Options)
+		expect(shutdownSpy).toHaveBeenCalledTimes(0)
+
+		store.init({} as Options)
+		expect(shutdownSpy).toHaveBeenCalledTimes(1)
+	})
 })
--- a/source/memory-store.ts
+++ b/source/memory-store.ts
@@ -60,6 +60,9 @@ export default class MemoryStore implements Store {
 		// Initialise the hit counter map.
 		this.hits = {}
 
+		// Make sure to reset the previous interval if one was present
+		if (this.interval) this.shutdown()
+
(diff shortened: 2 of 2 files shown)


test/library/memory-store-test.ts
```diff
@@ -132,4 +132,16 @@ describe('memory store test', () => {
 			clearTimeout(realTimeoutId)
 		}
 	})
+
+	it('When calling init it should reset interval when already initialized', async () => {
```
gamemaker1, line 136: nit: could you please rename the test to: "should reset the interval incase it already exists"

test/library/memory-store-test.ts
```diff
@@ -132,4 +132,16 @@ describe('memory store test', () => {
...
+		const shutdownSpy = jest.spyOn(MemoryStore.prototype, 'shutdown')
```
gamemaker1, line 137: Is the `.prototype` here necessary? If not, could you please drop it?

IIRC we should just need to pass the class to the `spyOn` function, no need to mess with prototype.

gamemaker1 (commented):
Hi @RobbeCl,

Thanks for taking the time to make this PR! It looks good to me, just left two suggestions.

Also, is there any way we can make sure there are actually no more memory leaks? Maybe by adding another external test (tests in the `tests/external` folder, which are like mini projects in which to test the library's usage)?

gamemaker1:
Also, it looks like the tests on Node 12 failed :/

nfriedly:
I kinda feel like "you're holding it wrong", `init()` wasn't really meant to be called more than once. But, OK, we can add this in as a protection against that kind of abuse.

I'm not sure what's going on with the test failures - it appears that typescript isn't happy, but I'm not really sure why.

RobbeCl:
@nfriedly So you must create a store per rate limiter then? But if I have like 20 rate limiters and want to close them, I need to call shutdown on each of them. Or is it weird to have 20 of them in the first place? :-)

nfriedly:
@RobbeCl yeah, having 20 if them at the same time is a bit weird. For a testing scenario, I'd create a new instance (of the rate limiter and the store) for each test, and tear it down after the test. That way the behavior of one test isn't dependent on any other test.

PR #330, opened 2022-10-21
6.5.2 and 6.6.0 release notes

ilmartyrk:
Updated source/lib.ts
Change Configuration settings 
from: 
`windowMs: 60 * 1000,`
to
`windowMs: passedOptions.windowMs ?? 60 * 1000,`

This meant that `windowMs `was always hardcoded to be 60 seconds, but not as it should work by the configuration instructions where users could set `windowMs `as optional parameter.

All test are passing

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -6,13 +6,30 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [6.5.0](https://github.com/nfriedly/express-rate-limit/releases/tag/v6.5.0)
+## [6.6.0](https://github.com/nfriedly/express-rate-limit/releases/tag/v6.5.2)
 
-## Changed
+### Added
+
+- Added `shutdown` method to the Store interface and the MemoryStore.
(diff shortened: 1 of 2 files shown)


nfriedly:
I think you're mistaken, the default is `60 * 1000`, but that gets overridden a few lines lower if you pass in a different value in the options:

https://github.com/nfriedly/express-rate-limit/blob/d7a1ddd973a52de0d8e76d16e26dddd24ff6b0ac/source/lib.ts#L176

See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax#spread_in_object_literals for more details on the spread syntax (`...`)

(`notUndefinedOptions` is literally just the `options` you pass in, filtered to remove any keys with a value of `undefined`.)

There are quite a few tests that validate the current behavior, for example

https://github.com/nfriedly/express-rate-limit/blob/d7a1ddd973a52de0d8e76d16e26dddd24ff6b0ac/test/library/middleware-test.ts#L177-L190

That test only covers 60ms; it would fail if the default limit of 6000ms was being used instead of the configured limit of 50ms.

nfriedly:
I'm going to go ahead and close this one.

PR #357, opened 2023-07-06
fix(ts): export diff type files for cjs and esm

### LINKED ISSUE #355: `package.json` exports and types are not compatible with TypeScript's `node16` `moduleResolution`
### Description

TypeScript added a new module resolution strategy `node16` that better mimicks Node.js. Only under this option (or `nodenext`) is the `exports` field evaluated.

When authoring a `.ts` file in a CommonJS project and using this import
```ts
import rateLimit from 'express-rate-limit'
```
TypeScript complains about trying to require an ES module from CommonJS. This is because there's only one `.d.ts` file for both the CJS and ESM entrypoint and TypeScript considers that to be ESM due to the nearest `package.json`.
https://www.typescriptlang.org/docs/handbook/esm-node.html#packagejson-exports-imports-and-self-referencing mentions at the bottom that each needs their own declaration file.

As awkward as this is, this can be fixed by
1. renaming `index.d.ts` to `index.d.mts`
2. copying `index.d.mts` to `index.d.cts`
3. changing the `package.json` exports like this:
```
	"exports": {
		".": {
			"import": {
				"types": "./dist/index.d.mts",
				"default": "./dist/index.mjs"
			},
			"require": {
				"types": "./dist/index.d.cts",
				"default": "./dist/index.cjs"
			}
		}
	},
```

### Library version

6.7.0

### Node version

16.20.0

5.1.6

### Module system

CommonJS

gamemaker1:
## Related Issues

- Fixes #355

### Changed

- Bumps test project dependencies.
- Changes `export` field in package.json to export different `.d.ts` files for CommonJS and ESM projects.

## Checklist

      added for the same.

### DIFF
--- a/test/external/imports/named-import/ts-cjs/source/index.ts
+++ b/test/external/imports/named-import/ts-cjs/source/index.ts
@@ -1,7 +1,7 @@
 // /source/index.ts
 // Run the server
 
-import app from './app'
+import { app } from './app'
 
 app.listen(8080, () =>
 	console.log('Make a GET request to http://localhost:8080!'),
--- a/package-lock.json
+++ b/package-lock.json
@@ -4845,9 +4845,9 @@
 			"dev": true
 		},
 		"node_modules/fast-glob": {
-			"version": "3.2.12",
-			"resolved": "https://registry.npmjs.org/fast-glob/-/fast-glob-3.2.12.tgz",
-			"integrity": "sha512-DVj4CQIYYow0BlaelwK1pHl5n5cRSJfM60UA0zK891sVInoPri2Ekj7+e1CT3/3qxXenpI+nBBmQAcJPJgaj4w==",
+			"version": "3.3.0",
+			"resolved": "https://registry.npmjs.org/fast-glob/-/fast-glob-3.3.0.tgz",
+			"integrity": "sha512-ChDuvbOypPuNjO8yIDf36x7BlZX1smcUMTTcyoIjycexOxd6DFsKsg21qVBzEmr3G7fUKIRy2/psii+CIUt7FA==",
 			"dev": true,
 			"dependencies": {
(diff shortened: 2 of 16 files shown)


test/external/imports/named-import/ts-cjs/source/index.ts
```diff
@@ -1,7 +1,7 @@
 // /source/index.ts
 // Run the server
 
-import app from './app'
+import { app } from './app'
```
nfriedly, line 4: how did that work before...?

nfriedly (approved):
I think we can leave it as-is, but it would have been easier to review this a series of PRs: one for the dependency updates and associated script, one for node16 types fix, and maybe one or two more for the app import statements and autofix tweaks.

nfriedly:
@AlCalzone do you think you can check this branch out and test it locally to confirm that it works for you before we ship it?

It's been a bit since I've done this, but I believe the process is something along the lines of this:

```sh
cd ~
cd express-rate-limit
git checkout fix/diff-dts-files
npm install
npm run compile
npm link
npm link express-rate-limit
```

Once you're done testing the fix, I believe a regular `npm install express-rate-limit` will undo the link.

gamemaker1:
You are right, I should have done that.

I had to update the dependencies to reproduce the bug first, so I added the `upgrade_dependencies` function to the script and did the autofix thing 😅

test/external/imports/named-import/ts-cjs/source/index.ts
gamemaker1: I was shocked too, I have literally no idea how it worked before

AlCalzone:
Seems to be working

nfriedly:
Awesome, v6.7.1 is now live on npm with the fix.

PR #358, opened 2023-07-08
feat: add validation checks

nfriedly:
This is a first stab at the validation checks I talked about in https://github.com/express-rate-limit/express-rate-limit/discussions/356

There's more that I want to do, but wanted to get this bit up for now.


package.json
```diff
@@ -63,8 +63,8 @@
 		"lint:code": "xo --ignore test/external/",
 		"lint:rest": "prettier --ignore-path .gitignore --ignore-unknown --check .",
 		"lint": "run-s lint:*",
-		"autofix:code": "run-s lint:code --fix",
-		"autofix:rest": "run-s lint:rest --write .",
+		"autofix:code": "npm run lint:code -- --fix",
+		"autofix:rest": "npm run lint:rest -- --write .",
```
nfriedly, line 67: I know, I just bitched about including unrelated stuff in a PR, and then I went and did the same thing 🤦

source/types.ts
```diff
@@ -164,6 +164,12 @@ export type Store = {
 	shutdown?: () => Promise<void> | void
 }
 
+export enum ValidationLevel {
+	Off,
+	Warn,
+	Throw,
+}
```
nfriedly, line 171: I'm not sure if this belongs in this file, since it's not actually a type. But I need to reference it in the Options type and this prevented me from needing circular imports.

source/validations.ts
```diff
@@ -0,0 +1,57 @@
...
+		} else if (level === ValidationLevel.Throw) {
+			throw new Error(message)
+		}
+	}
+}
+
+// https://www.regextester.com/104038
+// matches IPv4 and IPv6, but not if they include a port number at the end
+// eslint-disable-next-line @typescript-eslint/naming-convention
+const reIP =
```
nfriedly, line 26: I should probably just import an npm module to do this.. but I kind of like this being a zero-dependency library.

source/validations.ts
```diff
@@ -0,0 +1,57 @@
...
...|25[0-5])\.){3}(\d|[1-9]\d|1\d{2}|2[0-4]\d|25[0-5])$|^(([a-zA-Z]|[a-zA-Z][a-zA-Z\d-]*[a-zA-Z\d])\.)*([A-Za-z]|[A-Za-z][A-Za-z\d-]*[A-Za-z\d])$|^\s*((([\dA-Fa-f]{1,4}:){7}([\dA-Fa-f]{1,4}|:))|(([\dA-Fa-f]{1,4}:){6}(:[\dA-Fa-f]{1,4}|((25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)(\.(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3})|:))|(([\dA-Fa-f]{1,4}:){5}(((:[\dA-Fa-f]{1,4}){1,2})|:((25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)(\.(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3})|:))|(([\dA-Fa-f]{1,4}:){4}(((:[\dA-Fa-f]{1,4}){1,3})|((:[\dA-Fa-f]{1,4})?:((25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)(\.(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3}))|:))|(([\dA-Fa-f]{1,4}:){3}(((:[\dA-Fa-f]{1,4}){1,4})|((:[\dA-Fa-f]{1,4}){0,2}:((25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)(\.(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3}))|:))|(([\dA-Fa-f]{1,4}:){2}(((:[\dA-Fa-f]{1,4}){1,5})|((:[\dA-Fa-f]{1,4}){0,3}:((25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)(\.(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3}))|:))|(([\dA-Fa-f]{1,4}:)(((:[\dA-Fa-f]{1,4}){1,6})|((:[\dA-Fa-f]{1,4}){0,4}:((25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)(\.(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3}))|:))|(:(((:[\dA-Fa-f]{1,4}){1,7})|((:[\dA-Fa-f]{1,4}){0,5}:((25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)(\.(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3}))|:)))(%.+)?\s*$/
```
nfriedly, line 27: I should probably just import an npm module to do this.. but I kind of like this being a zero-dependency library.

package.json
gamemaker1: Lmaooo happens XD

source/types.ts
gamemaker1: The original purpose of this file was to store all public facing types, interfaces and enums, so I'd say it very much does belong here.

Maybe we could change the file name?

source/validations.ts
gamemaker1: Sounds good.

We should keep an eye on the library's issues though, so we would be alerted if this regex is vulnerable to an attack.

gamemaker1 (commented):
Looks good so far :D

Linter doesn't seem very happy though, let me try fixing that.

gamemaker1:
What if we made `validation` a `boolean` and allowed an `onValidationError` function to be passed with the options instead, so the user can decide what to do with the error?

source/types.ts
nfriedly: 🤷 it's probably fine. TS enums are kind of weird in that they're both a type and compile actual JS code. Nearly everything else in TS is one or toe other, rather than both.

nfriedly:
> What if we made validation a boolean and allowed an onValidationError function to be passed with the options instead, so the user can decide what to do with the error?

Yeah, related to that, it occurred to me that `Throw` will have different behavior for validations that run at startup (where throwing likely kills the server) vs in the context of a request (where throwing will be caught by express's error handler and the error message potentially passed to the end user in the HTTP response.) In light of that, I'm not sure I even want to support the `Throw` option.

What if `validation` had 3 potential  values: `false`, a default function that calls `console.warn`, or a custom function.  Although, then the question arises:  should the custom function be expected to handle the request, or the regular path? And, should the regular rate-limiting still apply?

Honestly, that's starting to seem more confusing.  Maybe we should just go with the `validate` setting being a boolean value for now: true = warn, false = skip all checks. We can always add a `Throw` option or an `onValidationError ` setting later if there's demand for it.

----

On an unrelated note, I was thinking that each validation error should have a unique code, like `ERR_ERL_INVALID_IP`. We could set it in the [error.code](https://nodejs.org/docs/latest-carbon/api/errors.html#errors_error_code) property, include it in the error message, and also include some information about it in our wiki. (Of course, `error.code` only matters if we throw or have a callback...)

nfriedly:
Oh, and switching the config option to a boolean would also solve the "enum in the types file" confusion.

nfriedly:
> That makes sense! Should I change it to that then?

Yeah, feel free. I'd like to do some more work on this, but I'm not sure when I'll get to it.

gamemaker1:
Done.

gamemaker1:
Although now we have an `Error` class in the types file :/

source/lib.ts
```diff
@@ -136,12 +171,13 @@ const parseOptions = (passedOptions: Partial<Options>): Configuration => {
 			response.statusCode < 400,
 		skip: (_request: Request, _response: Response): boolean => false,
 		keyGenerator(request: Request, _response: Response): string {
-			if (!request.ip) {
-				console.error(
```
nfriedly, line 140: Maybe we should stick with `console.error` instead of `console.warn` for consistency's sake. 

I originally chose `warn` to make it distinct from the `Throw` option before we got rid of that, but I had also forgotten that we had this check that did a `console.error`.

source/validations.ts
```diff
@@ -0,0 +1,109 @@
...
+ * See https://github.com/express-rate-limit/express-rate-limit/wiki/Troubleshooting-Proxy-Issues.
+ *
+ * @param ip {string | undefined} - The IP address provided by Express as request.ip.
+ *
+ * @returns {void}
+ */
+export const validateIp = (shouldRun: boolean, ip: string | undefined) => {
+	runCheck(
+		shouldRun,
+		() => ipAddressRegex.test(ip ?? 'no-ip'),
```
nfriedly, line 52: When express-rate-limit is used correctly, `req.ip` shouldn't ever be `undefined`, so I don't think we should put that in our type signature. (I wanted the test to check for that, but `undefined as any` should be fine there.)

Then this should work without the `?? 'no-ip'`:

```
> var a
undefined
> /[0-9-]+/.test(a)
false
```
(`undefined` is automatically coerced to a string as needed in JS):
```
> /undefined/.test(a)
true
```

source/validations.ts
```diff
@@ -50,18 +42,20 @@ const ipAddressRegex =
...
+		shouldRun,
 		() => ipAddressRegex.test(ip ?? 'no-ip'),
-		`an invalid 'request.ip' (${
-			ip ?? 'undefined'
-		}) was detected. Consider passing a custom 'keyGenerator' function to the rate limiter.`,
+		new ValidationError(
+			'ERR_ERL_INVALID_IP_ADDRESS',
+			`an invalid 'request.ip' (${
+				ip ?? 'undefined'
+			}) was detected. Consider passing a custom 'keyGenerator' function to the rate limiter.`,
```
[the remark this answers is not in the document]
nfriedly: Oh, `undefined` became a valid option in the method signature. See my other comment about that.

source/types.ts
```diff
@@ -352,3 +327,28 @@ export type RateLimitInfo = {
...
+	 */
+	constructor(code: string, message: string) {
+		super(message)
+
+		// `this.constructor.name` is the class name
+		this.name = this.constructor.name
+		this.code = code
+		this.message = message
+	}
+}
```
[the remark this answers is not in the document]
gamemaker1: Could we still keep the error class, in case we add the `onValidationError` function later on?

source/lib.ts
gamemaker1: Okay, will change to `console.error`

source/validations.ts
gamemaker1: Ideally, yes, `request.ip` would never be undefined. However, there might be a misconfiguration that causes the value to be undefined, and since the purpose of the check is to make sure there's a valid IP, I check for undefined values too.

gamemaker1: The `?? 'no-ip'` part is to please the linter, it doesn't like templates being possibly undefined.

source/types.ts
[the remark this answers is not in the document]
nfriedly: Yeah, I suppose so. I think we could alias the node.js error type here and define it in the validation file

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/82aeb9991c945a2929d4b4d60db1771e5e70dedf/types/node/globals.d.ts#L202-L207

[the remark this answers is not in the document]
gamemaker1: Okay, sure!

source/validations.ts
nfriedly: What if we made it two separate checks?  We could first run the original check to see if `req.ip` is even defined, and then if that passes, run this check to see if it's valid. They're distinct enough issues that having separate error messages probably makes sense.

I'm also thinking about setting up wiki pages (or perhaps a website) for each error with a bit more info, and then linking to it from the error message.

gamemaker1: Sure, that sounds good.

A website seems a bit overkill, I think a wiki page with all the error codes will do. We already have some info in the [Troubleshooting Proxy Issues](https://github.com/express-rate-limit/express-rate-limit/wiki/Troubleshooting-Proxy-Issues) section, we can use that :)

gamemaker1: I added the [Error Codes](https://github.com/express-rate-limit/express-rate-limit/wiki/Error-Codes) page to the wiki, and linked to it in the error messages.

nfriedly: Thank you, that's amazing!

The only reason I was thinking about having a website would be to have shorter URLs; they could even redirect to the wiki. But, I think just putting the wiki link in there is fine for now.

source/lib.ts
```diff
@@ -136,12 +171,13 @@ const parseOptions = (passedOptions: Partial<Options>): Configuration => {
...
-					'WARN | `express-rate-limit` | `request.ip` is undefined. You can avoid this by providing a custom `keyGenerator` function, but it may be indicative of a larger issue.',
-				)
-			}
+			// Run the validation checks on the IP and headers to make sure everything
+			// is working as intended.
+			validateIp(config.validation, request.ip)
+			validateTrustProxy(config.validation, request)
+			validateXForwardedForHeader(config.validation, request)
```
nfriedly, line 178: Looks like we also need to add `checkIfIpIsUndefined(request.ip)` here. Alternatively, it could just be run as part of `validateIp()` and it could return early - if it fails checkIfIpIsUndefined, then it's guarenteed to fail validateIp.

source/validations.ts
```diff
@@ -0,0 +1,150 @@
...
+
+import type { Request } from 'express'
+
+/**
+ * An error thrown/returned when a validation error occurs.
+ */
+class ValidationError extends Error {
+	name: string
+	code: string
+	message: string
```
nfriedly, line 12: I just tested - we get message automatically by extending Error, so we don't need to define it here.

source/validations.ts
```diff
@@ -0,0 +1,150 @@
...
+	shouldRun: boolean,
+	check: () => boolean,
+	error: ValidationError,
+): void => {
+	if (!shouldRun) return
+
+	// Run the check. If it fails, then return the error message to the user.
+	if (!check()) {
+		const formattedMessage = `express-rate-limit: ${error.code} - ${error.message}`
+		console.error(formattedMessage)
```
nfriedly, line 50: IMO, this formatting should be done inside the `ValidationError` constructor, e.g. 

```js
super(`express-rate-limit: ${error.code} - ${error.message}`)
```

And then we could either do `console.error(error.message)` here, or `console.error(error)` to include the full stack trace. I'm a bit on the fence about that one.. do you think stack traces will be more annoying or helpful?

source/validations.ts
```diff
@@ -0,0 +1,150 @@
...
+
+	/**
+	 * The code must be a string, in snake case and all capital, that starts with
+	 * the substring `ERR_ERL_`.
+	 *
+	 * The message must be a string, starting with a lowercase character,
+	 * describing the issue in detail.
+	 */
+	constructor(code: string, message: string) {
+		super(message)
```
nfriedly, line 22:
```suggestion
		super(`express-rate-limit: ${code} - ${message}`)
```

see below

nfriedly:
I left a couple of comments, but I'm going to play with this a bit. I was a little unhappy with building strings on each invocation, but I really don't like creating new error instances for each validation for each request.

nfriedly:
Ok, I reworked the validations into a class that you create an instance of. This way it can keep track of weather or not to run the checks.

I also replaced `runCheck()` with a decorator that allows the checks to create and throw an error when they fail, without creating it when they pass or are disabled. Way less garbage created this way.

And.. ugh. Decorators can't be transformed. It passed the lint and tests locally, but I guess I didn't run a build. Gonna have to rework this some more.

nfriedly:
ok, decorators are out, runCheck is back in. (although I renamed it to `wrap()` since half it's job is to *not* run the checks. build looks to be passing now.

nfriedly:
It just occurred to me that all of these checks are run per-request and can potentially log something on every single request. We might want to limit that. The ones here are probably fine to run on only the first request and none after.

nfriedly:
Ok, it now disables validations near the end of the first request, but logs a full stack trace if an issue is found.

I did a bit of testing and realized that the X-Forwarded-For check didn't work because the default `trust proxy` value is `false`. I updated things to account for this.

I also templatized the more info link part of the error message, since I initially forgot to update it when changing the error code.

I think it needs a little more testing (both manual and automated), and then I think it will be good to ship.

gamemaker1:
This looks great!!

source/validations.ts
```diff
@@ -0,0 +1,123 @@
...
+	ip(ip: string | undefined) {
+		this.wrap(() => {
+			if (ip === undefined) {
+				throw new ValidationError(
+					'ERR_ERL_UNDEFINED_IP_ADDRESS',
+					`An undefined 'request.ip' was detected. This might indicate a misconfiguration or the connection being destroyed prematurely.`,
+				)
+			}
+
+			if (!isIP(ip)) {
```
gamemaker1, line 59: I had thought about using this function, but it also allows port numbers in the IP address so I didn't use it.

readme.md
```diff
@@ -451,6 +451,21 @@ const limiter = rateLimit({
 })
 ```
 
+### `validation`
```
nfriedly, line 454: I'm thinking about renaming this to "validate" - I think it fits better with the other method names, and it's actually what I typed here at first.

source/validations.ts
nfriedly: Oh, does it? I have a couple of tests for that, but maybe I did it wrong. I need to get some sleep, but I'll try and take a closer look tomorrow.

gamemaker1: Nevermind, seems like it does not allow port numbers, so it works :)

readme.md
gamemaker1: I agree, renamed it in ee5c18a

nfriedly:
Ok, I think this is good to go, but lets wait a day or two to see if we want to do a point release with #361 first. I'm leaning towards that, but I'd really like to hear back that it resolves the issue first.

nfriedly:
Alright, I think we've waited long enough. I'm going to ship this.

SimonSchick:
Thank you for this, we did in-fact discover config issues due to this PR :)

Issue #360, opened 2023-07-16
Not callable expression in 6.7.1


hyochan:
### Description

Facing below error after upgrading to `6.7.1`. It is fine again after downgrading to `6.7.0`.

```
src/app.ts:39:20 - error TS2349: This expression is not callable.
      Type 'typeof import("/Users/hyo/Github/app/node_modules/express-rate-limit/dist/index")' has no call signatures.

    39 const apiLimiter = rateLimit({
```

### Library version

6.7.1

### Node version

v18.12.1

### Typescript version (if you are using it)

5.1.6

### Module system

ESM

nfriedly:
Can you show me what yopur code looks like for the import statement?

nfriedly:
Also, I have a theory about what happened - can you check out the `dts-gen-7` branch locally and see if it works for you?

I think the commands are going to be something along these lines:

```
cd ~
git clone https://github.com/express-rate-limit/express-rate-limit.git
cd express-rate-limit
git checkout dts-gen-7
npm install
npm run compile
npm link
cd <directory where your project is>
npm link express-rate-limit
```

(The `npm link` command lets you use a local version of a dependency instead of the one installed from npm - see https://docs.npmjs.com/cli/v9/commands/npm-link for more.)

Once you're done testing the fix, I believe a regular npm install express-rate-limit will undo the link.

nfriedly:
Hey @hyochan were you able to try out the proposed fix? I haven't been able to recreate your issue, so I'm not sure if that PR will make any difference.

I also have a couple of questions:
* Can you show me what your code looks like for the import statement?
  * If it has a complete path in your code, you should try changing it to just `import rateLimit from 'express-rate-limit')`
* Can you show me your command for compiling the TypeScript to JavaScript? (Probably a `build` or `compile` line in the scripts section of your `package.json`)
   * If you're not using `tsc` to compile, can you tell me the version of the compiler you are using?

hyochan:
@nfriedly Sorry for the late response.
The import statement is same `import rateLimit from 'express-rate-limit'`.
It is working fine with `tsc`. It starts to fail in `ts-jest` from `6.7.1` as mentioned above. Not even `6.8.0`.

Below is the screenshot.
![Screenshot 2023-07-26 at 2 35 58 PM](https://github.com/express-rate-limit/express-rate-limit/assets/27461460/419886fd-19b8-46ba-93b2-7e03eb1f5282)

It is working fine in `6.7.0`
[...]

PR #361, opened 2023-07-16
Revert typescript to 4.x and dts-bundle-generator to 7.x

nfriedly:
This is a potential fix for #360 and might be a good idea regardless, as dts-bundle-gen v8.0.0 appears to have [dropped support for typescript <5](https://github.com/timocov/dts-bundle-generator/pull/240). (I should have scrutinized this more closely before including it in a point release...)

Additionally, the types it generates vary slightly depending on which version of typescript is used, so I rolled that back to 4.x as well.

express-rate-limit@6.7.0 used
* typescript@4.8.4
* dts-bundle-generator@7.0.0

express-rate-limit@6.7.1 used
* typescript@5.1.6
* dts-bundle-generator@8.0.1

This PR has 
* typescript@4.9.5
* dts-bundle-generator@7.2.0. 
 
If that doesn't solve the issue, I may revert all the way back to the exact versions used for 6.7.0.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16,7 +16,7 @@
 				"@types/supertest": "2.0.12",
 				"cross-env": "7.0.3",
 				"del-cli": "5.0.0",
-				"dts-bundle-generator": "8.0.1",
+				"dts-bundle-generator": "7.2.0",
 				"esbuild": "0.18.11",
 				"express": "4.18.2",
 				"husky": "8.0.3",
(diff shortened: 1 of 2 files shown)


gamemaker1:
Looks good to me! Although I'm not sure why the external tests didn't catch this when I upgraded their depedencies to use typescript v5.

nfriedly:
I'm still not sure what the actual problem is. I looked over the typescript 5.0.0 breaking changes list again, and I don't see anything that should affect us, but I noticed that that dts-bundle-generator@7 outputs `export declare type` or `export type` for certain things depending on if typescript 4.x or 5.x is used, so I decided to roll back that change too.

Issue #363, opened 2023-07-26
ValidationError: The 'X-Forwarded-For' header is set but the Express


arnabg95:
### Description

I am getting a ValidationError: The 'X-Forwarded-For' header is set but the Express....
the error code: 'ERR_ERL_UNEXPECTED_X_FORWARDED_FOR'

![image](https://github.com/express-rate-limit/express-rate-limit/assets/110101459/a1239fe0-acda-4cf8-a4b2-b47347e274b3)

happens every time a request comes from a new browser window

### Library version

6.8.0

### Node version

18.17.0

### Module system

CommonJS

nfriedly:
Hum, it looks like it's cutting off the end of the error message in your example - it should have a link to https://github.com/express-rate-limit/express-rate-limit/wiki/Error-Codes#err_erl_unexpected_x_forwarded_for yet?

> This error is logged when the X-Forwarded-For header is set (indicating use of a proxy), but the trust proxy setting is false (which is the default value).
> 
> This usually indicates a configuration issue that will cause express-rate-limit to apply it's limits global rather than on a per-user basis. Refer to the [Troubleshooting Proxy Issues](https://github.com/express-rate-limit/express-rate-limit/wiki/Troubleshooting-Proxy-Issues) page for a guide to set the trust proxy value correctly.
> 
> If this error occurs only rarely, and you do not have a reverse proxy, it may indicate a malicious user probing for vulnerabilities.
> 
> This check will be prevented if a custom keyGenerator is supplied.

Basically, it looks like there is a reverse proxy in between your app and the internet and you're rate-limiting it's IP rather than the end-user's IP.

arnabg95:
> Hum, it looks like it's cutting off the end of the error message in your screenshot - it should have a link to https://github.com/express-rate-limit/express-rate-limit/wiki/Error-Codes#err_erl_unexpected_x_forwarded_for
> 
> > This error is logged when the X-Forwarded-For header is set (indicating use of a proxy), but the trust proxy setting is false (which is the default value).
> > This usually indicates a configuration issue that will cause express-rate-limit to apply it's limits global rather than on a per-user basis. Refer to the [Troubleshooting Proxy Issues](https://github.com/express-rate-limit/express-rate-limit/wiki/Troubleshooting-Proxy-Issues) page for a guide to set the trust proxy value correctly.
> > If this error occurs only rarely, and you do not have a reverse proxy, it may indicate a malicious user probing for vulnerabilities.
[...]

PR #365, opened 2023-08-03
fix: include url in `ValidationError`

nfriedly:
I also cleaned up the text a bit since the error code and the fact that it's coming from express-rate-limit are both fairly apparent

## Related Issues

* https://github.com/express-rate-limit/express-rate-limit/issues/363
* https://github.com/express-rate-limit/express-rate-limit/discussions/356#discussioncomment-6551094
* https://github.com/express-rate-limit/express-rate-limit/pull/358

### Changes
This fixes the mistake that caused the help URL to be removed from ValidationError messages. 
Additionally, it adds help a property on the object with the URL, for cases like #363 where the text isn't wrapped around.

I also tweaked the wording to remove redundant info and get to the point more quickly (although this redundant info was never actually shown because of the aforementioned mistake).

Previously errors looked like this:
Screenshot 2023-08-03 at 3 25 41 PM

Now they look like this
Screenshot 2023-08-03 at 4 09 53 PM

## Checklist

      added for the same.

### DIFF
--- a/source/validations.ts
+++ b/source/validations.ts
@@ -20,14 +20,12 @@ class ValidationError extends Error {
 	 * describing the issue in detail.
 	 */
 	constructor(code: string, message: string) {
-		super(
-			`express-rate-limit: ${code} - ${message} See https://github.com/express-rate-limit/express-rate-limit/wiki/Error-Codes#${code.toLowerCase()} for more information on this error.`,
-		)
+		const url = `https://github.com/express-rate-limit/express-rate-limit/wiki/Error-Codes#${code.toLowerCase()}`
+		super(`${message} See ${url} for more information on this error.`)
 
 		// `this.constructor.name` is the class name
 		this.name = this.constructor.name
 		this.code = code
-		this.message = message
 	}
 }
 
--- a/source/validations.ts
+++ b/source/validations.ts
@@ -21,7 +21,7 @@ class ValidationError extends Error {
 	 * describing the issue in detail.
 	 */
 	constructor(code: string, message: string) {
-		const url = `https://express-rate-limit.github.io/errors/#${code.toLowerCase()}`
+		const url = `https://express-rate-limit.github.io/${code.toLowerCase()}`
 		super(`${message} See ${url} for more information on this error.`)
 
 		// `this.constructor.name` is the class name
--- a/source/validations.ts
+++ b/source/validations.ts
@@ -11,6 +11,7 @@ import type { Store } from './types'
 class ValidationError extends Error {
 	name: string
 	code: string
+	help: string
 
 	/**
 	 * The code must be a string, in snake case and all capital, that starts with
@@ -26,6 +27,7 @@ class ValidationError extends Error {
 		// `this.constructor.name` is the class name
 		this.name = this.constructor.name
 		this.code = code
+		this.help = url
 	}
 }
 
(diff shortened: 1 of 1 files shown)


nfriedly:
I still wish we had shorter URLs. It'd be pretty easy to make express-rate-limit.github.io/whatever automatically redirect to https://github.com/express-rate-limit/express-rate-limit/wiki/Error-Codes#whatever

The only annoying part is that you'd have to use a custom 404 page to do it, so the status code would be wrong.

Alternatively, we could generate all of the redirects as separate pages, then they'd have the correct status code. I might do that.

gamemaker1 (approved):

gamemaker1:
> Alternatively, we could generate all of the redirects as separate pages, then they'd have the correct status code. I might do that.

Done, created the https://github.com/express-rate-limit/errors repo for it. The links now look like `https://express-rate-limit.github.io/errors/<code>`.

nfriedly:
> Done, created the https://github.com/express-rate-limit/errors repo for it. The links now look like `https://express-rate-limit.github.io/errors/`.

Oh, wow, that's amazing!

I was thinking we'd use the repo name `express-rate-limit.github.io` so that the URL wouldn't have the `errors` part in it ([*username*.github.io  is treated differently from other repo names](https://pages.github.com/)), and then instead of `err_erl_double_count.html`, it'd be `ERR_ERL_DOUBLE_COUNT/index.html` so that the url would just be `https://express-rate-limit.github.io/ERR_ERL_DOUBLE_COUNT`.

(and we could also make an index page that just redirected `https://express-rate-limit.github.io/` to https://github.com/express-rate-limit/express-rate-limit#readme)

gamemaker1:
> I was thinking we'd use the repo name express-rate-limit.github.io so that the URL wouldn't have the errors part in it ([username.github.io is treated differently from other repo names](https://pages.github.com/)), and then instead of err_erl_double_count.html, it'd be ERR_ERL_DOUBLE_COUNT/index.html so that the url would just be https://express-rate-limit.github.io/ERR_ERL_DOUBLE_COUNT.

That is much smarter, changed the `errors` repo name to `express-rate-limit.github.io`.

> (and we could also make an index page that just redirected https://express-rate-limit.github.io/ to https://github.com/express-rate-limit/express-rate-limit#readme)

Added that too!

gamemaker1:
Should we merge once the checks pass?

nfriedly:
Thanks! 

Yeah, I'll go ahead and merge.

source/validations.ts
```diff
@@ -20,14 +21,13 @@ class ValidationError extends Error {
 	 * describing the issue in detail.
 	 */
 	constructor(code: string, message: string) {
-		super(
-			`express-rate-limit: ${code} - ${message} See https://github.com/express-rate-limit/express-rate-limit/wiki/Error-Codes#${code.toLowerCase()} for more information on this error.`,
-		)
+		const url = `https://express-rate-limit.github.io/${code.toLowerCase()}`
```
nfriedly, line 24: Oh, shoot, I wanted this to be uppercase so that it matched error.code. It was only downloaded because GitHub automatically downcases anchor tags :/

nfriedly:
```suggestion
		const url = `https://express-rate-limit.github.io/${code}/`
```

Might as well put in the trailing slash and spare the user one redirect, also.

nfriedly: I did this fix and then updated express-rate-limit.github.io also so, for example, express-rate-limit.github.io/ERR_ERL_DOUBLE_COUNT works! And, if we ever need to change the URLs or give an error it's own page, it should be easy to do.

gamemaker1: Thanks!!

PR #370, opened 2023-08-10
validations: add `max: 0` change warning

gamemaker1:
## Related Issues

- Implements #369

### Added

- A validation check that warns users that set `max: 0` that the behaviour will change in the not-so-distant future. 

## Checklist

      added for the same.

### DIFF
--- a/source/lib.ts
+++ b/source/lib.ts
@@ -304,9 +304,8 @@ const rateLimit = (
 
 			// Get a unique key for the client
 			const key = await config.keyGenerator(request, response)
-			// Increment the client's hit counter by one
+			// Increment the client's hit counter by one, and make sure it's only by one
 			const { totalHits, resetTime } = await config.store.increment(key)
-
 			config.validations.singleCount(request, config.store, key)
 
 			// Get the quota (max number of hits) for each client
@@ -314,8 +313,9 @@ const rateLimit = (
 				typeof config.max === 'function'
 					? config.max(request, response)
 					: config.max
-
 			const maxHits = await retrieveQuota
+			config.validations.max(maxHits)
+
 			// Set the rate limit information on the augmented request object
 			augmentedRequest[config.requestPropertyName] = {
 				limit: maxHits,
@@ -394,7 +394,7 @@ const rateLimit = (
 
 			// Call the `onLimitReached` callback on the first request where client
 			// exceeds their rate limit
-			// NOTE: `onLimitReached` is deprecated, this should be removed in v7.x
+			// TODO: `onLimitReached` is deprecated, this should be removed in v7.x
 			if (maxHits && totalHits === maxHits + 1) {
 				config.onLimitReached(request, response, options)
 			}
--- a/source/validations.ts
+++ b/source/validations.ts
@@ -202,7 +202,6 @@ export class Validations {
 	 *
 	 * @returns {void}
 	 */
-	// TODO: Remove in v7.0.0
 	max(max: number) {
 		this.wrap(() => {
 			if (max === 0) {
--- a/source/validations.ts
+++ b/source/validations.ts
@@ -35,23 +35,7 @@ class ValidationError extends Error {
  * A warning logged when the configuration used will/has been changed by a
  * newly released version of the library.
  */
-class ChangeWarning extends ValidationError {
-	name: string
-
-	/**
-	 * The code must be a string, in snake case and all capital, that starts with
-	 * the substring `WRN_ERL_`.
-	 *
-	 * The message must be a string, starting with an uppercase character,
-	 * describing the issue in detail.
-	 */
-	constructor(code: string, message: string) {
-		super(code, message)
-
-		// `this.constructor.name` is the class name
-		this.name = this.constructor.name
-	}
-}
+class ChangeWarning extends ValidationError {}
 
 /**
  * The validations that can be run, as well as the methods to run them.
--- a/source/validations.ts
+++ b/source/validations.ts
@@ -17,7 +17,7 @@ class ValidationError extends Error {
 	 * The code must be a string, in snake case and all capital, that starts with
 	 * the substring `ERR_ERL_`.
 	 *
-	 * The message must be a string, starting with a lowercase character,
+	 * The message must be a string, starting with an uppercase character,
 	 * describing the issue in detail.
 	 */
 	constructor(code: string, message: string) {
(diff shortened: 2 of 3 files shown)


source/lib.ts
```diff
@@ -394,7 +394,7 @@ const rateLimit = (
 
 			// Call the `onLimitReached` callback on the first request where client
 			// exceeds their rate limit
-			// NOTE: `onLimitReached` is deprecated, this should be removed in v7.x
+			// TODO: `onLimitReached` is deprecated, this should be removed in v7.x
```
nfriedly, line 397: Oh, this should get a warning too. Doesn't have to be part of this PR, though.

source/validations.ts
```diff
@@ -172,6 +172,25 @@ export class Validations {
...
 
+	/**
+	 * Warns the user that the behaviour for `max: 0` is changing in the next
+	 * major release.
+	 *
+	 * @param max {number} - The maximum number of hits per client.
+	 *
+	 * @returns {void}
+	 */
+	// TODO: Remove in v7.0.0
```
nfriedly, line 183: Rather than removing it, I think we'll just update the language in v7.0.0 to clarify that the behavior has now changed. Maybe remove it in a future 7.x release, but keep it there for now in case someone jumps from 6.9.0 to 7.0.0

source/validations.ts
```diff
@@ -172,6 +172,25 @@ export class Validations {
...
+	max(max: number) {
+		this.wrap(() => {
+			if (max === 0) {
+				console.warn(
+					`The behaviour shown when 'max' is set to zero will change in v7.0.0. Please see https://express-rate-limit.github.io/WRN_MAX/ for more information on this change.`,
```
nfriedly, line 188: Lets make the code `ERL_WRN_MAX_ZERO`

Also, might as well describe the old and new behavior briefly here. Maybe something like this:
```suggestion
					`Setting max to 0 disables rate limiting in in express-rate-limit 6 and older, but will cause all requests to be blocked in v7. See https://express-rate-limit.github.io/ERL_WRN_MAX_ZERO/ for more information on this change.`,
```

Finally, I think it does make sense to different breaking change warnings from other validation errors, but I kind of like the plumbing we have in place that gets a stack trace and formats the URL and whatnot. Perhaps we make a new `ChangeWarning extends ValidationError {}` and throw that, and then the try/catch can do console.error for validation errors and console.warn for change warnings? I'm not sure that's the best answer, but I think it's worth considering a bit more.

nfriedly: Oh, and it doesn't have to be part of the error message, but in the wiki we can suggest that users create a custom `skip()` function to replicate the behavior of setting max to 0 without getting warned.

source/validations.ts
gamemaker1: That makes sense, alright!

source/validations.ts
gamemaker1: Okay!

gamemaker1:
Changed the code, added the ChangeWarning class, and updated the wiki!

source/validations.ts
```diff
@@ -31,6 +31,28 @@ class ValidationError extends Error {
...
+	 * the substring `WRN_ERL_`.
+	 *
+	 * The message must be a string, starting with an uppercase character,
+	 * describing the issue in detail.
+	 */
+	constructor(code: string, message: string) {
+		super(code, message)
+
+		// `this.constructor.name` is the class name
+		this.name = this.constructor.name
```
nfriedly, line 52: Is this needed? I think the one in ValidationError would set it correctly. (Because `this` will be the ChangeWarning instance even in the ValidationError constructor (?)

in fact, I'm not sure we even need a constructor.

gamemaker1:
Should we merge this?

source/validations.ts
```diff
@@ -17,7 +17,7 @@ class ValidationError extends Error {
 	 * The code must be a string, in snake case and all capital, that starts with
 	 * the substring `ERR_ERL_`.
 	 *
-	 * The message must be a string, starting with a lowercase character,
+	 * The message must be a string, starting with an uppercase character,
 	 * describing the issue in detail.
 	 */
 	constructor(code: string, message: string) {
```
nfriedly, line 23: Oh, we should update the default message in the `ValidationError` constructor from 
`... See ${url} for more information on this error.` 
to just 
`... See ${url} for more information.`
since it now applies to both warnings and errors.

nfriedly (approved):
I like it! A couple of thoughts below, but otherwise this is good to go.

I've been thinking about how to let certain validations run on more than just the first request, and I think this warning would be a good candidate for that. But I'm not quite sure how I want to implement it. (My original plan involved using decorators, but esbuild doesn't support them.)

source/validations.ts
gamemaker1: It's needed since it sets the name of the error to `ChangeWarning`.

gamemaker1: If we don't call it here I think the name would be `ValidationError`.

source/validations.ts
gamemaker1: Done

source/validations.ts
gamemaker1: Turns out I am wrong, the constructor and this stuff isn't needed. Will change it, thanks!

gamemaker1:
I need to catch some sleep right now, will get back to you with some suggestions on how to implement the multiple-time-validation check tomorrow :)

Also I'll merge this once the checks complete.

source/validations.ts
cyrille1804-a11y: It seems like the same GitHub Copilot documentation has been sent multiple times across this conversation, alongside a web security guide. Could you clarify what you’d like me to do? Are you trying to build a carrousel from one or both of these documents?​​​​​​​​​​​​​​​​

source/lib.ts
cyrille1804-a11y: It seems like the same GitHub Copilot documentation has been sent multiple times across this conversation, alongside a web security guide. Could you clarify what you’d like me to do? Are you trying to build a carrousel from one or both of these documents?​​​​​​​​​​​​​​​​

source/validations.ts
cyrille1804-a11y: It seems like the same GitHub Copilot documentation has been sent multiple times across this conversation, alongside a web security guide. Could you clarify what you’d like me to do? Are you trying to build a carrousel from one or both of these documents?​​​​​​​​​​​​​​​​

source/validations.ts
cyrille1804-a11y: It seems like the same GitHub Copilot documentation has been sent multiple times across this conversation, alongside a web security guide. Could you clarify what you’d like me to do? Are you trying to build a carrousel from one or both of these documents?​​​​​​​​​​​​​​​​

source/validations.ts
cyrille1804-a11y: It seems like the same GitHub Copilot documentation has been sent multiple times across this conversation, alongside a web security guide. Could you clarify what you’d like me to do? Are you trying to build a carrousel from one or both of these documents?​​​​​​​​​​​​​​​​

source/validations.ts
cyrille1804-a11y: It seems like the same GitHub Copilot documentation has been sent multiple times across this conversation, alongside a web security guide. Could you clarify what you’d like me to do? Are you trying to build a carrousel from one or both of these documents?​​​​​​​​​​​​​​​​

PR #371, opened 2023-08-10
build: use `pnpm` for external tests instead of `npm`

gamemaker1:
## Related Issues

- Implements https://github.com/express-rate-limit/express-rate-limit/pull/364#issuecomment-1663888914

### Changed

- Uses `pnpm` to install dependencies during the external tests. It reduces the running time for the longest job from [~12 minutes](https://github.com/express-rate-limit/express-rate-limit/actions/runs/5792235842/usage) to [~5 minutes](https://github.com/express-rate-limit/express-rate-limit/actions/runs/5820302393/usage), and the total cpu time from 1 hour to 30 minutes.

## Checklist

      added for the same.


package.json
```diff
@@ -52,7 +52,8 @@
 		"changelog.md"
 	],
 	"engines": {
-		"node": ">= 14.0.0"
+		"node": ">= 14",
+		"pnpm": "7"
```
nfriedly, line 56: Why is this here? Does it prevent end-users from using older pnpm versions to install the package? Do we know that doesn't work?

nfriedly (approved):
Couple of questions, but this looks like a big win overall

.github/workflows/ci.yaml
```diff
@@ -22,7 +22,7 @@ jobs:
           node-version: ${{ matrix.node-version }}
       - name: Check for lint/formatting errors
         run: |
-          npm ci
+          npm install
```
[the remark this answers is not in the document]
gamemaker1: Oops I forgot to revert this :/

I had actually changed everything to use pnpm, but then reverted to using npm for the main project. `pnpm` does a ci-install by default so it doesn't have a `ci` command, and I forgot to change it back.

package.json
nfriedly: hum... but we're only using pnpm for the external tests, that we don't even recommend that people run locally. I kind of think we should just put v7 in our lockfile and omit this.

Additionally/Alternatively, we could skip running the ext tests on node 14 - I actually just made a [change to the test matrix for express-slow-down](https://github.com/express-rate-limit/express-slow-down/pull/43/files#diff-7829468e86c1cc5d5133195b5cb48e1ff6c75e3e9203777f6b2e379d9e4882b3) such that it runs linting linux and node lts/* (currently 18), it runs all tests on node lts/* and node latest (currently 20) on win/lin/mac, and then it runs the tests on node 14 on linux. 

We could do something similar here where we run everything on node lts/* and latest, and then run just the library tests on node 14 on linux. Then we could sidestep the pnpm problems with node 14, and just use pnpm v8.

What do you think about that?

gamemaker1: Alright, sure!

Just to confirm:

- `lint` runs on linux + node lts
- `test:library` runs on linux/win/mac + node 14/16/18/20
- `test:external` runs on linux/win/mac + node lts/latest

gamemaker1: PS: npm has the same problem with lockfile v2 on node 14 vs v3 on 16/18/20; our current lockfile on the main branch is v2 from before the rewrite so it still works lol

nfriedly: > *    lint runs on linux + node lts
> *    test:library runs on linux/win/mac + node 14/16/18/20
> *    test:external runs on linux/win/mac + node lts/latest

Yeah, that works

gamemaker1: > test:external runs on linux/win/mac + node lts/latest

We run the external tests only on linux right now, I'm *not* changing that to linux/windows/mac then.

PR #378, opened 2023-08-23
feat!: improve `MemoryStore`

nfriedly:
I've been thinking about how to make the MemoryStore have per-user reset times and still only a single global timer (like `precise-memory-rate-limit`), but *not* have to iterate through all known clients to remove expired ones (unlike `precise-memory-rate-limit`).

I came up with the idea for a previous and current set of clients, where we could pull active ones forward from previous to current, and then at the end of the time window, just blow away the previous, move the current to previous, and create a new current set.

I then refined it a bit to perform less allocations (and by extension, need less garbage collection) by merely swapping previous and current, and then moving the leftover clients from the old previous to a client pool. With an appropriately sized pool, it should be able to completely avoid allocations for most requests.

I had to tweak the tests to use a positive number for `windowMs`, but after that they all passed.

I've also been thinking about a 'ClusterMemoryStore' where there would be a MemoryStore instance on the primary process, and then each worker would communicate with it via IPC to increment and decrement keys.  Then you would only need an external data store for multi-server deployments. Probably should go in it's own module, though.

### DIFF
--- a/source/memory-store.ts
+++ b/source/memory-store.ts
@@ -188,17 +188,13 @@ export default class MemoryStore implements Store {
 	}
 
 	/**
-	 * Clear out expired clients in previous, then move all current clients to previous.
+	 * Move current clients to previous, create a new map for current.
 	 *
 	 * This function is called every `windowMs`.
 	 */
 	private clearExpired(): void {
-		// At this point, all clients in previous are expired, so delete them all
-		this.previous.clear()
-
-		// Swap previous and current so that current clients are moved to previous and current will have a blank slate
-		const temporary = this.previous
+		// At this point, all clients in previous are expired
 		this.previous = this.current
-		this.current = temporary
+		this.current = new Map()
 	}
 }
--- a/source/memory-store.ts
+++ b/source/memory-store.ts
@@ -200,8 +200,9 @@ export default class MemoryStore implements Store {
 
 		let client
 		if (this.previous.has(key)) {
-			// If it's in the `previous` map, return it and bump it up to the `current` map.
+			// If it's in the `previous` map, take it out
 			client = this.previous.get(key)!
+			this.previous.delete(key)
 		} else if (this.pool.length > 0) {
 			// If it's in neither the `current` nor the `previous` maps, animate a corpse
 			// from the pool. Spoooooooooooky!
--- a/source/memory-store.ts
+++ b/source/memory-store.ts
@@ -8,6 +8,10 @@ type Client = {
 	totalHits: number
 	resetTime: Date
 }
+
+const average = (array: number[]) =>
+	array.reduce((a, b) => a + b) / array.length
+
 /**
  * A `Store` that stores the hit count for each client in memory.
  *
@@ -23,22 +27,30 @@ export default class MemoryStore implements Store {
 	 * These two maps store usage (requests) and reset time by key (IP)
 	 *
 	 * They are split into two to avoid having to iterate through the entire set to determine which ones need reset.
-	 * Instead, Clients are moved from previous to current as new requests come in from them.
-	 * Once windowMS has elapsed, all clients left in previous are reset at once.
+	 * Instead, Clients are moved from previous to current as new requests come in.
+	 * Once windowMS has elapsed, all clients left in previous are known to be expired.
+	 * At that point, the cache pool is filled from previous, and any remaining clients are cleared.
 	 *
 	 */
-	previous!: Map<string, Client>
-	current!: Map<string, Client>
+	previous = new Map<string, Client>()
+	current = new Map<string, Client>()
 
 	/**
-	 * Pool of unused clients, kept to reduce the number of objects created and destroyed
+	 * Cache of unused clients, kept to reduce the number of objects created and destroyed.
+	 *
+	 * Improves performance, at the expense of a small amount of RAM.
+	 *
+	 * Each individual entry takes up 152 bytes.
+	 * In one benchmark, the total time taken to handle 100M requests was reduced from 70.184s to 47.862s (~32% improvement) with ~5.022 MB extra RAM used.
+	 * .
 	 */
-	pool!: Client[]
+	pool: Client[] = []
 
 	/**
-	 * Maximum number of unused clients to keep in the pool
+	 * Used to calculate how many entries to keep in the pool
 	 */
-	poolSize = 100
+	newClients = 0
+	recentNew: number[] = []
 
 	/**
 	 * Reference to the active timer.
@@ -52,20 +64,6 @@ export default class MemoryStore implements Store {
 	localKeys = true
 
 	/**
-	 * Create a new MemoryStore with an optional custom poolSize
-	 *
-	 * Note that the windowMS option is passed to init() by express-rate-limit
-	 *
-	 * @param [options]
-	 * @param [options.poolSize] - Maximum number of unused objects to keep around. Increase to reduce garbage collection.
-	 */
-	constructor({ poolSize }: { poolSize?: number } = {}) {
-		if (typeof poolSize === 'number') {
-			this.poolSize = poolSize
-		}
-	}
-
-	/**
 	 * Method that initializes the store.
 	 *
 	 * @param options {Options} - The options used to setup the middleware.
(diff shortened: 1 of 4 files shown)


nfriedly:
(Sorry I've been kind of all over the place recently. I'll try and get a few of these things wrapped up sometime soon...)

source/memory-store.ts
```diff
@@ -60,17 +73,25 @@ export default class MemoryStore implements Store {
...
-		this.interval = setInterval(async () => {
-			await this.resetAll()
+		// Initialize the spare client pool
+		this.pool = []
+
+		// Indicates that init was called more than once.
+		// Could happen if a store was shared between multiple instances.
+		if (this.interval) {
+			clearTimeout(this.interval)
+		}
```
nfriedly, line 88: This fixes #326

gamemaker1:
Wow! This looks really promising, I'll try reviewing it in detail soon.

source/memory-store.ts
```diff
@@ -30,16 +20,25 @@ export default class MemoryStore implements Store {
...
-	 * The time at which all hit counts will be reset.
+	 * Pool of unused clients, kept to reduce the number of objects created and destroyed
 	 */
-	resetTime!: Date
+	pool!: Client[]
+
+	/**
+	 * Maximum number of unused clients to keep in the pool
+	 */
+	poolSize = 100
```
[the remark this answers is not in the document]
nfriedly: Maybe it should take [`os.freemem()`](https://nodejs.org/api/os.html#osfreemem) into account also, since it's essentially trading RAM usage for performance.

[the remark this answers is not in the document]
nfriedly: Actually, inspecting with chrome dev tools shows that a single client object is only ~152 bytes, so 60k would be less than 1mb. So `os.freemem()` won't really have much affect.

I put together a quick test https://gist.github.com/nfriedly/fdba25bcf9f23d18fe8ac5bd1bad5dac and 5000 seems to be the winner from a few runs, but it honestly doesn't make much difference either way.

source/memory-store.ts
```diff
@@ -30,16 +24,33 @@ export default class MemoryStore implements Store {
...
+	current = new Map<string, Client>()
 
 	/**
-	 * The time at which all hit counts will be reset.
+	 * Cache of unused clients, kept to reduce the number of objects created and destroyed.
+	 *
+	 * Improves performance, at the expense of a small amount of RAM.
+	 *
+	 * Each individual entry takes up 152 bytes.
+	 * In one benchmark, the total time taken to handle 100M requests was reduced from 70.184s to 47.862s (~32% improvement) with ~5.022 MB extra RAM used.
```
gamemaker1, line 44: Should we add this benchmark to the repo as `test/benchmark/memory-store.benchmark.ts`?

source/memory-store.ts
```diff
@@ -102,9 +117,9 @@ export default class MemoryStore implements Store {
 	 * @public
 	 */
 	async decrement(key: string): Promise<void> {
-		const current = this.hits[key]
+		const client = this.getClient(key)
 
-		if (current) this.hits[key] = current - 1
+		if (client.totalHits > 1) client.totalHits--
```
gamemaker1, line 122: Shouldn't this be `> 0`?

gamemaker1, line 122: Shouldn't this be `> 0`?

source/memory-store.ts
```diff
@@ -136,5 +152,72 @@ export default class MemoryStore implements Store {
...
+		this.newClients = 0
+		const targetPoolSize = Math.round(average(this.recentNew))
+
+		// Calculate how many entries to potentially copy to the pool
+		let poolSpace = targetPoolSize - this.pool.length
+
+		// Fill up the pool with expired clients
+		for (const client of this.previous.values()) {
+			if (poolSpace > 0) {
+				this.pool.push(client)
```
gamemaker1, line 181: Should we run `resetClient` on the client when adding it to the pool? That way, the pool might take up less resources.

source/memory-store.ts
nfriedly: No, at least not in its current state. It's pretty ad hoc, and coupled to the implementation. I was making changes to the testing script as I changed the implementation details, so that's not even the current iteration.

Also it's very dependent on the machine you run it on, so I don't think it would be very useful to compare across machines.

source/memory-store.ts
nfriedly: No, because we don't want it to ever go negative.

source/memory-store.ts
nfriedly: Nah, it takes up the same amount of resources either way, and we have to reset the date when we pull it out of the pool, so it would just be doing extra work to also reset it when we put it in.

source/memory-store.ts
gamemaker1: Oh alright then

source/memory-store.ts
gamemaker1: But if the number of hits is 1 it should be able to decrement it to 0 right?

source/memory-store.ts
gamemaker1: Ah I didn't consider the date part, makes sense not to do it here then.

source/memory-store.ts
```diff
@@ -155,69 +167,96 @@ export default class MemoryStore implements Store {
...
+		// If we already have a client for that key in the `current` map, return it.
+		if (this.current.has(key)) return this.current.get(key)!
 
 		let client
 		if (this.previous.has(key)) {
+			// If it's in the `previous` map, return it and bump it up to the `current` map.
 			client = this.previous.get(key)!
 		} else if (this.pool.length > 0) {
+			// If it's in neither the `current` nor the `previous` maps, animate a corpse
+			// from the pool. Spoooooooooooky!
```
nfriedly, line 207: 👻 😂

source/memory-store.ts
```diff
@@ -30,19 +36,37 @@ export default class MemoryStore implements Store {
...
+	 */
+	pool: Client[] = []
 
 	/**
-	 * The time at which all hit counts will be reset.
+	 * Used to calculate how many entries to keep in the pool.
 	 */
-	resetTime!: Date
+	numberOfCreatedClients = 0
+	recentClientsCount: number[] = []
```
nfriedly, line 66: Hum.. I concede that the old names weren't great, but this doesn't feel right either.

The first one is the number of keys that we've seen during the current window that we didn't see during the previous window. Ideally pool should begin the window with exactly this many Client objects, so that 0 are created.

The second one is a series of previous values for the first. So it should probably just be the plural form of the name of the first one. 

Some ideas: 
* `idealPoolSize` & `idealPoolSizes`?
* `currentNewKeys` & `previousNewKeys`
* `uncachedCount` & `uncachedCounts`

I don't know. Naming things is hard.

source/memory-store.ts
```diff
@@ -155,69 +167,96 @@ export default class MemoryStore implements Store {
...
-		if (this.current.has(key)) {
-			return this.current.get(key)!
-		}
+		// If we already have a client for that key in the `current` map, return it.
+		if (this.current.has(key)) return this.current.get(key)!
 
 		let client
 		if (this.previous.has(key)) {
+			// If it's in the `previous` map, return it and bump it up to the `current` map.
 			client = this.previous.get(key)!
```
nfriedly, line 204: Oh, I think there's a bug here. Client should be *removed* from `previous`, this just gets a reference to it. The client could potentially end up in the `pool` and `current` at the same time, and then get assigned to a new key, thus ending up in `current` more than once and combining the hits from both keys.

I should write a test to make that scenario happen, and then fix the code.

nfriedly: Fixed in e0801ce

source/memory-store.ts
gamemaker1: `uncachedCount` & `uncachedCounts` sounds good 👍🏾

source/memory-store.ts
nfriedly: 🤦 yes.

source/memory-store.ts
nfriedly: done in d89e514

nfriedly:
Ugh, I need to redo the performance testing, that bug invalidated everything I did.

nfriedly:
Ok, now that I fixed the bug with not deleting clients from previous when moving to current, I re-tested and found that the `pool` actually makes things slightly slower. So I removed it.

I also updated the Use Cases section of the readme although, honestly, we could probably delete it at this point. I wrote that because of how often people were confused by the behavior of the old MemoryStore, and I think the new one will just do what they expect.

nfriedly:
Ok, I believe this is now ready to go. 

It's simplified a lot, because my earlier implementation was too clever by half.

gamemaker1 (approved):
LGTM!

> Ok, now that I fixed the bug with not deleting clients from previous when moving to current, I re-tested and found that the pool actually makes things slightly slower. So I removed it.

Looks like we didn't need to play with ghosts after all XD

PR #380, opened 2023-08-31
fix!: `max: 0` should rate limits all requests

gamemaker1:
## Overview

This PR fixes #259. Once this PR is merged, setting the `max` option to zero will 'block' all requests to that endpoint.

---

### DIFF
--- a/source/lib.ts
+++ b/source/lib.ts
@@ -401,7 +401,7 @@ const rateLimit = (
 			config.validations.disable()
 
 			// If the client has exceeded their rate limit, set the Retry-After header
-			// and call the `handler` function
+			// and call the `handler` function.
 			if (typeof maxHits === 'number' && totalHits > maxHits) {
 				if (config.legacyHeaders || config.standardHeaders) {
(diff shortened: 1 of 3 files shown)


source/lib.ts
```diff
@@ -411,7 +411,7 @@ const rateLimit = (
 			// Call the `onLimitReached` callback on the first request where client
 			// exceeds their rate limit
 			// TODO: `onLimitReached` is deprecated, this should be removed in v7.x
-			if (maxHits && totalHits === maxHits + 1) {
+			if (typeof maxHits === 'number' && totalHits === maxHits + 1) {
```
nfriedly, line 414: Isn't `maxHits` always a number? (See [line 350](https://github.com/express-rate-limit/express-rate-limit/pull/380/files#diff-5c2516e9822679491492c832b9ace79b6945bdd891586e07b0232ebef1c7e0b1L350))

gamemaker1: Oh yea that's really silly of me lol

gamemaker1: Fixed in c7433c2

PR #381, opened 2023-08-31
feat: rename `max` to `limit` while still supporting `max`

gamemaker1:
## Overview

Renames the `max` option to `limit`, but sill supports the deprecated `max` option.

---

This PR should probably be merged last.


test/library/options-test.ts
```diff
@@ -50,6 +37,30 @@ describe('options test', () => {
 		expect(store.options.draft_polli_ratelimit_headers).toEqual(true)
 	})
 
+	// TODO: Update in v8.
```
nfriedly, line 40: I think we're probably going to keep `max` around forever. It's been both correct and more-or-less required for essentially every release up to now, so I think removing it or even adding warnings is probably too much.

test/library/options-test.ts
```diff
@@ -50,6 +37,30 @@ describe('options test', () => {
...
+	it('should allow the use of the `max` option', async () => {
+		const store = new MockStore()
+		rateLimit({
+			store,
+			max: 7,
+		})
+
+		expect(store.options.limit).toEqual(7)
```
nfriedly, line 48: hah, that's a clever way to check it!

test/library/options-test.ts
gamemaker1: Makes sense, removed the comment in 2a0d80d.

readme.md
```diff
@@ -247,7 +247,7 @@ twice, once here and once on the store. In some cases the units also differ
 
 Defaults to `60000` ms (= 1 minute).
 
-### `max`
+### `limit`
```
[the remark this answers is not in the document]
gamemaker1: Done in a4d08d2.

test/library/options-test.ts
```diff
@@ -37,7 +37,6 @@ describe('options test', () => {
 		expect(store.options.draft_polli_ratelimit_headers).toEqual(true)
 	})
 
-	// TODO: Update in v8.
```
nfriedly: never gonna give you up

readme.md
```diff
@@ -259,6 +259,8 @@ Express `request` and `response` objects and then returns a number.
 
 Defaults to `5`. Set it to `0` to disable the rate limiter.
 
+> Renamed in v7.x from `max` to `limit`.
```
nfriedly, line 262: What about

```suggestion
Renamed in v7.x from `max` to `limit`. (However, `max` is still supported for backwards compatibility.)
```

gamemaker1: 👍🏾

PR #382, opened 2023-08-31
chore: drop deprecated options

gamemaker1:
## Overview

Once this PR is merged, setting the `onLimitReached` and `draft_polli_ratelimit_headers` option will have no effect, apart from triggering a `ChangeWarning` validation error.

---


nfriedly (approved):
Nice work

PR #383, opened 2023-08-31
build: drop node 14

gamemaker1:
## Overview

This PR excludes node 14 from the CI, and bumps the typescript build target to `es2022`. 

---


nfriedly (approved):
Man, you're on fire! 🔥

PR #387, opened 2023-09-03
chore: move jest config to standard location

nfriedly:
Per https://jestjs.io/docs/configuration moving the jest config to `jest.cnfig.json`  allows it to be automatically discovered. This in turn makes the jest plugin for vs code "just work" (presumably other editors too), allowing me to easily run or debug individual tests.

I also cleaned up the command line to execute the tests, because I don't think we need any of those flags any more. 

Finally, I added the various external tests package-lock.json's to the .gitignore because they started showing up for some reason.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,5 @@ coverage/
 *.tmp
 *.bak
 *.tgz
+
+test/external/imports/*/*/package-lock.json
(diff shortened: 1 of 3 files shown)


gamemaker1:
Nice, thanks!

PR #390, opened 2023-09-05
feat: add `get` method to the `Store` interface

### LINKED ISSUE #389: Get Remaining Rate-Limit
### Description

The ability to get the remaining rate-limit from the default MemoryStore using the key (i.e IP Address) rather than relying on response headers.

### Why

In the cases of APIs, other parts of the system or users outside the user being ratelimited may need to know how much of a ratelimit is remaining and will not necessarily be able to see the response headers provided.

### Alternatives

Nothing currently, if this already exists somewhere and I did not find it, please let me know and I will close this.

gamemaker1:
- The function is also bound to the middleware so it can be called externally.

---

This PR fixes #389. The tests are currently failing since the external stores do not yet have the `fetchKey` function.

### DIFF
--- a/source/types.ts
+++ b/source/types.ts
@@ -67,7 +67,7 @@ export type RateLimitReachedEventHandler = (
  * @property totalHits {number} - The number of hits for that client so far.
  * @property resetTime {Date | undefined} - The time when the counter resets.
  */
-export type IncrementResponse = {
+export type ClientRateLimitInfo = {
 	totalHits: number
 	resetTime: Date | undefined
 }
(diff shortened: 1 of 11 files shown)


source/types.ts
```diff
@@ -153,6 +167,20 @@ export type Store = {
...
 	resetKey: (key: string) => Promise<void> | void
 
+	/**
+	 * Method to fetch a client's hit count and reset time.
+	 *
+	 * @param key {string} - The identifier for a client.
+	 *
+	 * @returns {ClientRateLimitInfo} - The number of hits and reset time for that client.
+	 */
+	fetchKey: (
```
nfriedly, line 177: I think it should be optional so that older stores aren't invalidated by this change.

Also, I feel like `get()` might be a better name; I feel like "fetch" implies a network which isn't necessarily happening here, and then dropping 'key' makes it more inline with `increment()` and `decrement()`. 

(I know we have `resetKey()`, but that is to differentiate it from `resetAll()`. Otherwise I might have just called it `reset()`)

```suggestion
	get?: (
```

gamemaker1: > I think it should be optional so that older stores aren't invalidated by this change.

Yea I was thinking the same 😅 

> Also, I think get() might be a better name; I feel like "fetch" implies a network request, which isn't necessarily happening here, and then dropping 'key' makes it more inline with increment() and decrement().

Okay!

gamemaker1:
Tests pass now, so yay!

If this PR looks good to go, I'll merge it into the `next` branch, then cherry pick the commit onto the `main` branch and release v6.11.

PR #391, opened 2023-09-06
fix: rename current to used, while still supporting current

gamemaker1:
- This PR implements https://github.com/express-rate-limit/express-rate-limit/discussions/372#discussioncomment-6915685.

---

We should probably merge this in v7.

### DIFF
--- a/source/types.ts
+++ b/source/types.ts
@@ -328,7 +328,15 @@ export type AugmentedRequest = Request & {
  */
 export type RateLimitInfo = {
 	limit: number
-	current: number
+	used: number
 	remaining: number
 	resetTime: Date | undefined
+
+	/**
+	 * NOTE: The `current` field is deprecated and renamed to `used`. The library
+	 * will still set the `current` property, and you can still access it, but it
+	 * will be hidden from iteration and JSON.stringify calls. See:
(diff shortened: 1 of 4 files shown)


source/types.ts
```diff
@@ -328,7 +328,15 @@ export type AugmentedRequest = Request & {
...
 	remaining: number
 	resetTime: Date | undefined
+
+	/**
+	 * NOTE: The `current` field is deprecated and renamed to `used`. The library
+	 * will still set the `current` property, and you can still access it, but it
+	 * will be hidden from iteration and JSON.stringify calls. See:
+	 * https://github.com/express-rate-limit/express-rate-limit/discussions/372#discussioncomment-6915685
+	 */
+	// current: number
```
nfriedly, line 341: Perhaps we should document it as optional and @deprecated?

```suggestion
   * @deprecated
	 */
	current?: number
```

I don't know, maybe your way is better. I don't think typescript has a good way of handling this.

nfriedly (approved):
One thought on typing, but this is probably fine to ship as-is.

source/types.ts
gamemaker1: I didn't do the deprecation thing because we were hiding the fact that the property even existed, and we didn't want anyone to use it.

PR #392, opened 2023-09-06
Allow specific validations to be enabled or disabled

nfriedly:
This allows the `validate` option to be set to an object with keys for specific validation checks and boolean values to enable or disable the check. There is also a special `default` key that is used as the fallback value for unset keys. It defaults to `true`. 

It ended up being a kinda big change, but it also fixed the annoyance I had of needing to call `wrap()` inside of each validation check - they're now wrapped automatically - without using decorators!

Todo:
* [x] document change in readme
* [ ] document change on https://github.com/express-rate-limit/express-rate-limit/wiki/Error-Codes to indicate which key to set to prevent each code
    * This should wait until after v7 is published and/or clarify that v7.0.0 or newer is required for this feature.
* [x] rename `max` check to `limit` so that it has the `limit` key to disable it in  the config
* [x] Fix the `ValidationsEnabled` type to only include the allowed keys
* [x] Tests for various `ValidationsEnabled` configurations
* [x] Tests for the new `validationsConfig()` validation check
* [x] Document `ERR_ERL_UNKNOWN_VALIDATION` on https://github.com/express-rate-limit/express-rate-limit/wiki/Error-Codes
    * See https://github.com/express-rate-limit/express-rate-limit/wiki/Error-Codes#err_erl_unknown_validation
* [x] Redirect for `ERR_ERL_UNKNOWN_VALIDATION`

Should unblock https://github.com/express-rate-limit/express-slow-down/pull/43

Tip: review  with white space changes omitted: https://github.com/express-rate-limit/express-rate-limit/pull/392/files?w=1

---

I had a couple of other  thoughts that are outside the scope of this PR:
* It'd be nice for Stores to be able to hook into the validations logic, although it'd add a lot of complexity. It'd also probably break intelisense suggestions for `ValidationsEnabled` values. I think we should probably hold off on this for now.
* Some validations might make sense to run on more than the first request. We could allow  `ValidationsEnabled` values to be positive integers, and then decrement it each time the check is run until it gets to `0`. `Infinity` would allow the checks to run indefinitely. I'm not yet sure how I'd handle setting the default value, though.

I think we can ship v7.0.0 without either of those features.

### DIFF
--- a/source/types.ts
+++ b/source/types.ts
@@ -175,6 +175,12 @@ export type Store = {
 
 export type DraftHeadersVersion = 'draft-6' | 'draft-7'
 
+// Ideally, this should be:
+//   import type { Validations } from './validations.js'
+//   export type ValidationsEnabled = { [key in keyof Validations]: boolean }
+// But that causes a circular reference between the Validations and ValidationsEnabled types, so it doesn't work.
+export type ValidationsEnabled = { [key: string]: boolean }
+
 /**
  * The configuration options for the rate limiter.
  */
@@ -291,7 +297,7 @@ export type Options = {
 	/**
 	 * Whether or not the validation checks should run.
 	 */
-	validate: boolean
+	validate: boolean | ValidationsEnabled
 
 	/**
 	 * Whether to send `X-RateLimit-*` headers with the rate limit and the number
--- a/source/types.ts
+++ b/source/types.ts
@@ -177,7 +177,7 @@ export type Store = {
 export type DraftHeadersVersion = 'draft-6' | 'draft-7'
 
 /**
- * Validations configuration object for enabling or disabeling specific validations
+ * Validate configuration object for enabling or disabeling specific validations
  *
  * The keys must also be keys in the validations object (except `enabled` and `disable`), or `default`
  */
--- a/source/types.ts
+++ b/source/types.ts
@@ -2,6 +2,7 @@
 // All the types used by this package
 
 import type { Request, Response, NextFunction, RequestHandler } from 'express'
+import type { Validations } from './validations.js'
 
 /**
  * Callback that fires when a client's hit counter is incremented.
@@ -175,11 +176,14 @@ export type Store = {
 
 export type DraftHeadersVersion = 'draft-6' | 'draft-7'
 
-// Ideally, this should be:
-//   import type { Validations } from './validations.js'
-//   export type ValidationsEnabled = { [key in keyof Validations]: boolean }
-// But that causes a circular reference between the Validations and ValidationsEnabled types, so it doesn't work.
-export type ValidationsEnabled = { [key: string]: boolean }
+/**
+ * Validations configuration object for enabling or disabeling specific validations
+ *
+ * The keys must also be keys in the validations object (except `enabled` and `disable`), or `default`
+ */
+export type ValidationsEnabled = {
+	[key in keyof Omit<Validations, 'enabled' | 'disable'> | 'default']?: boolean
+}
 
 /**
  * The configuration options for the rate limiter.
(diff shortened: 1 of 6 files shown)


nfriedly:
I think I'm going to rename the `max` validation to `limit`.

That means that this change should be coupled to whatever release gets support for `limit`.

source/types.ts
```diff
@@ -175,6 +176,15 @@ export type Store = {
 
 export type DraftHeadersVersion = 'draft-6' | 'draft-7'
 
+/**
+ * Validate configuration object for enabling or disabeling specific validations
+ *
+ * The keys must also be keys in the validations object (except `enabled` and `disable`), or `default`
+ */
+export type ValidationsEnabled = {
+	[key in keyof Omit<Validations, 'enabled' | 'disable'> | 'default']?: boolean
```
nfriedly, line 185: I suppose technically `disable` could be valid here - setting it to `false` would cause validations to run on every request. I don't think I'm ready to support that just yet, though...

test/library/validation-test.ts
```diff
@@ -233,9 +233,60 @@ describe('validations tests', () => {
...
+			validations = getValidations({ invalid: false } as any)
+			validations.validationsConfig()
+			expect(console.error).toBeCalled()
+		})
+		it('should log an error if an unknown validation is enabled', () => {
+			validations = getValidations({ invalid: false } as any)
+			validations.validationsConfig()
+			expect(console.error).toHaveBeenCalledWith(
+				expect.objectContaining({ code: 'ERR_ERL_UNKNOWN_VALIDATION' }),
+			)
```
nfriedly, line 247: We should probably do this for all the tests in this file, but not as part of this PR

nfriedly:
Ok, I think this is ready. Please review whenever you have a good chance @gamemaker1. Ignoring white space-only changes will make `validations.ts` a bit easier to review: https://github.com/express-rate-limit/express-rate-limit/pull/392/files?w=1

source/types.ts
gamemaker1: No I think we shouldn't allow `disable` here, it's better for the users to just specify `false` instead.

gamemaker1: Also, `default` is not allowed as a key here, but is used in [the readme example](https://github.com/express-rate-limit/express-rate-limit/pull/392/files#diff-5a831ea67cf5cf8703b0de46901ab25bd191f56b320053be9332d9a3b0d01d15R528)?

test/library/validation-test.ts
gamemaker1: 👍🏾

gamemaker1 (commented):
LGTM, apart from the `default` thing I left a comment about above.

Amazing work!!

source/types.ts
nfriedly: > No I think we shouldn't allow disable here, it's better for the users to just specify false instead.

Just to clarify
* `false` = 0 runs
* `true` = 1 run
* `{disable: false}` = Infinity runs 

> Also, default is not allowed as a key here, but is used in [the readme example](https://github.com/express-rate-limit/express-rate-limit/pull/392/files#diff-5a831ea67cf5cf8703b0de46901ab25bd191f56b320053be9332d9a3b0d01d15R528)?

I think `default` is allowed by the `| 'default'` at the end (?)

gamemaker1: > I think default is allowed by the | 'default' at the end (?)

Oh I misread that line, so sorry!

gamemaker1: > Just to clarify

> false = 0 runs
> true = 1 run
> {disable: false} = Infinity runs

Oh. It seems weird at first but then kinda makes sense.

I agree with you on doing this sometime later.

gamemaker1 (approved):

PR #394, opened 2023-09-10
fix: consider store prefix in `singleCount` validation check

### LINKED ISSUE #395: ERR_ERL_DOUBLE_COUNT with multiple rate limits
### Discussed in https://github.com/express-rate-limit/express-rate-limit/discussions/393

<div type='discussions-op-text'>

<sup>Originally posted by **stephent** September  9, 2023</sup>
Before this warning was added, I'd had multiple rate limits configured for a long time, all pointing to the same Redis store. With the addition of the warning, the double counting is now apparent.

However, wouldn't it be a friendlier approach to have the library look after unique scoping keys per configured rate limit rather than imposing this on the library user? If you're able to configure multiple rate limits, I suspect many would naturally assume this would already be the case by default, but it seems not.

(Not sure if there are other use cases which make this difficult or undesirable to implement?)

UDPATE: on reviewing my code more thoroughly, I see I already have unique `prefix` values per rate limit for the RedisStore. Confused as to why I'm seeing this error...</div>

nfriedly:
This makes prefix an optional property on the store Interface (memcached and redis already use it), and then if set, prepends it to keys for the singleCount validation.

Fixes #395 (see #393)

### DIFF
--- a/source/types.ts
+++ b/source/types.ts
@@ -199,6 +199,13 @@ export type Store = {
 	 * Used to help detect double-counting misconfigurations.
 	 */
 	localKeys?: boolean
+
+	/**
+	 * Optional value that the store prepends to keys
+	 *
+	 * Used by the double-count check to avoid false-positives when a key is counted twice, but with different prefixes
+	 */
+	prefix?: string
 }
 
 export type DraftHeadersVersion = 'draft-6' | 'draft-7'
--- a/source/validations.ts
+++ b/source/validations.ts
@@ -183,14 +183,16 @@ export class Validations {
 				storeKeys.set(storeKey, keys)
 			}
 
-			if (keys.includes(key)) {
+			const prefixedKey = `${store.prefix ?? ''}${key}`
+
+			if (keys.includes(prefixedKey)) {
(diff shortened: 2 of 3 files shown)


nfriedly:
@stephent would you mind trying this out with your previous configuration and confirm that it no longer logs erroneous double count errors?

I think the commands are going to be something along these lines:

```
cd ~
cd express-rate-limit
git checkout prefix-double-count
npm install
npm run compile
npm link
npm link express-rate-limit
```

Once you're done testing the fix, I believe a regular npm install express-rate-limit will undo the link.

gamemaker1 (approved):

gamemaker1:
Nice!

gamemaker1:
Should we release this as v6.12?

nfriedly:
> Should we release this as a patch version, v6.11.1?

Yeah, I was initially thinking we should wait a day or two to give @stephent a chance to verify it, but I am fairly confident in this fix. So, go ahead and release it if you'd like. 

Patch sounds right to me.

gamemaker1:
Done!

PR #396, opened 2023-09-11
chore: v7.0.0 release

nfriedly:
I took the `next` branch, rebased against `main`, fixed several conflicts, and updated the changelog. I decided to do a PR so that we could both give it an extra pass before merging into `main` and releasing.

This PR should be merged rather than "squash and merge"d


readme.md
```diff
@@ -547,10 +565,12 @@ if you wish to create your own store.
 ## Request API
 
 A `request.rateLimit` property is added to all requests with the `limit`,
```
nfriedly, line 567: I'm fine with full names in our codebase, but perhaps we should rename `request` to `req` in our docs to match Express?

> In this documentation and by convention, the object is always referred to as req (and the HTTP response is res) 

http://expressjs.com/en/4x/api.html#req

gamemaker1: Ohkay, sure!

gamemaker1 (approved):
LGTM!

Let's get this released soon! :shipit:

gamemaker1:
Let me know if I should merge the PR

nfriedly:
Yeah, I think this is ready to go. Would you like to do the honors?

gamemaker1:
It would be my pleasure :D

PR #397, opened 2023-09-12
fix: restore IncrementResponse type

nfriedly:
Restores `IncrementResponse ` as an alias of `ClientRateLimitInfo`

fixes https://github.com/express-rate-limit/rate-limit-redis/pull/192 (hopefully)

### DIFF
--- a/source/types.ts
+++ b/source/types.ts
@@ -72,6 +72,8 @@ export type ClientRateLimitInfo = {
 	resetTime: Date | undefined
 }
 
+export type IncrementResponse = ClientRateLimitInfo
+
 /**
  * A modified Express request handler with the rate limit functions.
  */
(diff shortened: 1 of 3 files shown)


nfriedly:
@shenie would you mind trying this out and confirm that it fixes the typing error?

I think the commands are going to be something along these lines:

```
cd ~
cd express-rate-limit
git checkout fix-type
npm install
npm run compile
npm link
npm link express-rate-limit
```

Once you're done testing the fix, I believe a regular `npm install express-rate-limit` will undo the link.

nfriedly:
I'd also like to figure out why our integration test didn't catch this...

shenie:
> @shenie would you mind trying this out and confirm that it fixes the typing error?

done. it worked

nfriedly:
> done. it worked

awesome; I'll go ahead and ship this now and look into the tests later

nfriedly:
Ok, 6.11.2 is now live on npm with the fix. 

Thanks for bring it to our attention and helping me test it!

PR #400, opened 2023-09-15
docs: Added rate-limit-postgresql reference in readme.

adrianprelipcean:
See https://github.com/express-rate-limit/express-rate-limit/discussions/399#discussioncomment-7012726


nfriedly:
Thanks! I'll merge this in now; I'm not sure if we'll do a release just for this or if we'll include it with the next functional change. What do you think @gamemaker1 ?

gamemaker1:
I don't think we'll be doing a functional release anytime soon, so we could release this as a patch version.

Issue #404, opened 2023-10-01
getKey is undefined in Redis Store


Rohit3523:
### Description

When i am trying to get a key from ratelimit with getKey function, I am getting `ratelimit.getKey` is not a function

Here is the example code
```
const authRequestRateLimit = rateLimit({
    max: 10,
    windowMs: 60000 * 10,
    keyGenerator: (req) => {
        return `authRequestRateLimit:${req.ip}`
    },
    store: new RedisStore({
        sendCommand: (...args) => client.sendCommand(args)
    })
});

authRequestRateLimit.getKey("authRequestRateLimit:127.0.0.1")
```

When i logged authRequestRateLimit, It is showing getKey: undefined
```
[AsyncFunction (anonymous)] {
  resetKey: [Function: bound resetKey] AsyncFunction,
  getKey: undefined
}
```

### Library version

7.0.2

### Node version

18.13.0

### Module system

CommonJS

nfriedly:
Apologies; `getKey` was a recent addition and it depends on store support. I believe only the memoryStore supports it right now. I'm going to update the readme to make that more clear. 

Would you like to send in a PR to add it to the redis store?

Rohit3523:
I would like to add support for getKey in redis store but i am not sure how the store thing work

nfriedly:
Ok, I'll see if I can do it, but it probably won't be today.

Rohit3523:
Okay will wait for the update 🙂

Rohit3523:
I guess you will have to change lot of things in redis because it only store hitCount in redis, not the resetTime

Rohit3523:
Any update?

nfriedly:
https://github.com/express-rate-limit/rate-limit-redis/pull/195

Rohit3523:
Noticed a minor bug, It is returning NaN instead of 0 if the store doesn't have the key we are passing

gamemaker1:
Sorry about that, I released v4.1.1 with a fix for the issue.

Rohit3523:
Thanks for the quick fix 🙂

PR #405, opened 2023-10-01
feat: make getKey throw if the store doesn't have a get method

nfriedly:
Still needs to be tested.

Relates to #404


gamemaker1:
@nfriedly Could you please review the changes I made?

nfriedly (commented):
Yep, this looks good. I'm going to fill in the change log, and then we can merge this and put out a release.

Issue #418, opened 2023-11-05
Ratelimit headers empty while running on Bun v1.0.x


MikkelSVDK:
### Description

I've been creating a little project with bun as the runtime env, but I've noticed that the headers are always empty, even though the info variable has the information as seen in the photo below.
![image](https://github.com/express-rate-limit/express-rate-limit/assets/31735044/f8031fa5-67dc-46c0-adb9-2e5e971bf8c8)

I simply added .toString() and it return the correct information in the headers
![image](https://github.com/express-rate-limit/express-rate-limit/assets/31735044/515b93c9-4c9a-4df5-8822-46ce31eb87b0)

### Library version

7.1.3

### Node version

bun 1.0.7

### Typescript version (if you are using it)

6.0.0

### Module system

ESM

nfriedly:
Fixed by #419

PR #419, opened 2023-11-05
fix: Ratelimit headers empty while running on Bun v1.0.x #418

MikkelSVDK:
Related to #418

### DIFF
--- a/source/headers.ts
+++ b/source/headers.ts
@@ -41,8 +41,8 @@ export const setLegacyHeaders = (
 ): void => {
 	if (response.headersSent) return
 
-	response.setHeader('X-RateLimit-Limit', info.limit)
-	response.setHeader('X-RateLimit-Remaining', info.remaining)
+	response.setHeader('X-RateLimit-Limit', info.limit.toString())
+	response.setHeader('X-RateLimit-Remaining', info.remaining.toString())
 
 	// If we have a resetTime, also provide the current date to help avoid
 	// issues with incorrect clocks.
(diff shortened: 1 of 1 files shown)


nfriedly (approved):
This looks good, I think we can merge it in as-is.

However, to ensure it doesn't regress, do you think you can add automated tests in bun? Ideally run the full suite of unit tests in bun, but if that turns out to be unfeasible then perhaps just a few bun-speciffic ones, at least something that would catch this specific issue. Could be part of this PR or in a separate one.

nfriedly:
Oh, and then `npm run format` will fix most lint issues automatically.

MikkelSVDK:
Yes, i'll make a new PR with some tests. However im not quite sure what scheme i should follow for the tests. Should I create a folder test/bun/*.test.ts or just put them with the others, since jests only looks for *-test.?

Bun only has one requirement `Tests need ".test", "_test_", ".spec" or "_spec_" in the filename`

nfriedly:
I think it's possible to tweak the jest config to use whatever format we want. If we renamed the existing test, would they work in bun?

nfriedly:
OK, this will get published in v7.1.4 shortly. You might also want to file a bug on bun, since they [claim full compatibility with the `node:http` module](https://bun.sh/docs/runtime/nodejs-apis#node-http), and it defines [`setHeader` as accepting any value](https://nodejs.org/api/http.html#requestsetheadername-value).

PR #420, opened 2023-11-08
chore: move changelog to docs folder/mintlify

nfriedly:
Also removed the changelog and some other files from what gets published to npm. I think it's pretty rare for users to look in `node_modules` for documentation (especially for something like this that's designed for online use), and npm automatically includes the bits it needs (like the readme & package.json), so we don't need to specify those. 

It won't make a huge difference in file size, but multiplied by millions of downloads per week, it starts to add up to something significant.

(I originally noticed that we had a [broken link to docs/changelog.md](https://github.com/express-rate-limit/express-rate-limit/blob/v7.1.4/docs/quickstart/installation.mdx#L54) so I fixed that and some other broken links and pushed the fix to `main` right away, but then I thought it probably would make sense to move the changelog over to the docs.)

### DIFF
--- a/docs/guides/contributing.mdx
+++ b/docs/guides/contributing.mdx
@@ -67,6 +67,8 @@ express-rate-limit/
 ├── config/
 │  └── husky/
 │     └── pre-commit # runs the linter on staged files
+├── docs/
+│  └── * # documentation & changelog
 ├── source/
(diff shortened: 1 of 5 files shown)


gamemaker1 (approved):
LGTM!

PR #423, opened 2023-11-20
fix: ci generation config of the url for changelog

Martii:
* Affects https://github.com/express-rate-limit/express-rate-limit/releases pages
* Assume directing devs/users to offsite area instead of [current dev location](https://github.com/express-rate-limit/express-rate-limit/blob/main/docs/reference/changelog.mdx)

Post #420


nfriedly:
Oops, I missed that. Thanks!

PR #426, opened 2023-11-27
Bugfix: requestWasSuccessful usage doesn't support returning a Promise

azjgard:
Fixes https://github.com/express-rate-limit/express-rate-limit/issues/425

The added test case fails on `main` but succeeds on this branch.


nfriedly:
Oh, good catch, thank you!

nfriedly:
Should be released as v7.1.5 shortly.

azjgard:
So quick, thank you!

PR #438, opened 2024-02-07
Check for instance creation while handling a request

nfriedly:
This is one of the ideas I wrote down on https://github.com/express-rate-limit/express-rate-limit/discussions/436 - it seems to work but I'm not yet certain I want to go with it. Also, I think it needs a better name than 'middleware'.

### DIFF
--- a/docs/reference/configuration.mdx
+++ b/docs/reference/configuration.mdx
@@ -348,6 +348,7 @@ Supported validations are:
 - [draftPolliHeaders](/reference/error-codes#wrn-erl-deprecated-draft-polli-headers)
 - [onLimitReached](/reference/error-codes#wrn-erl-deprecated-on-limit-reached)
 - [headersResetTime](/reference/error-codes#err-erl-headers-no-reset)
+- [creationStack](/reference/error-codes#err-erl-created-in-request-handler)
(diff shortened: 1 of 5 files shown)


gamemaker1:
Maybe we could call it `improperCreation`?

nfriedly:
> Maybe we could call it improperCreation?

I don't know.. the other validations are named after either the thing we're checking (e.g. `ip`), or the result we're expecting (`positiveHits`), so I don't think I want to name this one after the thing we don't want.

I have a local change where I renamed it to `creationStack`, which I like a little better. `creationCallSite` might also work. Still not 100% though.

gamemaker1:
`callPoint`? `callStack`? otherwise `creationStack` sounds good.

nfriedly:
Ok, I pushed the `creationStack` rename, and some tests. 

I think the help docs for this one is going to need to be a bit move detailed.

nfriedly:
Ok, I still need to set up a redirect for the error code, but otherwise I think this is ready to go.

I might want to make a few readme tweaks before we ship it, but they're unrelated to this PR.

gamemaker1 (approved):
LGTM! Great work :D

PR #440, opened 2024-02-20
docs: add Configuration & Thank You sections to readme

nfriedly:
I've been thinking that it would be good to have a quick overview of the supported configuration options in the readme. This lists them with types, a short description, and a link to the full docs for each one. 

I considered adding more complete signatures for the options that accept functions, but I felt like it was a bit too much.

I'm a bit on the fence about the Thank You section, but I think it's probably a good thing overall.

What do you think @gamemaker1 ?

This is the last bit I was thinking I wanted to merge in before putting out a new semver-minor release. (I suppose we also need to update the changelog...)

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -9,16 +9,6 @@
 
 </div>
 
----
-
-Sponsored by [Zuplo](https://zuplo.link/express-rate-limit) a fully-managed API
-Gateway for developers. Add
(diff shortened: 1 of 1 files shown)


gamemaker1:
I like the thank you section. Also, I converted the options list into a table, do you think that looks better?

gamemaker1 (approved):

PR #443, opened 2024-03-08
docs: update the create-a-store guide

nfriedly:
A couple of updates to the [creating a store guide](https://express-rate-limit.mintlify.app/guides/creating-a-store):

* call out `prefix` and `localKeys` fields
* make the sample code block wider

The second part required removing some of the formatting. It honestly looked nicer before, but I think this is more functional, because you don't have to scroll left-and-right as much to read the example code.

I also looked for a way to have the sample code in it's own standalone .ts and .js files, but couldn't figure out a way to do that and automatically import them into the document, and I think having example code in this document is more valuable. Reducing the level of indentation at least helped.

Context: https://github.com/express-rate-limit/express-rate-limit/discussions/442#discussioncomment-8721789

Hiding whitespace changes may make reviewing easier: https://github.com/express-rate-limit/express-rate-limit/pull/443/files?diff=split&w=1

### DIFF
--- a/docs/guides/creating-a-store.mdx
+++ b/docs/guides/creating-a-store.mdx
@@ -46,7 +46,7 @@ hits per hour). Keys in the database should be prefixed with this value.
 validation check also takes the `prefix` field into account and does not report
 that a user is being double-counted if the stores have different prefixes.)
 
-The `localKeys` field is an alternative to `prefix` for Stores such as the
+The `localKeys` field is an alternative to `prefix` for stores such as the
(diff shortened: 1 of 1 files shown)


nfriedly:
Oh, cool, mintlify automatically deployed a staging instance, so you can compare [before](https://3000.code.home.nfriedly.com/guides/creating-a-store) and [after](https://express-rate-limit-rework-create-store-guide.mintlify.app/guides/creating-a-store)

gamemaker1 (approved):
This is great, thanks!!

PR #444, opened 2024-03-14
fix: update readme to reflect current new issue

iennae:
This is just a readme update; I noticed on https://www.npmjs.com/package/express-rate-limit?activeTab=readme that the readme for filing an issue was sending to a 404 when I was checking out the package.


gamemaker1 (approved):
Thank you!!

PR #447, opened 2024-03-20
feat: Added passOnStoreError to skip rate limitter if the store is not available

Gp2mv3:
This PR adds the parameter passOnStoreError to allow requests if the store is not available.
For instance, if using Redis and the Redis server disconnects, the request will still pass.

This feature can be important for HA issues.

### DIFF
--- a/test/library/middleware-test.ts
+++ b/test/library/middleware-test.ts
@@ -803,26 +803,27 @@ describe('middleware test', () => {
 			saveRequestObject,
 			rateLimit({
 				legacyHeaders: false,
+				limit: 6,
 			}),
 		])
 
 		await request(app).get('/').expect(200)
 		expect(savedRequestObject?.rateLimit).toMatchObject({
-			limit: 5,
+			limit: 6,
 			used: 1,
-			remaining: 4,
+			remaining: 5,
 			resetTime: expect.any(Date),
 		})
 
-		// Make sure the hidden proerty is also set.
+		// Make sure the hidden property is also set.
 		expect(savedRequestObject?.rateLimit.current).toBe(1)
 
 		savedRequestObject = undefined
 		await request(app).get('/').expect(200)
 		expect(savedRequestObject?.rateLimit).toMatchObject({
-			limit: 5,
+			limit: 6,
 			used: 2,
-			remaining: 3,
+			remaining: 4,
 			resetTime: expect.any(Date),
 		})
 		expect(savedRequestObject?.rateLimit.current).toBe(2)
--- a/test/library/options-test.ts
+++ b/test/library/options-test.ts
@@ -49,6 +49,16 @@ describe('options test', () => {
 		expect(store.options.limit).toEqual(7)
 	})
 
+	it('should allow the use of the `limit` option', async () => {
+		const store = new MockStore()
+		rateLimit({
+			store,
+			limit: 7,
+		})
+
+		expect(store.options.limit).toEqual(7)
+	})
+
 	it('should not allow the use of an invalid store', async () => {
 		class InvalidStore {
 			invalid = true


nfriedly (changes requested):
Good idea, thanks!


A few thoughts:

1. There's a lot of "stuff" inside the try/catch besides the store.increment() call, so I think it should either be renamed to `passOnError` or else the try/catch should be more narrowly scoped to only the increment call.
    * I'm leaning towards the rename option, but I could see a good argument for the other side.
2. If we go with the rename option, it should probably hook into the existing `handleAsyncErrors`: https://github.com/express-rate-limit/express-rate-limit/blob/e3e3ca9f672fa5a0708fa62e5c1358c2cbdcecc3/source/lib.ts#L284
3. Regardless, I think it should log the error, not just silently allow the request.

What do you think @gamemaker1 ?

nfriedly:
Oh, and this PR closes https://github.com/express-rate-limit/rate-limit-redis/pull/193

gamemaker1:
Hi @gp2mv3, thanks for making this PR!

> 1. There's a lot of "stuff" inside the try/catch besides the store.increment() call, so I think it should either be renamed to `passOnError` or else the try/catch should be more narrowly scoped to only the increment call.

I think we should keep it `passOnStoreError`, and scope it to the increment call as you said.

The `passOnStoreError` option has already been requested in `rate-limit-redis`, and I think it'd be better if we just provide it as part of `express-rate-limit`.

> 3. Regardless, I think it should log the error, not just silently allow the request.

I agree, we should make this addition to `handleAsyncErrors` regardless of the `passOnStoreError` option.

gamemaker1:
By point 3, we mean logging the error before calling `next()`.

Also logging the error in `handleAsyncError` can be done in a separate pr.

nfriedly:
> Also logging the error in `handleAsyncError` can be done in a separate pr.

`handleAsyncError` currently passes the error onto express.js to handle however it's configured to (show an error page by default), so no need to change it unless we add an option that ignores those errors.

https://github.com/express-rate-limit/express-rate-limit/blob/e3e3ca9f672fa5a0708fa62e5c1358c2cbdcecc3/source/lib.ts#L291

source/lib.ts
```diff
@@ -324,11 +326,24 @@ const rateLimit = (
...
+
+			let totalHits = 0
+			let resetTime
+			try {
+				const incrementResult = await config.store.increment(key)
+				totalHits = incrementResult.totalHits
+				resetTime = incrementResult.resetTime
+			} catch (error) {
+				if (config.passOnStoreError) {
+					next()
```
nfriedly, line 341:
```suggestion
					console.error('express-rate-limit: error from store, allowing request without rate-limiting.', error)
					next()
```

This is what I meant by point 3.

test/library/middleware-test.ts
```diff
@@ -943,4 +961,25 @@ describe('middleware test', () => {
...
+	it('should pass if the store throws an error and passOnStoreError is true', async () => {
+		const app = createServer(
+			rateLimit({
+				limit: 1,
+				store: new StoreThrowingErrors(),
+				passOnStoreError: true,
+			}),
+		)
+		await request(app).get('/').expect(200)
+	})
```
nfriedly, line 984: I forgot to say it last time, but thank you for adding these tests!

nfriedly (approved):

nfriedly:
Oops, I wasn't trying to close this. I checked it out locally to fix the lint errors, and then I also rebased it and changed the `throw` to a `next(error)`, although I want to test that a bit more. 

Also, we need to add documentation for this

nfriedly:
Re-opening

PR #456, opened 2024-05-09
meta : export getValidations function

bejewel-kyoungmin:
```
		keyGenerator(request: Request, _response: Response): string {
			// Run the validation checks on the IP and headers to make sure everything
			// is working as intended.
			validations.ip(request.ip)
			validations.trustProxy(request)
			validations.xForwardedForHeader(request)

			// By default, use the IP address to rate limit users.
			// eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
			return request.ip!
		},
```
above code is default keyGenerator.
I wanna make custom keyGenerator that is based on default keyGenerator.
For using validations, I have to call `getValidations` function.
But getValidations is not exported.
That's why I made this pr. Thank you : )

### DIFF
--- a/source/index.ts
+++ b/source/index.ts
@@ -11,3 +11,6 @@ export { default, default as rateLimit } from './lib.js'
 // Export the memory store in case someone wants to use or extend it
 // (see https://github.com/nfriedly/express-rate-limit/issues/289)
 export { default as MemoryStore } from './memory-store.js'
+
+// Export getValidations function
+export { getValidations } from './validations.js'


gamemaker1:
Hi @bejewel-kyoungmin,

Thanks for making the PR!

The validations in the default `keyGenerator` are only to be called if you are using the IP address of the user as the key. You do not need to call these validations if you are defining a custom key generator (which should ideally use something more reliable than an IP address, like an authentication token).

That said, I believe it would be beneficial to work on exporting the validation API;

1. for users to build on it
2. for better validation support in express-slow-down, which uses this package under the hood

What do you think, @nfriedly?

nfriedly:
Yea, my thinking with not exporting the validations part was that if you're supplying your own functions, then you're probably a more advanced user that is unlikely to hit the issues the validations check for, and also as @gamemaker1 said, the validations may not even apply in your use case.

I've had the idea of moving the validations code to a standalone library, but I haven't fully fleshed out what that would look like and what all it would provide. (If you omit the parts that are specific to this library, there's not that much code left.)

bejewel-kyoungmin:
```
validations.ip(request.ip)
validations.trustProxy(request)
validations.xForwardedForHeader(request)
```

when I first met this code, I think these validations make sense.
so I wanna take these validations for my own key.
I'm considering to make key like this `${ip}.${otherkey}` 
It is not hard to copy and paste that validation code to my codebase.
I respect you guys opinions. and I appreciate making useful package.

PR #458, opened 2024-05-19
docs: prefer limit over max

x-dune:
v7 prefers the naming limit over max. This aligns the preference on the quickstart docs.


nfriedly:
Good catch, thank you!

PR #459, opened 2024-05-28
feat: storePerLimiter validation check

nfriedly:
This checks for cases like https://github.com/express-rate-limit/express-rate-limit/discussions/457 where a single Store instance is shared across multiple rate limiter instances.

I saw there was an unrelated formatting change to the one test, so I updated my node modules and got a vulnerability warning, so I went ahead and fixed that too.

Example error messages:

> ValidationError: A Store instance must not be shared across multiple rate limiters. Create a new instance of MemoryStore for each limiter instead. See https://express-rate-limit.github.io/ERR_ERL_STORE_REUSE/ for more information.

Example for a store that doesn't have `localKeys` set:

> ValidationError: A Store instance must not be shared across multiple rate limiters. Create a new instance of ClusterMemoryStoreWorker (with a unique prefix) for each limiter instead. See https://express-rate-limit.github.io/ERR_ERL_STORE_REUSE/ for more information.

It feels a little wordy to me, but I think it gets the idea across.

### DIFF
--- a/source/validations.ts
+++ b/source/validations.ts
@@ -150,7 +150,7 @@ const validations = {
 	/**
 	 * Ensures a single store instance is not used with multiple express-rate-limit instances
 	 */
-	storePerLimiter(store: Store) {
+	unsharedStore(store: Store) {
 		if (usedStores.has(store)) {
 			const maybeUniquePrefix = store?.localKeys
 				? ''
--- a/source/validations.ts
+++ b/source/validations.ts
@@ -38,6 +38,11 @@ class ValidationError extends Error {
 class ChangeWarning extends ValidationError {}
 
 /**
+ * List of store instances that have been used with any express-rate-limit instance
+ */
+const usedStores = new Set<Store>()
+
+/**
  * Maps the key used in a store for a certain request, and ensures that the
  * same key isn't used more than once per request.
  *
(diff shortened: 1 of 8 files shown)


source/validations.ts
```diff
@@ -142,6 +147,23 @@ const validations = {
 		}
 	},
 
+	/**
+	 * Ensures a single store instance is not used with multiple express-rate-limit instances
+	 */
+	storePerLimiter(store: Store) {
```
gamemaker1, line 153: I think a more appropriate name (in accordance with the naming of the rest of the validations) might be `singleStorePerLimiter`. What do you think?

gamemaker1 (approved):
LGTM! Just that one suggestion about the name, otherwise looks good to go :shipit:

source/validations.ts
nfriedly: Hum, I think `singleStorePerLimiter` implies that it's protecting against multiple stores being passed to a single limiter, when it's really the other way around. Maybe `singleLimiterPerStore`, or just `limiterPerStore`. Or maybe `uniqueStore` or `ownStore`?

nfriedly: `freshStore` or `freshStoreInstance`?

nfriedly: Some other ideas:
* `separateStore`
* `dedicatedStore`
* `unsharedStore`

gamemaker1: `unsharedStore` sounds good!

PR #461, opened 2024-06-05
fix: narrow scope of creationStack check to only stores with localKeys (e.g. MemoryStore)

nfriedly:
This disables the check unless using the MemoryStore or similar. It should prevent the false positive reported in https://github.com/express-rate-limit/express-rate-limit/issues/453#issuecomment-2148518565

FYI @WesCossick

### DIFF
--- a/source/validations.ts
+++ b/source/validations.ts
@@ -290,9 +290,13 @@ const validations = {
 	},
 
 	/**
-	 * Checks to see if the instance was created inside of a request handler, which would prevent it from working correctly.
+	 * Checks to see if the instance was created inside of a request handler, which would prevent it from working correctly, with the default memory store (or any other store with localKeys.)
 	 */
-	creationStack() {
+	creationStack(store: Store) {
+		if (!store.localKeys) {
+			return // Using an external store. This should be safe, even if not ideal.
+		}
+
 		const { stack } = new Error(
 			'express-rate-limit validation check (set options.validate.creationStack=false to disable)',
 		)
--- a/docs/reference/changelog.mdx
+++ b/docs/reference/changelog.mdx
@@ -9,7 +9,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [7.3.0](https://github.com/express-rate-limit/express-rate-limit/releases/tag/v7.2.1)
+## [7.3.1](https://github.com/express-rate-limit/express-rate-limit/releases/tag/v7.3.1)
+
+### Fixed
+
+- Narrowed the scope of the `creationStack` validation check in order to prevent
+  false-positives
+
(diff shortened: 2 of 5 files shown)


source/validations.ts
```diff
@@ -290,9 +290,13 @@ const validations = {
 	},
 
 	/**
-	 * Checks to see if the instance was created inside of a request handler, which would prevent it from working correctly.
+	 * Checks to see if the instance was created inside of a request handler, which would prevent it from working correctly, with the default memory store (or any other store with localKeys.)
 	 */
-	creationStack() {
+	creationStack(store: Store) {
+		if (!store.localKeys) {
+			return // Using an external store. This should be safe, even if not ideal.
```
gamemaker1, line 297: I don't think we should narrow down this check.

In the example that prompted this PR, the external store used was re-using the same redis client, with the same settings (prefix, server url, etc) for each limiter. That is what allowed it to be safe.

I'm not really sure if what I'm thinking is right, so please correct me if I'm wrong: if someone creates a limiter for each request, but with a new client that possibly stores data in a different server/shard/etc. or uses a different prefix, it should trigger this warning. If the user is entirely sure that the keys will be stored with the same prefix and in the same external server/shard, they can disable the warning. However, if they are not, this warning would serve to alert them to the possiblity that the rate limiter isn't working.

We could add this to the note in the documentation.

gamemaker1 (changes requested):

source/validations.ts
nfriedly: Hum, that's a fair point. My thinking was that this check is mainly to warn less-experienced users of an obvious mistake. 

If you're using an external store a) I'm assuming you have a bit more experience and b) even if you do create a new instance on each request, the data is stored externally, so it's not getting reset for each request. I feel like a different shard/prefix/etc. is pretty unlikely to happen by accident.

I took a quick look through several stores to see if any had any kind of initialization logic that could cause trouble if run repeatedly, and the redis one will re-upload it's scripts each time, which adds a bit of overhead. 

Maybe you're right that we should leave the check as-is and just update the docs to make it clear that you can disable it if you're confident in your use-case, but warn of the potential pitfalls.

WesCossick: Throwing in my two cents, seeing as it was my comment that prompted this PR... I agree with @nfriedly that it seems pretty unlikely that someone would accidentally use different prefixes/shards/etc. with an external store. But, the performance concerns with the initialization overhead seem compelling enough to leave the `ERR_ERL_CREATED_IN_REQUEST_HANDLER` warning in place for all stores. However, I really think the description that goes along with that warning should be updated to be more clear about when it's a true problem and when it's not, including when performance concerns could be a factor. That way, an engineer who encounters this warning can evaluate their situation and decide whether it's safe to disable the warning or whether they need to alter their design.

gamemaker1: Created #462 as per your suggestions, it logs a different error message if the store does not have `localKeys` set to true. Please do review it.

Thanks!

PR #462, opened 2024-06-06
fix: print a different error for `creationStack` if the stores has `localKeys` set to false

gamemaker1:
Supersedes #461, based on our discussion [here](https://github.com/express-rate-limit/express-rate-limit/pull/461#discussion_r1626940562).

### DIFF
--- a/docs/reference/error-codes.mdx
+++ b/docs/reference/error-codes.mdx
@@ -270,7 +270,10 @@ it is called at app initialization, not in response to each request:
 + app.use(rateLimitFactory()); // Works
 ```
 
-Set `validate: {creationStack: false}` in the options to disable the check.
+Set `validate: {creationStack: false}` in the options to disable the check, only
+if you are sure that the rate limiter is working as intended, and is not facing
+performance problems (that may be caused by repeated initialization of an
+external data store).
 
 ## Warnings
 
--- a/source/validations.ts
+++ b/source/validations.ts
@@ -200,8 +200,8 @@ const validations = {
 	},
 
 	/**
-	 * Warns the user that the behaviour for `max: 0` / `limit: 0` is changing in the next
-	 * major release.
+	 * Warns the user that the behaviour for `max: 0` / `limit: 0` is
+	 * changing in the next major release.
 	 *
 	 * @param limit {number} - The maximum number of hits per client.
 	 *
@@ -234,8 +234,8 @@ const validations = {
 	},
 
 	/**
-	 * Warns the user that the `onLimitReached` option is deprecated and will be removed in the next
-	 * major release.
+	 * Warns the user that the `onLimitReached` option is deprecated and
+	 * will be removed in the next major release.
 	 *
 	 * @param onLimitReached {any | undefined} - The maximum number of hits per client.
 	 *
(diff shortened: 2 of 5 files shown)


source/validations.ts
```diff
@@ -290,13 +292,29 @@ const validations = {
...
 		if (stack?.includes('Layer.handle [as handle_request]')) {
+			if (!store.localKeys) {
+				// This means the user is using an external store, which may be safe.
+				// Print out an error anyways, to alert them of the possibility that
+				// the rate limiter may not work as intended.
+
+				// See the discussion here: https://github.com/express-rate-limit/express-rate-limit/pull/461#discussion_r1626940562.
+				throw new ValidationError(
+					'ERR_ERL_CREATED_IN_REQUEST_HANDLER',
+					'Dynamically creating an express-rate-limit instance for every request may result in unexpected behaviour and unnecessary re-initialization of external stores. Make sure the same data store, prefix, etc. are used.',
```
nfriedly, line 313: I think we should re-word this a bit, because at the point we log it, we don't know that it's for every request, we just know it was created for _this_ request. (They could be caching instances after the first request.)

docs/reference/error-codes.mdx
```diff
@@ -270,7 +270,10 @@ it is called at app initialization, not in response to each request:
 + app.use(rateLimitFactory()); // Works
 ```
 
-Set `validate: {creationStack: false}` in the options to disable the check.
+Set `validate: {creationStack: false}` in the options to disable the check, only
+if you are sure that the rate limiter is working as intended, and is not facing
+performance problems (that may be caused by repeated initialization of an
+external data store).
```
nfriedly, line 276: Perhaps this should be it's own paragraph. Something like this:

```suggestion
In certain advanced use cases, such as serving multiple websites from a single app, it may be necessary to dynamically create a rate limiter in response to a request. In this case the validation check may be disabled, but care should be taken to ensure everything works as expected. Consider caching the limiter after it's first usage to avoid repeating the initialization cost and any background work the store may require.

Set `validate: {creationStack: false}` in the options to disable the check.
```

source/validations.ts
nfriedly: Maybe just something like this:

```suggestion
					'express-rate-limit instance should *usually* be created at app initialization, not when responding to a request.',
```

... and then send them to the documentation for more information. (It'll automatically get a "See [docs url] for more information." appended to the end.)

gamemaker1:
Merged your suggestions as-is, thank you!

nfriedly (approved):

gamemaker1:
No problem! I was working on creating a GitHub action to release a new version using the manual dispatch trigger, I'll get the PR up in some time.

nfriedly:
Oh, cool!

gamemaker1:
I pushed it directly to `main` 😅 and released 7.3.1 with it!

You can run it by going to the actions tab and then clicking on the `Release` workflow in the left sidebar, then clicking on the `Run Workflow` button on the right hand side. It takes the version (major/minor/patch/x.y.z) as input, bumps the version in `package*.json` and creates the git tag to trigger the actual release workflow as defined in `ci.yaml`.

On another note, I see that the Mintlify deployments have been failing due to an error in 'verifying update permissions`, and I can't deploy it manually from the website; it tells me to upgrade the current plan. Could you please try deploying it from your account too? Otherwise we could shoot them an email.

nfriedly:
Yeah, I just sent them a message about that the day before yesterday. I
have a suspicion that it only updates from commits that happen on GitHub:
either in the web editor or via merging a PR 🤦‍♂️
Thanks,
-Nathan

nfriedly:
Welp, I just disproved that theory. https://github.com/express-rate-limit/express-rate-limit/commit/730dc484f0dab8dc356026ad04b8a3ec9cf869a3 was done in the github UI, so it's a verified commit. But the Mintlify deployment still failed with the same error.

nfriedly:
Oh, I just realized that I had an email from a month ago about mintlify needing new permissions on github. I just granted those, pushed another change, and.. it still failed. I'm going to send them an email.

PR #471, opened 2024-07-18
feat: Added passOnStoreError to skip rate limitter if the store is not available (#447)

nfriedly:
I don't know why github closed #447 and won't let me re-open it, but I just gave up and recreated the PR here.

This PR:
* replaces #447 
* closes https://github.com/express-rate-limit/rate-limit-redis/pull/193
* fixes https://github.com/express-rate-limit/rate-limit-redis/issues/190
* fixes https://github.com/express-rate-limit/rate-limit-redis/issues/72

Original description below:

---

This PR adds the parameter passOnStoreError to allow requests if the store is not available.
For instance, if using Redis and the Redis server disconnects, the request will still pass.

This feature can be important for HA issues.

### DIFF
--- a/source/lib.ts
+++ b/source/lib.ts
@@ -115,6 +115,7 @@ type Configuration = {
 	requestWasSuccessful: ValueDeterminingMiddleware<boolean>
 	store: Store
 	validations: Validations
+	passOnStoreError: boolean
 }
 
 /**
@@ -241,6 +242,7 @@ const parseOptions = (passedOptions: Partial<Options>): Configuration => {
 				response.send(message)
 			}
 		},
+		passOnStoreError: false,
 		// Allow the default options to be overriden by the options passed to the middleware.
 		...notUndefinedOptions,
 		// `standardHeaders` is resolved into a draft version above, use that.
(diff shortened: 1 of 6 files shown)


nfriedly:
I'm adding docs for this and I'm probably also going to give it an error code.

nfriedly:
Ok @Gp2mv3 & @gamemaker1, I think this is ready for review. The tests are failing, but it seems to be an issue with github/npm, so I'm going to let it go for now and see if it's resolved itself in a day or two.

source/lib.ts
```diff
@@ -330,8 +332,26 @@ const rateLimit = (
...
+			let totalHits = 0
+			let resetTime
+			try {
+				const incrementResult = await config.store.increment(key)
+				totalHits = incrementResult.totalHits
+				resetTime = incrementResult.resetTime
+			} catch (error) {
+				if (config.passOnStoreError) {
+					console.error(
+						'express-rate-limit: error from store, allowing request without rate-limiting.',
```
gamemaker1, line 346: I think we should add a note informing the user what option caused this to happen. Does something like the following sound good?

```
express-rate-limit: store threw error, allowing request without rate limiting since `passOnStoreError` is true.
```

gamemaker1 (approved):
Amazing work, thank you!! Just left one comment regarding the error message printed, let me know what you think about it.

source/lib.ts
nfriedly: I thought about adding an ERR_ERL_* code here, but decided against it since there could be a million different errors that all cause this line to get logged.

gamemaker1: Yes, good call.

nfriedly:
This is the issue that's breaking CI: https://github.com/npm/cli/issues/7657

gamemaker1:
I think we can merge this now @nfriedly, all checks are passing. Are there any changes you want to make before we ship it?

nfriedly:
Ok, should be released as v7.4.0 once CI finishes

modestaspruckus:
Hello we're having issues with async middlewares due this flag, I believe it is missing `return` statement after next() is called though it did not helped. We're also receiving promise errors
```js
(node:28949) PromiseRejectionHandledWarning: Promise rejection was handled asynchronously (rejection id: 4)
```

image

nfriedly:
Yeah, I think you're right, there should be a return there. I'll fix that.

PR #472, opened 2024-07-18
Update error-codes.mdx

Ptitet:
Removed mention to old setting `max` to use `limit`

### DIFF
--- a/docs/reference/error-codes.mdx
+++ b/docs/reference/error-codes.mdx
@@ -286,7 +286,7 @@ Set `validate: {creationStack: false}` in the options to disable the check.
 > Added in `6.10.0`.
 
 In express-rate-limit version 6 and older, the rate limiter would be disabled
-when setting `limit` to `0` in the options. Starting with version
+when setting `limit` (previously `max` until version `7.0.0`) to `0` in the options. Starting with version
 `7.0.0`, this is no longer the case - the rate limiting will apply from the very
 first request instead. See
 [#369](https://github.com/express-rate-limit/express-rate-limit/discussions/369)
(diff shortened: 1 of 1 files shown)


docs/reference/error-codes.mdx
```diff
@@ -286,7 +286,7 @@ Set `validate: {creationStack: false}` in the options to disable the check.
 > Added in `6.10.0`.
 
 In express-rate-limit version 6 and older, the rate limiter would be disabled
-when setting `limit` (now `max`) to `0` in the options. Starting with version
+when setting `limit` to `0` in the options. Starting with version
```
nfriedly, line 289: Oh 🤦 

Those two were supposed to be flipped. It's talking about older versions in that sentence, so it should read "max (now limit)"

```suggestion
when setting `max` (now `limit`) to `0` in the options. Starting with version
```

Ptitet:
I thought about this but I wasn't sure.
Is it really necessary to keep the `max` ? It's deprecated so people should use `limit`, and the help to know that they are the same should be in the doc for `limit` no ?

nfriedly:
Yeah, looking at the [npm download stats](https://www.npmjs.com/package/express-rate-limit?activeTab=versions), the majority of users are still on older versions, so I think we should keep the mention of `max` in the docs for now.

Also the error code has `MAX` in it, because it predates the rename.

Ptitet:
@nfriedly what about "setting `limit` (previously `max`)" ?

nfriedly:
> what about "setting limit (previously max)" ?

Yeah, I like that 

> We can even specify that this was until release 7.0.0.

Also a good idea.

Ptitet:
I’ve made a new commit to reflect what we discussed. Does it look like what you expected?

nfriedly (approved):
Yep, this looks good, thanks!

nfriedly:
Merged in [ff3702e](https://github.com/express-rate-limit/express-rate-limit/commit/ff3702e4d92c2ab93022d8f66e0c3786c3b8d799)

(Sorry, I did it locally to fix the lint error, but github doesn't connect squash commits with the original PR, so it's going to show up as closed rather than merged.)

PR #482, opened 2024-10-03
fix: return after calling next() when passOnStoreError is used

nfriedly:
Fixes at least part of the issue reported at https://github.com/express-rate-limit/rate-limit-redis/issues/72#issuecomment-2389588306 and https://github.com/express-rate-limit/express-rate-limit/pull/471#issuecomment-2391091012

@modestaspruckus I know you said you tried this already and it didn't fix the issue, one thing that comes to mind is that we ship two copies of the code, one for commonjs (`require`) and one for esm (`import`) - it's possible that you edited one but your test used the other(?) Alternatively, it could be something to do with the store.

I think we should ship this fix regardless, because not returning was definitely wrong. Then you can try it out and if you're still experiencing the issue, we can dig in further.

### DIFF
--- a/test/library/middleware-test.ts
+++ b/test/library/middleware-test.ts
@@ -972,6 +972,7 @@ describe('middleware test', () => {
 	})
 
 	it('should pass if the store throws an error and passOnStoreError is true', async () => {
+		jest.spyOn(console, 'error').mockImplementation(() => {})
 		const app = createServer(
 			rateLimit({
 				limit: 1,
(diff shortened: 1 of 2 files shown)


test/library/middleware-test.ts
```diff
@@ -980,5 +981,31 @@ describe('middleware test', () => {
...
 		)
 		await request(app).get('/').expect(200)
+		expect(console.error).toHaveBeenCalledTimes(1)
+		expect(console.error).toHaveBeenCalledWith(
+			expect.stringContaining('allowing'),
+			expect.any(Error),
+		)
+	})
+
+	it('should should only call next once when passOnStoreError causes it to skip limiting', async () => {
```
gamemaker1, line 991: should ~should~

gamemaker1 (approved):
LGTM :)

test/library/middleware-test.ts
nfriedly: 🤦‍♂️

nfriedly:
Ok, v7.4.1 should be published with this fix shortly

gamemaker1:
Great :D

modestaspruckus:
Awesome, it fixed an issue

PR #486, opened 2024-12-06
feat(headers): implement `draft-8` headers

### LINKED ISSUE #484: Add support for emitting headers in IETF RateLimit Headers draft 8 format
### Description

https://datatracker.ietf.org/doc/html/draft-ietf-httpapi-ratelimit-headers-08 was just published about a week ago with some changes that we should add support for

### Why

--

### Alternatives

--

gamemaker1:
Closes #484.


source/headers.ts
```diff
@@ -29,6 +31,40 @@ const getResetSeconds = (
...
+	const minutes = windowMs / (1000 * 60)
+	const hours = windowMs / (1000 * 60 * 60)
+	const days = windowMs / (1000 * 60 * 60 * 24)
+
+	if (seconds < 60) return `${seconds}sec`
+	if (minutes < 60) return `${minutes}min`
+	if (hours < 24) return `${hours}hrs`
+
+	return `${days}day`
+}
```
nfriedly, line 51: Nice!

I like that this will have a decimal for values that don't divide evenly, but omit it otherwise. (E.g. `1000` -> `'1sec'`, and `1500` -> `'1.5sec'`)

source/headers.ts
```diff
@@ -107,6 +143,36 @@ export const setDraft7Headers = (
...
+	key: string,
+): void => {
+	if (response.headersSent) return
+
+	const windowSeconds = Math.ceil(windowMs / 1000)
+	const resetSeconds = getResetSeconds(info.resetTime, windowMs)
+	const duration = getDurationInWords(windowMs)
+	const partitionKey = getPartitionKey(key)
+
+	const name = `rl-${info.limit}-in-${duration}`
```
nfriedly, line 168: I like this as a default, but I think we should provide an option to customize the name.

source/headers.ts
```diff
@@ -107,6 +143,36 @@ export const setDraft7Headers = (
...
+	const policy = `q=${info.limit}; w=${windowSeconds}; pk=:${partitionKey}:`
+	const header = `r=${info.remaining}; t=${resetSeconds!}`
+
+	response.setHeader('RateLimit-Policy', `"${name}"; ${policy}`)
+	response.setHeader('RateLimit', `"${name}"; ${header}`)
```
nfriedly, line 173: I think this will overwrite the existing one if two instances set headers - we should instead append in that case, with an array of values.

source/lib.ts
```diff
@@ -373,11 +374,23 @@ const rateLimit = (
...
+
+					case 'draft-8': {
+						config.validations.headersResetTime(info.resetTime)
+						setDraft8Headers(response, info, config.windowMs, key)
+						break
+					}
+
+					default: {
+						setDraft6Headers(response, info, config.windowMs)
+						break
```
nfriedly, line 392: I'm not sure I like defaulting to `draft-6`. Maybe add a `draft-6` case and log or throw an error in the default case? (Could be a validation check so it only runs on the first request.)

nfriedly (changes requested):
Nice work, I really like it overall!

The main reason I'm hitting "Request changes" is for the `setHeader()` clobbering any previous header. I think we should play nice with other instances now that it's more feasible, and perhaps add a note to our docs somewhere about that.

I may also ping one or two of the folks that worked on the draft and ask for their opinion on this implementation.

nfriedly:
@ioggstream & @darrelmiller do you have any feedback on this implementation of draft 8 of RateLimit header fields for HTTP?

source/lib.ts
gamemaker1: Done!

source/headers.ts
gamemaker1: How would we accept the custom name? We could probably resurrect the `header` option we had deprecated, and make it accept a an object like so:

```js
const options = {
	// ...
	headers: {
		legacy: false,
		standard: {
			version: 'draft-8',
			name: 'burst',
			partition: 'storage'
		}
	}
}
```

This would also let us allow the user to generate their own `partitionKey`.

Or, we could add a new parameter called `name` that acts as the more generic name of the rate limiter, and the partition key remains auto-generated by us.

I'm in favour of this option for now since it is much simpler and does not involve a possibly breaking change, but the first option might be better in the long term if the drafts keep changing and adding more parameters.

source/headers.ts
nfriedly: `res.append()` is a good solution, but we'll need to bump the express peer dependency from 4 to 4.11 in `package.json` - http://expressjs.com/en/4x/api.html#res.append

(Honestly we should probably double check see if there are any other apis we're using that require versions of Express newer than 4.0.0)

source/headers.ts
nfriedly: I was thinking of a new option like `identifier`, or `headersIdentifier`, or similar. `name` works too. It would just be ignored unless draft8+ is used. I don't think we should resurrect the `headers` option until the standard is completed.

gamemaker1: Okay, will use `identifier` then.

source/headers.ts
gamemaker1: Done

source/headers.ts
```diff
@@ -29,6 +34,40 @@ const getResetSeconds = (
...
+	const seconds = windowMs / 1000
+	const minutes = windowMs / (1000 * 60)
+	const hours = windowMs / (1000 * 60 * 60)
+	const days = windowMs / (1000 * 60 * 60 * 24)
+
+	if (seconds < 60) return `${seconds}sec`
+	if (minutes < 60) return `${minutes}min`
+	if (hours < 24) return `${hours}hrs`
+
+	return `${days}day`
```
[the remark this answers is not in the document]
gamemaker1: Made it fancy, so the 's' is only added if the hours or days are more than 1.

source/headers.ts
```diff
@@ -107,6 +146,38 @@ export const setDraft7Headers = (
...
+	key: string,
+): void => {
+	if (response.headersSent) return
+
+	const windowSeconds = Math.ceil(windowMs / 1000)
+	const resetSeconds = getResetSeconds(info.resetTime, windowMs)
+	const duration = getDurationInWords(windowMs)
+	const partitionKey = getPartitionKey(key)
+
+	const name = id === 'rl' ? `rl-${info.limit}-in-${duration}` : id
```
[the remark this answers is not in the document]
gamemaker1: Makes sense.

source/validations.ts
```diff
@@ -250,6 +251,26 @@ const validations = {
...
+	 * @returns {void}
+	 */
+	headersDraftVersion(version?: any) {
+		if (
+			typeof version !== 'string' ||
+			!SUPPORTED_DRAFT_VERSIONS.includes(version)
+		) {
+			const versionString = SUPPORTED_DRAFT_VERSIONS.join(', ')
+			throw new ValidationError(
+				'ERR_ERL_HEADERS_UNSUPPORTED_DRAFT_VERSION',
```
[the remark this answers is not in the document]
gamemaker1: Done

source/lib.ts
```diff
@@ -189,11 +191,15 @@ const parseOptions = (passedOptions: Partial<Options>): Configuration => {
...
-	// `true`, it resolve to `draft-6`. `draft-7` (recommended) is used only if
-	// explicitly set.
+	// `true`, it resolve to `draft-6`. `draft-7` and draft-8` (recommended) are
+	// used only if explicitly set.
 	let standardHeaders = notUndefinedOptions.standardHeaders ?? false
 	if (standardHeaders === true) standardHeaders = 'draft-6'
 
+	// The actual default is generated by `setDraft8Headers()`, so if it is
+	// undefined, set it to `rl` for now.
+	const identifier = notUndefinedOptions.identifier ?? 'rl'
```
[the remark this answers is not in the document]
gamemaker1: Made it so that `identifier` accepts a function, which can access rate limit info about the current request from the `request[property]`.

nfriedly (approved):
I like it!

I'm going to be away next week, possibly without good internet. Do you want to go ahead and ship this?

gamemaker1:
Yes, sure!

PR #495, opened 2025-02-27
Next.js usage note in docs

nfriedly:
This calls it out as not officially supported, but provides enough info to get someone started if they wanted to do it anyways.

Fixes https://github.com/express-rate-limit/rate-limit-postgresql/issues/36

@gamemaker1 what do you think of my [suggestion of an official next.js support package](https://github.com/express-rate-limit/rate-limit-postgresql/issues/36#issuecomment-2688541101)?

### DIFF
--- a/docs/quickstart/usage.mdx
+++ b/docs/quickstart/usage.mdx
@@ -31,7 +31,7 @@ const { rateLimit } = require('express-rate-limit')
 
 </CodeGroup>
 
-### Using the Library
+### Using the Library in Express
 
 The `rateLimit` function accepts an options object and returns the rate limiting
 middleware.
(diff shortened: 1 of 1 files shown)


gamemaker1:
Hi, sorry for the late reply.

I think we could work on an official support package for NextJS, provided it is just a wrapper around the current rate limiter, and does not require us to duplicate any of the logic (I am a bit concerned about the header-setting or response-sending code requiring duplication).

Let me see if I can get something working in the next few days. The name `@express-rate-limit/nextjs` sounds good, I'll make a separate repository for it.

gamemaker1 (approved):

PR #496, opened 2025-03-19
Add unit test for resetKey method in MemoryStore

gamzeozgul:
Added a unit test for resetKey in MemoryStore.

### DIFF
--- a/test/library/memory-store-test.ts
+++ b/test/library/memory-store-test.ts
@@ -64,6 +64,24 @@ describe.only('memory store test', () => {
 		expect(totalHits).toEqual(2)
 	})
 
+	test('resetKey should remove the key from storage', async () => {
+	const store = new MemoryStore()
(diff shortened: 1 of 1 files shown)


nfriedly:
Hey, thanks for this! 

We actually do have a test for `resetKey` already:

https://github.com/express-rate-limit/express-rate-limit/blob/3c94ac2b97466e11d0ac6c3c8bf7def523ce0fe3/test/library/memory-store-test.ts#L67-L77

But I suppose yours is a little different in that it calls `get` rather than `increment`, so maybe it makes sense to have both. Would you mind moving your new test to come right after the existing one?

nfriedly (approved):
LGTM, thanks!

nfriedly:
Op, it doesn't like the indentation. `npm run format` can fix that automatically.

PR #503, opened 2025-06-14
Default to /56 subnet block for IPv6

nfriedly:
This enables blocking of entire subnets for IPv6 with a somewhat aggressive default block of /56.

It also adds a helper function that can be used as a fallback in custom keyGenerators, and a couple of new validation checks.

Todo:
* [ ] Readme example
* [ ] Readme config option
* [ ] Docs for new config option
* [ ] Docs for new error codes
* [ ] Redirects for new error codes (different repo)

I also added the key to the `key` to the `RateLimitInfo` object since it's no longer quite as obvious.

### DIFF
--- a/source/ip-key-generator.ts
+++ b/source/ip-key-generator.ts
@@ -1,0 +1,25 @@
+import { isIPv6 } from 'node:net'
+import iptools from 'ip'
+
+/**
+ * Returns the IP address itself for IPv4, or a CIDR-notation subnet for IPv6 (e.g. '1234:abcd::/64')
+ *
+ * If you write a custom keyGenerator that allows a fallback to IP address for unauthenticated users, return ipKeyGenerator(req.ip) rather than just req.ip.
+ *
+ * See [Options.ipv6Subnet] for more info.
+ *
+ * @param ip request.ip
+ * @param [ipv6Subnet=64] subnet mask for IPv6 addresses
+ */
+export function ipKeyGenerator(ip: string, ipv6Subnet: number | false = 64) {
+	if (ipv6Subnet && isIPv6(ip)) {
+		// For IPv6, return the network address of the subnet in CIDR format
+		return `${iptools.mask(
+			ip,
+			iptools.fromPrefixLen(ipv6Subnet),
+		)}/${ipv6Subnet}`
+	}
+
+	// For IPv4, just return the IP address itself
+	return ip
+}
--- a/source/types.ts
+++ b/source/types.ts
@@ -314,6 +314,41 @@ export type Options = {
 	keyGenerator: ValueDeterminingMiddleware<string>
 
 	/**
+	 * IPv6 subnet mask applied to IPv6 addresses in the default keyGenerator.
+	 *
+	 * Generally, ISPs that support IPv6 give each of their customers a range of IPv6 addresses, whereas they usually provide only a single IPv4 address per customer.
+	 * A malicious user could iterate through their range of IPv6 addresses and bypass simple IP-based ratel imiting.
+	 * This setting counteracts that by allowing the rate-limiter to block an entire range of IPv6 addresses at once.
+	 *
+	 * The default value is 56, corresponding to a /56 subnet.
+	 *
+	 * The valid range is 1-128, but typically ISPs assign subnets in the range of 32-64, with 64 being the most common. The validation check will warn for values outside the 32-64 range, (but can be disabled).
+	 *
+	 * In a 128 bit IPv6 address, the number of bits in this setting will be kept, and the rest will be ignored when determining the user's IP address.
+	 * A smaller number means the ISP is reserving fewer bits for their control and allowing a larger portion of the address to be in the customer's control, thereby allowing the customer to have a greater number of unique IPs.
+	 *
+	 * 56 is a moderately aggressive default. It may be increased to if users are being incorrectly blocked (try 60 or 64), or decreased if you are seeing evidence of abuse.
+	 * 64, 60 ([Comcast][1]), 56, and 48 are all common values used by various ISPs.
+	 *
+	 * The option may also be set to a function that returns the value if you want to apply different subnets to different different users.
+	 *
+	 * Set to false to disable and always use the IP without masking.
+	 *
+	 * [1]: https://news.ycombinator.com/item?id=44228908
+	 */
+	ipv6Subnet:
+		| 64 // A few common values, followed by number as a catch-all
+		| 60 // Apparently comcast allows customers to request up to a /60, which is effectively 16 /64s
+		| 56
+		| 52
+		| 50
+		| 48
+		| 32
+		| number // Todo: figure out how to do a "range type" to replace `number` with "1-128". (The validator limits to 32-64, but typescript should probably allow the whole range.)
+		| ValueDeterminingMiddleware<number>
+		| false
+
+	/**
 	 * Express request handler that sends back a response when a client is
 	 * rate-limited.
 	 *
@@ -330,7 +365,7 @@ export type Options = {
 	skip: ValueDeterminingMiddleware<boolean>
 
 	/**
-	 * Method to determine whether or not the request counts as 'succesful'. Used
+	 * Method to determine whether or not the request counts as 'successful'. Used
 	 * when either `skipSuccessfulRequests` or `skipFailedRequests` is set to true.
 	 *
 	 * By default, requests with a response status code less than 400 are considered
@@ -393,6 +428,7 @@ export type RateLimitInfo = {
 	used: number
 	remaining: number
 	resetTime: Date | undefined
+	key: string // IP address, etc.
 
 	/**
 	 * NOTE: The `current` field is deprecated and renamed to `used`. The library
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,11 +8,15 @@
 			"name": "express-rate-limit",
 			"version": "7.5.0",
 			"license": "MIT",
+			"dependencies": {
+				"ip": "2.0.1"
+			},
 			"devDependencies": {
 				"@express-rate-limit/prettier": "1.1.1",
 				"@express-rate-limit/tsconfig": "1.0.2",
 				"@jest/globals": "29.7.0",
 				"@types/express": "4.17.20",
+				"@types/ip": "1.1.3",
 				"@types/jest": "29.5.6",
 				"@types/node": "20.8.7",
 				"@types/supertest": "2.0.15",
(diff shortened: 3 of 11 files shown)


nfriedly:
Ugh, I went through three different IP libraries looking for one that played nice with typescript and the one I landed on apparently has issues with ESM....

nfriedly:
Oh, the issue might be that we're bundling the dependency into our dist mjs file, which means it's mixing ESM and CJS in the same file. I bet there's a way to tell the bundler to not do that...

nfriedly:
Okay, @gamemaker1 I think this has everything I want for this repo. /56 default, helper for custom key generators, validation checks, tests, changelog, and lots of documentation.

Please review when you have the time. Sorry it's such a huge PR for what seemed initially like it was going to be a very small fix.

source/types.ts
```diff
@@ -313,6 +313,41 @@ export type Options = {
...
+	 */
+	ipv6Subnet:
+		| 64 // A few common values, followed by number as a catch-all
+		| 60 // Apparently comcast allows customers to request up to a /60, which is effectively 16 /64s
+		| 56
+		| 52
+		| 50
+		| 48
+		| 32
+		| number // Todo: figure out how to do a "range type" to replace `number` with "1-128". (The validator limits to 32-64, but typescript should probably allow the whole range.)
```
gamemaker1, line 347: I think we can use something like this: https://stackoverflow.com/a/39495173

gamemaker1 (changes requested):
The documentation is really thorough, I love it!

Could we remove the `ipv6Subnet` field from the rate limiter's options, and instead add it to the options for the helper function? Kind of how custom stores accept their own options, these custom key generators could as well. This would help if we decide to add more features in the future, like stripping port numbers from IPv4, etc. I had suggested this in my comment [here](https://github.com/express-rate-limit/express-rate-limit/issues/502#issuecomment-2969378333), please let me know your thoughts on the same.

docs/reference/configuration.mdx
```diff
@@ -292,6 +299,74 @@ rate limiter. This could be combined with a second instance of
 
 </Note>
 
+## `ipv6Subnet`
```
nfriedly, line 302: We'll need to rework this, maybe the configuration page gets level-2 headings for `rateLimit` and `ipKeyGenerator`, and then all of the options themselves become level-3 headers?

Alternatively we split them into two pages, but I don't really like that.

source/ip-key-generator.ts
```diff
@@ -1,25 +1,124 @@
...
+	// validator warns outside of 32-64, but typescript allows the whole range.
+	| 64
+		| 60
+		| 56
+		| 52
+		| 50
+		| 48
+		| 32
+		| number // TODO: figure out how to do a "range type" to replace `number` with "1-128".
+		| ValueDeterminingMiddleware<number> // This allows setting a custom value per-request.
```
nfriedly, line 58: I feel like the `ValueDeterminingMiddleware` option makes this extra complex - you call the `ratelimit` function, passing it options including a `keyGenerator` that is the return of the `ipKeygeneratorFunction`, which is passed options including a `ipv6Subnet` function. I already felt like the blocklist example to use this as a function was a bit complex, and it was as simple as I could make it. Here's what it looks like with this version:

```
app.use(rateLimit({
  // ...
  keyGenerator: ipKeyGenerator({
    ipv6Subnet: (req) => 


```js
import { BlockList } from 'node:net'
// or
// const { BlockList } = require('node:net')

const networks_60 = new BlockList()
// comcast defaults to /64 but allows customers to request /60
// non-exhaustive list, there are many more
networks_60.addSubnet('2a0c:93c0:10::', 44, 'ipv6')
networks_60.addSubnet('2a0c:93c0:6000::', 48, 'ipv6')
networks_60.addSubnet('2a0c:93c0:6002::', 48, 'ipv6')

// AWS allows customers to request up to a /44!
// full list at https://ip-ranges.amazonaws.com/ip-ranges.json
const networks_44 = new BlockList()
networks_44.addSubnet('2600:1f69:7400::', 40, 'ipv6')

app.use(
	rateLimit({
    ipKeyGenerator({
  		ipv6Subnet: function (req, res) {
	  		if (networks_60.check(req.ip, 'ipv6')) {
	  			return 60
  			}
	  		if (networks_44.check(req.ip, 'ipv6')) {
	  			return 44
	  		}
	  		// fallback
	  		return 56
	  	}, // end of ipv6Subnet
    }), // end of ipKeyGenerator()
		// ...
	}),
)
```

(I also noticed that it made the tests a bit more complex, but I'm not too concerned about that.)

nfriedly (commented):
Hum, I'm kind of divided on this. I appreciate the work you put into this, and it does make a bit of sense to put the keyGenerator config into the keyGenerator, but I think I like the previous version better. 

I left a lot of comments below, and I think we could address most of them, but there are a couple that I don't see good solutions for: 

My biggest concern is that it makes the `keyGeneratorIpFallback` validation much less useful, because we no longer know the name of the function we're looking for it to use.

The second thing is that I really liked the simplicity of having a flat config where everything was at the same level. We do it that way for the default MemoryStore, and I don't mind having an extra validation check for incompatible top-level options. I've always been a bit annoyed that we needed to have separate configs for the rate limiter and external stores, but I think it's unavoidable. And that it makes a bit more sense because they are separate libraries.

I don't know. I hope I'm not being mean or egotistical here. I do think we could make your version work, with the possible exception of the one validation check. I just like my version better. What do you think?

nfriedly:
> [...] I'll go ahead and revert the changes, and then we can ship it if you think it's ready.

Thanks, I appreciate that. Let me give it another look over, I think I want to mention somewhere in the docs that there isn't really a good way to know what subnet a user has - we can guess based on their ISP, but many ISPs offer a range of subnet sizes, so it's not a sure thing even if you look up the user's ISP.

> > Also, I'm second-guessing myself about making this a patch release. A minor version feels more correct, and can still be backported. I think a CVE will get dependabot and others to upgrade a lot of people regardless of the version.
> 
> I have been working on #507, and it contains a few major changes:
> 
>     * upgrading all dev dependencies
>     * making the library work with express v5
>     * dropping support for node 16 and 18 (both out of maintenance now as of april this year)
>     * running the tests on node 20, 22, 24
>     * upgrading all the external tests and running them on node 24 with express v4 and v5.

Oh, that's awesome! I just noticed that express v5 had finally been released, but I hadn't had a chance to look into what changed.

> Should we ship this PR along with #507, so we can do a major version bump (since we're dropping support for node versions)? And then backport only this PR as a minor version bump for v7 and v6?

I'd lean towards shipping this first and then doing the major version bump after, but mostly because it seems like less work. I don't feel too strongly about it, though.

gamemaker1:
> Thanks, I appreciate that. Let me give it another look over, I think I want to mention somewhere in the docs that there isn't really a good way to know what subnet a user has - we can guess based on their ISP, but many ISPs offer a range of subnet sizes, so it's not a sure thing even if you look up the user's ISP.

Okay, sure!

> I'd lean towards shipping this first and then doing the major version bump after, but mostly because it seems like less work. I don't feel too strongly about it, though.

Since #507 is no longer breaking we can ship them together.

nfriedly:
OK, I think I've got this where I want it. Sorry it took me a while. Do you want to review the changes?

gamemaker1 (approved):

gamemaker1:
Re-worded the message a bit for the new validation, and added some docs for it. I also updated the [`express-rate-limit.github.io`](https://express-rate-limit.github.io) site with the new links.

I think this is good to go, merge it when you're ready! We can cherry pick some of the commits to the 6.11.2 tag and publish that as 6.12 once this is merged.

nfriedly:
Oh, shoot, I didn't push my last change with the docs I wrote up for the new validation and some other changes 🤦 

I'll merge what you did with mine, I had some example code in the docs for the validation that I want to keep.

nfriedly:
Ok, I've merged our changes together, then I did a little local testing and realized that I wasn't able to trigger the ipv6 validation error, even with obviously incorrect values, because the check on ran if the first request happened to be IPv6. I tweaked the logic to run the validation at instance-creation for non-function values. 

If the setting is set to a function, then it still has the limitation of only being run if the first request happens to be IPv6. I think we should address this (perhaps by disabling validation checks after they're run once, rather than after the first request?) but I think it's outside the scope of this PR.

docs/reference/changelog.mdx
```diff
@@ -9,6 +9,32 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [7.6.0](https://github.com/express-rate-limit/express-rate-limit/releases/tag/v7.6.0)
+
+Note: a semver-patch version increment was chosen for this release when it could
+have been considered a minor or arguably even a major version change because we
```
nfriedly, line 15: Op, I need to rewrite this

docs/reference/changelog.mdx
```diff
@@ -9,6 +9,32 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
+### Added
+
+- `ipv6Subnet` configuration option used by the default `keyGenerator`, defaults
+  to `56`
+- `ipKeyGenerator(ip, ipv6Subnet)` helper method to apply the desired subnet to
+  IPv6 addresses (returns IPv4 unchanged)
+- `ipv6Subnet` validation check on above configuration option's value (allowed
+  range is 32-64)
+- `keyGeneratorIpFallback` validation check on custom `keyGenerator`s to ensure
+  they're using `ipKeyGenerator` if they reference `req.ip` or `request.ip`
```
nfriedly, line 36: One more validation check to mention here

docs/reference/configuration.mdx
```diff
@@ -297,6 +304,78 @@ rate limiter. This could be combined with a second instance of
...
+
+The valid range is 1-128, but typically ISPs assign subnets in the range of
+32-64, with 64 being the most common. The validation check (which can be
+disabled) will warn for values outside the 32-64 range.
+
+In a 128 bit IPv6 address, the number of bits in this setting will be kept, and
+the rest will be ignored when determining the user's IP address. A smaller
+number means the ISP is reserving fewer bits for their control and allowing a
+larger portion of the address to be in the customer's control, thereby allowing
+the customer to have a greater number of unique IPs.
```
nfriedly, line 330: I feel like this paragraph in particular is kind of confusing. I might try rewording this section a bit.

nfriedly:
OK, this time I think I'm done for real - and I actually pushed the changes!

PR #506, opened 2025-06-21
fix: type of DraftHeadersVersion

### LINKED ISSUE #505: Incorrect type for DraftHeadersVersion
### Description

The type for `DraftHeadersVersion` is `string` rather than `"draft-6" | "draft-7" | "draft-8"`, this is because `SUPPORTED_DRAFT_VERSIONS` is defined as a string array rather than a const array

### Library version

7.5.0

### Node version

v22.14.0

5.2.2

### Module system

ESM

JasonMan34:
- closes #505

### DIFF
--- a/test/library/options-test.ts
+++ b/test/library/options-test.ts
@@ -95,4 +95,11 @@ describe('options test', () => {
 			})
 		}).toThrowError(/store/)
 	})
+
+	it('should not allow an invalid standardHeaders option', async () => {
+		rateLimit({
+			// @ts-expect-error Check correct typing of standardHeaders
+			standardHeaders: 'not-a-valid-draft',
+		})
+	})
 })
--- a/source/headers.ts
+++ b/source/headers.ts
@@ -7,7 +7,11 @@ import type { Response } from 'express'
 import type { RateLimitInfo } from './types.js'
 
 // eslint-disable-next-line @typescript-eslint/naming-convention
-export const SUPPORTED_DRAFT_VERSIONS = ['draft-6', 'draft-7', 'draft-8']
+export const SUPPORTED_DRAFT_VERSIONS = [
+	'draft-6',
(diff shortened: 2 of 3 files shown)


nfriedly:
Thanks, appreciate the fix!

nfriedly:
Does this change usage of the library at all, or only the internal testing / tslint stuff?

(In other words, do you think it merits a new release?)

test/library/options-test.ts
```diff
@@ -95,4 +95,11 @@ describe('options test', () => {
 			})
 		}).toThrowError(/store/)
 	})
+
+	it('should not allow an invalid standardHeaders option', async () => {
+		rateLimit({
+			// @ts-expect-error Check correct typing of standardHeaders
+			standardHeaders: 'not-a-valid-draft',
+		})
+	})
```
gamemaker1, line 104: This should be surrounded by an `expect(...).toThrow()` block.

gamemaker1 (commented):
Hi,

I noticed that the added test is incomplete, but since it's already merged I'll make the change separately.

Thanks for the PR!

test/library/options-test.ts
JasonMan34: This is only a TypeScript test. Although to make it truly TypeScript-only it would need to be surrounded by an `if (false)` block or something.

The code doesn't currently check the value given on instantiation, only when the time actually comes to set the headers, so that code block doesn't currently throw.
I think it should throw actually, as there's no reason to wait if we know the value is wrong, but I didn't want to dive into the flow of the actual code, so I only made the type test

gamemaker1: Ah okay I misunderstood it. We can keep it as-is, I'll just add a comment explaining that its a Typescript-only test then. Thanks!

PR #507, opened 2025-06-25
chore: upgrade deps (incl express to v5) and drop node 16 tests

gamemaker1:
This PR started out with me just upgrading all dependencies to their latest versions, and that cascaded into a large number of fixes needed to get everything working again:

1. Upgraded all development dependencies to their latest versions (jest, supertest, etc.)
2. Swapped out the xo + prettier for [biome](https://github.com/biomejs/biome), which is a combined linter + formatter. It is much easier to configure and use. This change was necessary because of my inability to make our existing configuration work with the latest versions of xo, which uses eslint under the hood - both packages had a lot of breaking changes released recently.
3. Express released 5.0.0 a while back, and the latest version is 5.1.0. This PR upgrades our dev dependency on express to 5.1.0, and keeps the peer dependency on express the same (`>= 4.11`). The only change I had to make to support Express v5 is to the `creationStack` validation.
4. Node 16 and 18 have [gone out of maintenance](https://github.com/nodejs/Release#release-schedule), so this PR drops these two versions from the testing matrix and adds Node 20 (maintenance) and 22 (LTS) in their place.
5. Upgraded dependencies in all external test folders to their latest. I had to comment out the test for the `precise-memory-rate-limit` store since the current jest configuration doesn't seem to be able to load libraries written as `.ts` files that are not compiled to `.js`.

The fourth point constitutes a breaking change, so this would have to be released with a major version bump to v8.0.0. Since we are doing a major version bump, we should also remove things that have been deprecated in v6 (and v7?) before releasing.

Could you also please take a look at the fifth point about the external stores test and see if you can fix the jest issue?

---

P.S.: We should probably do this 'upgrade-to-latest' once a month to avoid these huge PRs.

### DIFF
--- a/source/rate-limit.ts
+++ b/source/rate-limit.ts
@@ -1,6 +1,7 @@
 // /source/lib.ts
 // The option parser and rate limiting middleware
 
+import { isIPv6 } from 'node:net'
 import type { NextFunction, Request, RequestHandler, Response } from 'express'
 import {
 	setDraft6Headers,
@@ -9,6 +10,7 @@ import {
 	setLegacyHeaders,
 	setRetryAfterHeader,
 } from './headers.js'
+import { ipKeyGenerator } from './ip-key-generator.js'
 import { MemoryStore } from './memory-store.js'
 import type {
 	AugmentedRequest,
@@ -226,9 +228,24 @@ const parseOptions = (passedOptions: Partial<Options>): Configuration => {
 			validations.trustProxy(request)
 			validations.xForwardedForHeader(request)
 
-			// By default, use the IP address to rate limit users.
+			// Note: eslint thinks the ! is unnecessary but dts-bundle-generator disagrees
 			// biome-ignore lint/style/noNonNullAssertion: validations.ip is called above
-			return request.ip!
+			const ip: string = request.ip!
+			let subnet: number | false = 56
+
+			if (isIPv6(ip)) {
+				// Apply subnet to ignore the bits that he end-user controls and rate-limit on only the bits their ISP controls
+				subnet =
+					typeof config.ipv6Subnet === 'function'
+						? await config.ipv6Subnet(request, response)
+						: config.ipv6Subnet
+
+				// If it was a function, check the output now (otherwise it got checked earlier)
+				if (typeof config.ipv6Subnet === 'function')
+					validations.ipv6Subnet(subnet)
+			}
+
+			return ipKeyGenerator(ip, subnet)
 		},
 		ipv6Subnet: 56,
 		async handler(


nfriedly:
Thanks, this is awesome! I'll take a more thorough look once I have some time, but I have a few thoughts based on your description:

> 4\. Node 16 and 18 have [gone out of maintenance](https://github.com/nodejs/Release#release-schedule), so this PR drops these two versions from the testing matrix and adds Node 20 (maintenance) and 22 (LTS) in their place.
> 
> The fourth point constitutes a breaking change, so this would have to be released with a major version bump to v8.0.0. 

Hum, I'm not sure I agree with you there. I think if we were making changes that broke compatibility with older node.js versions, that would probably constitute a major release, but just because we aren't testing against it any more doesn't actually mean it's a breaking change.

What about this as a compromise: instead of testing against node 20, 22, & 24 we instead test against node 16, 22, & 24? (And we just kind of assume that if 16 and 22 both work, then everything in between does also.) I think there's even a way to tell github to test against "whatever the current LTS and Stable versions of node.js are", so we just do 16 & that and then we automatically get the new versions after every node.js release.

> Could you also please take a look at the fifth point about the external stores test and see if you can fix the jest issue?

Yeah, I'll take a look at that sometime, but it might take me a bit to get to it.

> P.S.: We should probably do this 'upgrade-to-latest' once a month to avoid these huge PRs.

Yeah, not a bad idea. There used to be a tool called greenkeeper that would send you a PR any time one of your dependencies released a new version, not just when there's a security issue. I think they've shut down now, but maybe there's a similar replacement.

gamemaker1:
Ah the latest jest version does not support the old node versions anymore I guess..

nfriedly:
> Ah the latest jest version does not support the old node versions anymore I guess..

Bummer. It looks like it works on node 18, want to use that instead of 16?

I think there's room for both "we support all current versions of node.js" and the gray area of "we don't intentionally break compatibility with older versions, but also don't test against them". (Also, I don't really see anything breaking getting introduced unless we change the target in TSConfig, so we can deal with that when the need arises.)

gamemaker1:
Done. Also added configuration for dependabot to make PRs for updates weekly.

source/headers.ts
```diff
@@ -24,7 +23,7 @@ const getResetSeconds = (
 	resetTime?: Date,
 	windowMs?: number,
 ): number | undefined => {
-	let resetSeconds: number | undefined = undefined // eslint-disable-line no-undef-init
+	let resetSeconds: number | undefined
```
nfriedly, line 26: We could probably make windowMs a required param, and then we wouldn't have to worry about undefined. The one place where we call `getResetSeconds()` without windowMs, we have windowMs available, so it should be a pretty easy change.

On the other hand, there's already a lot changing here, so maybe best to leave it for now...

source/index.ts
```diff
@@ -1,13 +1,13 @@
...
-// Export all the types as named exports
-export * from './types.js'
+// Export the memory store in case someone wants to use or extend it
+// (see https://github.com/nfriedly/express-rate-limit/issues/289)
+export { MemoryStore } from './memory-store.js'
 
 // Export the rateLimit function as a default export and as a named export, if
 // the default export does not work (see https://github.com/nfriedly/express-rate-limit/issues/280)
-export { default, default as rateLimit } from './lib.js'
+export { default, default as rateLimit } from './rate-limit.js'
```
nfriedly, line 10: Hah, I was just thinking that we should rename this. Good call.

source/validations.ts
```diff
@@ -323,7 +322,10 @@ const validations = {
 			'express-rate-limit validation check (set options.validate.creationStack=false to disable)',
 		)
 
-		if (stack?.includes('Layer.handle [as handle_request]')) {
+		if (
+			stack?.includes('Layer.handle [as handle_request]') || // express v4
+			stack?.includes('Layer.handleRequest') // express v5
```
nfriedly, line 327: Nice

test/external/imports/default-import/js-cjs/package.json
```diff
@@ -1,41 +1,21 @@
...
-    "testTimeout": 30000,
-    "testMatch": [
-      "**/test/**/*-test.js"
-    ],
-    "moduleFileExtensions": [
-      "js",
-      "json"
-    ]
-  }
+	"name": "express-rate-limit-example-default-import-js-cjs",
```
nfriedly, line 2: don't we have indent set to spaces?

test/external/run-all-tests
```diff
@@ -116,31 +118,36 @@ function run_store_tests {
...
 
 	# Get a fresh install of all the node modules
+	echo
 	pnpm install
+	echo
 
-	print_info "running store tests (express 4)"
-	# Run the tests with Express v4
+	print_info "running store tests (express 5)"
+	# Run the tests with Express v5
```
nfriedly, line 126: 🎉

test/external/stores/test/stores-test.ts
```diff
@@ -2,17 +2,19 @@
...
-import redisStore from '../source/redis-store.js'
-import mongoStore from '../source/mongo-store.js'
 import memcachedStore from '../source/memcached-store.js'
-import preciseStore from '../source/precise-store.js'
+import mongoStore from '../source/mongo-store.js'
+import redisStore from '../source/redis-store.js'
+
+// TODO: This package doesn't work with jest, since its entrypoint is
+// typescript, and jest only seems to support loading javascript modules.
+// import preciseStore from '../source/precise-store.js'
```
nfriedly, line 11: Didn't this work previously?

test/library/memory-store-test.ts
```diff
@@ -1,13 +1,20 @@
...
+	it,
+	jest,
+} from '@jest/globals'
 import type { Options } from '../../source/index.js'
+import { MemoryStore } from '../../source/memory-store.js'
 
 const minute = 60 * 1000
 
-describe.only('memory store test', () => {
+describe('memory store test', () => {
```
nfriedly, line 17: 😬

nfriedly (commented):
Looks pretty good overall, just a couple of tweaks I'd like to see.

I wasn't aware of Biome, but happy to give it a shot for the speedup.

package.json
```diff
@@ -52,7 +52,7 @@
 		"tsconfig.json"
 	],
 	"engines": {
-		"node": ">= 16"
+		"node": ">= 20"
```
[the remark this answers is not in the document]
gamemaker1: Done in 230810f.

test/external/imports/default-import/js-cjs/package.json
gamemaker1: The `.editorconfig` file has only `*.yaml` files set to space so I went according to that. I could revert it to spaces if needed.

source/headers.ts
gamemaker1: Yea, I'll fix this in the tests PR instead.

gamemaker1: Done in [6b33fe1](https://github.com/express-rate-limit/express-rate-limit/pull/509/commits/6b33fe183801f06907a1e1342f650da98c42858f).

gamemaker1:
Rebased the PR, @nfriedly please merge it if you think it's ready.

test/external/stores/test/stores-test.ts
nfriedly: Aah. Bummer.

gamemaker1:
> If I remember right, npm always outputs spaces, so it's just easier to leave it that way rather than having thrashing between npm and the liner. Although I don't know what pnpm does...

Looks like this was fixed a while back: https://github.com/npm/npm/issues/5694#issuecomment-307142476. It seems all the package managers respect the existing indentation.

source/rate-limit.ts
```diff
@@ -228,24 +226,9 @@ const parseOptions = (passedOptions: Partial<Options>): Configuration => {
...
-
-				// If it was a function, check the output now (otherwise it got checked earlier)
-				if (typeof config.ipv6Subnet === 'function')
-					validations.ipv6Subnet(subnet)
-			}
-
-			return ipKeyGenerator(ip, subnet)
+			// By default, use the IP address to rate limit users.
+			// biome-ignore lint/style/noNonNullAssertion: validations.ip is called above
+			return request.ip!
```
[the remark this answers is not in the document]
gamemaker1: Oops looks liked I picked the wrong hunk while fixing the conflict. Thanks for catching it!

nfriedly:
> > If I remember right, npm always outputs spaces, so it's just easier to leave it that way rather than having thrashing between npm and the liner. Although I don't know what pnpm does...
> 
> Looks like this was fixed a while back: [npm/npm#5694 (comment)](https://github.com/npm/npm/issues/5694#issuecomment-307142476). It seems all the package managers respect the existing indentation.

Aah, ok, then we can just leave it as-is.

PR #509, opened 2025-06-25
test: add more tests, fix existing for 100% coverage

gamemaker1:
Note that this is based off the `dependencies` branch, so it should be merged after #507.


source/memory-store.ts
```diff
@@ -68,6 +68,7 @@ export class MemoryStore implements Store {
 		}, this.windowMs)
 
 		// Cleaning up the interval will be taken care of by the `shutdown` method.
+		/* istanbul ignore if */
 		if (this.interval.unref) this.interval.unref()
```
nfriedly, line 72: Does this work?
```suggestion
		this.interval.unref?.()
```

source/rate-limit.ts
```diff
@@ -306,7 +306,7 @@ const handleAsyncErrors =
 			await Promise.resolve(fn(request, response, next)).catch(next)
 		} catch (error: unknown) {
 			/* istanbul ignore next */
-			next(error)
+			if (next) next(error)
```
nfriedly, line 309: Why do we need the if here?

source/rate-limit.ts
```diff
@@ -459,9 +462,19 @@ const rateLimit = (
...
+					// disabled for being too flaky.
+					/* istanbul ignore next */
 					response.on('close', async () => {
 						if (!response.writableEnded) await decrementKey()
 					})
+
+					// NOTE: this may not be useful. None of the tests can trigger this
+					// callback (see `/crash` endpoint in test/library/helpers/create-server).
+					// Perhaps it is similar to the case described in this issue comment:
+					// https://github.com/nodejs/node/issues/44884#issuecomment-1270968365
```
nfriedly, line 476: Good catch, that might well be the problem. Could be the cause of the flakiness on the 'close' event test also (?)

source/rate-limit.ts
```diff
@@ -64,6 +64,7 @@ const promisifyStore = (passedStore: LegacyStore | Store): Store => {
 						totalHits: number,
 						resetTime: Date | undefined,
 					) => {
+						/* istanbul ignore if */
```
nfriedly, line 67: IMO we don't really need these istanbul ignore comments. I'm OK with not having 100% test coverage; this kind of feels like we're faking it, anyways.

nfriedly (approved):
Love the new tests! And thanks for investigating the flaky ones.

A couple of suggestions below, but I'm going to go ahead and approve for now because I don't think any of these are too big of a deal.

Oh, and I changed the branch it's targeting to `dependencies` so that it only shows the commits that are newer. Once `dependencies` gets merged in, github will automatically reset the target branch to `main`.

source/memory-store.ts
gamemaker1: Yes it does, nice one, thanks! :)

source/rate-limit.ts
gamemaker1: The jest tests were crashing in between due to `next` not being a function, but that was while I was messing around with the `/crash` endpoint. Let me see if it works without that.

gamemaker1:
Rebased this PR too.

test/library/options-test.ts
```diff
@@ -29,6 +29,20 @@ describe('options test', () => {
 		async resetKey(_key: string): Promise<void> {}
 	}
 
+	it('should allow an empty options object', async () => {
+		expect(rateLimit()).not.toThrow()
```
nfriedly, line 33: Oh, this test is wrong. It should be something like this:
```suggestion
		expect(rateLimit).not.toThrow()
```

or possibly

```suggestion
		expect(() => rateLimit()).not.toThrow()
```

Otherwise expect is trying to execute the returned middleware function with no arguments to see if *it* throws, which is not what the test is supposed to be testing.

source/rate-limit.ts
nfriedly: OK, I think there's enough going on in this PR already, but maybe we can come back to it in the near future.

gamemaker1:
> I'm still waffling on the best default ipv6Subnet value. After writing up the changelog notes and thinking about it for a bit, I'm starting to think we should make the default 64 for 7.6.0 and any backports, and then release a 8.0.0 with the default changed to 56.

I'm not sure how much difference the 56 and 64 default would make. I think the CVE should be enough to bring the users' attention to the change and that it might possibly break existing setups.  

I am fine with making a 8.x release too, with backports to 7.x and 6.x then.

PR #518, opened 2025-07-11
v7.6 > 8.0

nfriedly:


nfriedly:
OK, I deleted `pnpm-workspace.yaml` and then added it to to `.gitignore` to keep it from coming back. It seems like it's presence prevents dependencies from getting installed for the external tests. 

(I didn't intend for it to create that file, it was probably from when I was trying to get pnpm to work.)

PR #526, opened 2025-07-15
IPv6 subnet backport to 7.x 

nfriedly:
I branched off of 7.5.1 to make the 7.x branch, then branched off of that to make this one. I cherry-picked the changes, and then changed the default to /64.

Todo:
* [ ] set up CI to correctly handle publishing of backports (or just do it manually?)
  * It needs to be something like `npm publis --tag backports` to ensure it doesn't get treated as the latest release - https://docs.npmjs.com/cli/v8/commands/npm-publish?v=true#tag
* [ ] Update docs (in `main` branch, not this one)
* [ ] Do we want to add in a ChangeWarning for this? I'm leaning towards no.

Once we're happy with this one, we can do the same thing for 6.x

### DIFF
--- a/docs/reference/changelog.mdx
+++ b/docs/reference/changelog.mdx
@@ -9,6 +9,32 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [7.6.0](https://github.com/express-rate-limit/express-rate-limit/releases/tag/v7.6.0)
+
+Note: There are reasonable arguments for why this release could have been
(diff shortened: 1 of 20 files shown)


docs/reference/changelog.mdx
```diff
@@ -9,6 +9,32 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [7.6.0](https://github.com/express-rate-limit/express-rate-limit/releases/tag/v7.6.0)
```
nfriedly, line 12: Op, need to edit this

PR #527, opened 2025-07-15
docs: document IPv6 backport

nfriedly:
We can merge this into main, then cherry-pick it into the backport branch.

### DIFF
--- a/docs/reference/configuration.mdx
+++ b/docs/reference/configuration.mdx
@@ -306,6 +306,8 @@ rate limiter. This could be combined with a second instance of
 
 ## `ipv6Subnet`
 
+> Added in `8.0.0` (/56 default), `7.6.0` (/64 default).
+
 > `number` (32-64) | `function` | `false`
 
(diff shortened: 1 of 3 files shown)


docs/reference/configuration.mdx
```diff
@@ -306,6 +306,8 @@ rate limiter. This could be combined with a second instance of
 
 ## `ipv6Subnet`
 
+> Added in `8.0.0` (/56 default), `7.6.0` (/64 default).
+
```
nfriedly, line 310: We should probably add this for all of the options.

nfriedly:
Why is the linter complaining about a version mismatch? I didn't change any dependencies in this PR.

gamemaker1 (approved):

docs/reference/configuration.mdx
nfriedly: Well, I mainly meant that we should document what version each option was added in. But, for ones where the default changed, then maybe we should do it like this.

PR #532, opened 2025-07-16
fix: export ipKeyGenerator in cjs build

### LINKED ISSUE #529: TypeError: ipKeyGenerator is not a function
### Description

```js
const express = require("express");
const { rateLimit, ipKeyGenerator } = require("express-rate-limit");

const app = express();

const createOrderLimiter = rateLimit({
  windowMs: 15 * 60 * 1000, // 15 minutes
  max: 10, 
  keyGenerator: (req) =>
    ipKeyGenerator(req.headers["cf-connecting-ip"] || req.ip),
  message: "Too many payment creation attempts. Please try again later.",
  standardHeaders: true,
  legacyHeaders: false,
});

app.use("/api/payment/create-order", createOrderLimiter);
```
When i go to the endpoint `/api/payment/create-order` i get the error 
```console
TypeError: ipKeyGenerator is not a function
```

### Library version

^8.0.0

### Node version

v22.17.0

_No response_

### Module system

CommonJS

nfriedly:
We add a footer to the cjs that redefines the exports in order to keep the original import style\[1\] working, but it had been hard-coded to specific fields rather than just re-exporting everything. Now it re-exports everything, so we shouldn't need to worry about it again if we add more exports in the future.

Fixes #529

\[1\]: 
```js
var rateLimit = require("express-rate-limit");
//...
app.use(rateLimit({ /*...*/ });
```

(For some annoying reason, every TS transpire seems to break this and only exports an object with a `default` field, instead of making the default export the actual thing.)

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,7 @@
 	},
 	"scripts": {
 		"clean": "del-cli dist/ coverage/ *.log *.tmp *.bak *.tgz",
-		"build:cjs": "esbuild --packages=external --platform=node --bundle --target=es2022 --format=cjs --outfile=dist/index.cjs --footer:js=\"module.exports = rateLimit; module.exports.default = rateLimit; module.exports.rateLimit = rateLimit; module.exports.MemoryStore = MemoryStore;\" source/index.ts",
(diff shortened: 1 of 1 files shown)


gamemaker1:
Nice, thank you!

PR #534, opened 2025-07-16
fix: express Request and Response types

### LINKED ISSUE #530: Incorrect types for Request / Response
### Description

The newly released v8.0.0 seems to have incorrect types for Request & Response. They are pointing to `@types/node` which seems to be because in the `.d.ts` files the proper Express types are imported but with an alias that is then unused:

```ts
import { NextFunction, Request as Request$1, RequestHandler, Response as Response$1 } from 'express';
```

### Library version

8.0.0

### Node version

v22.17.0

5.8.3

### Module system

CommonJS

nfriedly:
This reverts dts-bundle-generator to fix the incorrect renaming of express's Request and Response types in the generated .d.ts files

Fixes #530

related: https://github.com/timocov/dts-bundle-generator/issues/353

### DIFF
--- a/package.json
+++ b/package.json
@@ -90,7 +90,7 @@
 		"@types/node": "24.0.14",
 		"@types/supertest": "6.0.3",
 		"del-cli": "6.0.0",
-		"dts-bundle-generator": "9.5.1",
+		"dts-bundle-generator": "8.0.1",
 		"esbuild": "0.25.6",
 		"express": "5.1.0",
 		"husky": "9.1.7",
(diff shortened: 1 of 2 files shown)


nfriedly:
> Aah this issue.. we should consider dropping the CJS builds in a future major version bump once [`require`ing sync ESM modules](https://nodejs.org/api/modules.html#loading-ecmascript-modules-using-require) gets released into a stable node version.

Hum.. maybe, but I might suggest waiting until it's in LTS releases also. And maybe a bit after that.

PR #535, opened 2025-07-16
fix: replace vulnerable ip package

nfriedly:
This drops the `ip` library that v8.0.0 required due to an open vulnerability in the library. express-rate-limit did not use the vulnerable code, but it still caused warnings and reports for end users.

The replacement is a much smaller and more focused bit of code generated by Google Gemini that handles the ip bit fiddling we need to apply the subnet mask, but not much else.

The code looks correct to me, and passes all of our tests (including the ones where supertest sets the source IP to `::ffff:127.0.0.1`), and passes the tests that gemini wrote to go with it (which also look correct to me).

That said, please review with extra scrutiny since this code wasn't written by a human.

Fixes https://github.com/express-rate-limit/express-rate-limit/issues/533


nfriedly:
I thought about this a bit more and came up with an alternate that doesn't have any AI code in it: https://github.com/express-rate-limit/express-rate-limit/pull/536

gamemaker1:
This looks like a lot of code that will need continuous maintenance and patches in case of vulnerabilities. It might be best to rely on another well maintained package, even though it does mean adding a dependency.

Does the [`networkAddressFromCIDR`](https://github.com/whitequark/ipaddr.js/blob/main/lib/ipaddr.js#L839) function of [`ipaddr.js`](https://github.com/whitequark/ipaddr.js) look like something we could use?

PR #549, opened 2025-08-20
Add validation check for Forwarded header

nfriedly:
This alerts the user if a `Forwarded` header (standardized version of `X-Forwarded-For` & friends) is set when using the default `keyGenerator`, similar to our check for `X-Forwarded-For` when `trust proxy` is not enabled.

I also added an example to the docs with a custom `keyGenerator` that uses the [forwarded-parse](https://www.npmjs.com/package/forwarded-parse) library to parse and use the rightmost forwarded header (e.g. the last one, from the nearest proxy server).

This doesn't quite fix https://github.com/express-rate-limit/express-rate-limit/issues/488 but it definitely improves the situation. 

I'm hoping that support will eventually be added to express, e.g. via [proxy-addr](https://github.com/jshttp/proxy-addr/issues/20). If it doesn't, I may consider adding built-in support to express-rate-limit. But for now, at least, we have a warning and an example.

### DIFF
--- a/docs/guides/troubleshooting-proxy-issues.mdx
+++ b/docs/guides/troubleshooting-proxy-issues.mdx
@@ -12,7 +12,8 @@ request might be the IP of the load balancer/reverse proxy (making the rate
 limiter effectively a global one and blocking all requests once the limit is
 reached) or `undefined`.
 
-To solve this issue, please follow the steps given below.
+To solve this issue, assuming the proxy (or proxies) set the `X-Forwarded-For`
+header, please follow the steps given below.
 
 <Steps>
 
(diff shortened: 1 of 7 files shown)


nfriedly:
Oh, that's cool, you can see a preview of the new documentation at https://express-rate-limit-check-forwarded.mintlify.app/reference/error-codes#err-erl-forwarded-header

nfriedly:
Oh, I forgot about `X-Real-IP` (Nginx), `CF-Connecting-IP` (Cloudflare), or `True-Client-IP` (Cloudflare and Akamai), and others. I wonder if they should each get their own error code so that we can have accurate sample code (and instructions - some of them have some serious flaws if mis-configured.)

I think this is complete as-is for `Forwarded`, though, so maybe that should come in a followup PR.

nfriedly:
I think I want to make one more change to this: I want to suppress the warning if req.ip, is anything other than the req.connection.remoteAddress, on the assumption that if it's already been set to something, it's probably not wrong. I have it coded up with a test, but my tests are failing so I need to figure out why.

nfriedly:
Ok, I think this is ready. If you're happy with it @gamemaker1, lets merge it in and put out a release with what we have so far. We can work on proprietary headers in a future release.

(Relatedly, while I think they should probably have different error codes, they could all share the same validation check.)

gamemaker1:
> Relatedly, while I think they should probably have different error codes, they could all share the same validation check.

Sounds good.

gamemaker1 (approved):

PR #550, opened 2025-08-21
Document windowMs limit for MemoryStore and warn on invalid values

### LINKED ISSUE #546: Bug: windowMs does not apply correctly for long durations
### Description

Hi, dear.

I encountered this issue while trying to limit the number of password reset attempts.

I wanted to allow a maximum of 5 password resets per year, so I set the `windowMs` value to 365 days. However, I noticed that the limit value was not resetting as expected. While inspecting related options like keyGenerator, I found that the issue seemed to stem from windowMs.

Interestingly, when I reduced the time window to one hour, it worked correctly.

Maybe the default store, memory-store, uses the setTimeout API to handle the windowMs timer. However, since setTimeout only supports delays up to approximately 24.8 days, it seems the key was deleted and reset prematurely due to this limitation.

I couldn’t find any mention of time duration limits for windowMs in the official documentation. Could you please confirm whether there is a maximum allowed duration?

```
windowMs: 10 * 60 * 1000 // works fine
windowMs: 365 * 24 * 60 * 60 * 1000 // not working
```

### Library version

8.0.1

### Node version

v22.13.1

5.8.3

### Module system

ESM

nfriedly:
This is mainly to catch values over node.js's maximum (~28.4 days), but it will also catch other invalid values such as negative numbers and NaN.

Fixes #546 (Sort of. It at least clarifies things - it's about as much of a fix as we're going to do.)

### DIFF
--- a/docs/reference/configuration.mdx
+++ b/docs/reference/configuration.mdx
@@ -25,11 +25,18 @@ create multiple rate limiters.
 <Note>
 
 For stores that do not implement the `init` function (including all legacy
-stores - see the [data stores page](/reference/stores)), you may need to
-configure this value twice, once here and once on the store. In some cases the
-units also differ (e.g. seconds vs milliseconds).
+stores - see the [data stores page][stores]), you may need to configure this
+value twice, once here and once on the store. In some cases the units also
+differ (e.g. seconds vs milliseconds).
 
 </Note>
+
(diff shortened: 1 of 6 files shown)


gamemaker1:
Hi, so sorry for not being able to respond earlier, I got caught up in submission deadlines and exams.

Please do go ahead and merge the PR, I've just added a link to the documentation that mentions the upper limit in the comments of the validation function.

PR #553, opened 2025-08-31
fix: include RateLimit-Reset header when resetSeconds is 0

### LINKED ISSUE #552: RateLimit-Reset header appears to be incorrectly omitted when value is 0 due to falsy check
I believe the `RateLimit-Reset` header may be incorrectly omitted when the reset value is `0`, not because the value is absent, but due to what appears to be an unintended falsy check.

**Potentially problematic code in `source/headers.ts:101-102`:**
```typescript
// Set this header only if the store returns a `resetTime`.
if (resetSeconds)  // This seems to skip when resetSeconds is 0
    response.setHeader('RateLimit-Reset', resetSeconds.toString())
```

**When this occurs**

`resetSeconds` becomes 0 when:
- Rate limit window has just expired `(resetTime.getTime() - Date.now() <= 0)`
- `Math.max(0, deltaSeconds)` correctly returns 0
- But if (0) evaluates to false, so the header gets omitted

**Suggested behavior**

Also set the header when `resetSeconds` is equal to 0

```typescript
// Set this header only if the store returns a `resetTime`.
if (typeof resetSeconds === 'number')
    response.setHeader('RateLimit-Reset', resetSeconds.toString())
```

What do you think? Am I missing something about why the falsy check is there?

vieiralucas:
Previously, the RateLimit-Reset header was incorrectly omitted when resetSeconds was 0 due to a falsy check. This occurred when the rate limit window had just expired (resetTime <= current time), causing Math.max(0, deltaSeconds) to return 0.

Changed the condition from `if (resetSeconds)` to
`if (typeof resetSeconds === 'number')` to ensure the header is included for all numeric values, including 0.

Fix #552

### DIFF
--- a/source/headers.ts
+++ b/source/headers.ts
@@ -98,7 +98,7 @@ export const setDraft6Headers = (
 	response.setHeader('RateLimit-Remaining', info.remaining.toString())
 
 	// Set this header only if the store returns a `resetTime`.
-	if (resetSeconds)
+	if (typeof resetSeconds === 'number')
 		response.setHeader('RateLimit-Reset', resetSeconds.toString())
 }
 
--- a/test/library/headers-test.ts
+++ b/test/library/headers-test.ts
@@ -336,4 +336,48 @@ describe('headers test', () => {
 			})
 		})
 	})
+
+	it('should send RateLimit-Reset header with value 0 when window has just expired for draft-6', async () => {
+		class MockStore implements Store {
+			hits: Map<string, number> = new Map()
+
+			async get(key: string): Promise<ClientRateLimitInfo> {
+				return {
+					totalHits: this.hits.get(key) ?? 0,
+					resetTime: new Date(Date.now()), // Current time - this results in resetSeconds = 0
+				}
+			}
+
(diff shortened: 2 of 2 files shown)


nfriedly (approved):
Nice work!

I started to suggest using fake timers in the test to ensure the clock doesn't change between when the store sets the expiration to "now", and when the limiter computes the header value, but it would go negative if that happened, and `getResetSeconds` will return `0` in that case. So fake timers aren't needed, this test should behave correctly regardless of how long it takes to execute.

I'll merge this now and publish a new version to npm today or tomorrow; there are a couple of other changes I might try to bundle in.

vieiralucas:
@nfriedly any news on making a release?

nfriedly:
Apologies, I was wanting to include https://github.com/express-rate-limit/express-rate-limit/pull/549 also, but it ended up needing a bit of additional work.

nfriedly:
Ok, this is now published to npm in v8.1.0: https://www.npmjs.com/package/express-rate-limit

Thanks again!

PR #578, opened 2025-10-24
feat: Unknown Options validation check

### LINKED ISSUE #571: Validation check for unexpected config option, e.g. windowMS instead of windowMs
### Description

A spam comment alerted me to [this SO post](https://stackoverflow.com/questions/71945791/why-is-express-rate-limit-not-working-as-expected) where someone typed `windowMS` instead of `windowMs` in their config and express-rate-limit silently ignored the mistake and used the default window instead of the one they intended. 

I think we should probably warn on unexpected config options and, maybe even calculate the levenshtein distance from the allowed options to suggest an alternative.

### Why

I want it to be easier to use the library correctly than incorrectly. Even when the user isn't using typescript.

### Alternatives

do nothing

nfriedly:
This logs a warning when an unknown option is present, to catch typos like `windowMS` (which should be `windowMs`).

I considered trying to suggest a likely alternative, but it would involve quite a bit more code, so skipping that for now.

Fixes #571

I wrote this code over the weekend, and I feel like there was something else I wanted to add to this, but I can't remember what, so leaving it as a draft for right now.

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -1,5 +1,5 @@
 {
-	"$schema": "https://biomejs.dev/schemas/2.2.0/schema.json",
+	"$schema": "https://biomejs.dev/schemas/2.2.5/schema.json",
 	"vcs": {
 		"enabled": false,
 		"clientKind": "git",


source/validations.ts
```diff
@@ -313,6 +313,46 @@ const validations = {
 		}
 	},
 
+	knownOptions(passedOptions?: Partial<Options>) {
+		if (!passedOptions) return // no options is a valid config
+
+		type KeysEnum<T> = { [P in keyof Required<T>]: true }
+		// we have to manually update this when adding new options, because we want it to work even for JS users. But KeysEnum makes TypeScript ensures it's correct at build time!
+		const optionsMap: KeysEnum<Options> = {
```
gamemaker1: I don't like that we have to repeat the entire list again (once in the types, once as part of the `Configuration`), but oh well.

nfriedly:
Oh, I'll add it to the changelog before I merge.

PR #610, opened 2026-03-15
test: use numeric range for reset timestamp assertion

### LINKED ISSUE #608: Headers test for reset time can be flakey
This line https://github.com/express-rate-limit/express-rate-limit/blob/eb61179a49064c7e86f6f8688be742343b1f1b8e/test/library/headers-test.ts#L47 failed in https://github.com/express-rate-limit/express-rate-limit/actions/runs/22857490378/job/66301735150 because the actual timestamp was 1 second higher than the range it was expecting. It's currently using a regex that ignores the last two digits, and so the size of the range depends on exactly when the test is run. We should probably treat it as a number rather than a string so that the allowed range can be more consistent.

I don't have time to fix it right now, so just filing this to remind me later. (Or it could be a good first bug for someone else to fix.)

mahmoodhamdi:
## Summary

- Fixes a flaky test for the `x-ratelimit-reset` header in the legacy headers test
- The old approach used a regex that stripped the last 2 digits from the expected Unix timestamp, creating a variable-width tolerance window (0-99 seconds depending on when the test ran)
- When the timestamp ended in `99`, a 1-second delay during test execution could push the actual value past the regex boundary (e.g., regex expected `...170000XX` but got `...170001XX`)
- Replaced with a numeric comparison using `toBeGreaterThanOrEqual` / `toBeLessThanOrEqual` allowing exactly +/- 1 second tolerance

## Before (flaky)
```ts
const resetRegexp = new RegExp(
  `^${expectedResetTimestamp.slice(0, -2)}\\d\\d$`,
)
// Tolerance depends on where in the 100-second window the test runs
```

## After (deterministic)
```ts
const actualResetTimestamp = Number(response.get('x-ratelimit-reset'))
expect(actualResetTimestamp).toBeGreaterThanOrEqual(expectedResetTimestamp - 1)
expect(actualResetTimestamp).toBeLessThanOrEqual(expectedResetTimestamp + 1)
```

## Test results

All 179 tests pass:
```
Test Suites: 1 skipped, 6 passed, 6 of 7 total
Tests:       1 skipped, 179 passed, 180 total
```

Fixes #608

### DIFF
--- a/test/library/headers-test.ts
+++ b/test/library/headers-test.ts
@@ -31,9 +31,7 @@ describe('headers test', () => {
 			}),
 		)
 
-		const expectedResetTimestamp = Math.ceil(
-			(Date.now() + 60 * 1000) / 1000,
-		)
+		const expectedResetTimestamp = Math.ceil((Date.now() + 60 * 1000) / 1000)
 
 		const response = await request(app)
 			.get('/')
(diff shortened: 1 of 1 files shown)


mahmoodhamdi:
Done, ran `npm run format` and pushed the fix. Thanks for the review!

PR #611, opened 2026-03-16
Stabilize skipFailedRequests close-event test and fix lint

AzmeerX:
Description:
Rewrite the close-event test in test/library/middleware-test.ts to use mock request/response objects, emit close directly, and await the async decrement path; adds EventEmitter import and uses Promise.resolve() wait to avoid timer flakiness.
Adjust Express type imports to import type to satisfy Biome lint.

Tests: 
npm test (runs lint and jest).

### DIFF
--- a/test/library/middleware-test.ts
+++ b/test/library/middleware-test.ts
@@ -15,8 +15,6 @@ import rateLimit, {
 } from '../../source/index.js'
 import { createServer } from './helpers/create-server.js'
 
-const { platform } = process
-
 describe('middleware test', () => {
 	beforeEach(() => {
 		jest.useFakeTimers()


nfriedly (changes requested):
Hey, thanks again - I think it was a really good insight, figuring out that the issue was async work happening before we attached the event handlers, and that the solution was to attach the event handlers earlier.

That said, I think the proposed solution here is both a bit more complex than necessary, and also doesn't go far enough.

I think we need to attach the event handlers even before the skip determination (or else require skip to be non-async, but that would be a breaking change). 

You can create success & fail promises right at the start, something like this

```js
const reqFinished: false | Promise<void> = config.skipSuccessfulRequests && new Promise(resolve => response.on('finish', resolve));
// and similar for close/error
```

Then the rest of the code continues unchanged until you get to the `if (config.skipFailedRequests || config.skipSuccessfulRequests) {` section - in there, instead of adding event handlers, await the earlier promises and then perform the appropriate action.

I haven't tested all of this myself, but I believe it would make for a simpler and more complete solution to the issue.

With that in mind, you can probably ignore all of my other comments and just implement this solution instead. (Or if you don't have the time or hit any issues, let me know and I'll try it out myself.)

source/rate-limit.ts
```diff
@@ -361,8 +376,54 @@ const rateLimit = (
...
+						void config.store.decrement(key)
+					}
+				}
+
+				if (config.skipFailedRequests) {
+					if (closePromise) {
+						void (async () => {
+							await closePromise
+							if (!response.writableEnded) decrement()
+						})()
```
nfriedly: I don't think we need to wrap this in a new function. Same for the others below.

nfriedly (changes requested):
Nice work, this looks a lot closer to what I was thinking, but I have a few more changes requested below.

nfriedly (changes requested):
Awesome work! I think the library code is basically perfect, just a comment change there. 

And then for the tests, I'd like to move your changes into a new test and also see the old test passing un-modified, just as an extra confirmation that the code is correct.

nfriedly (approved):
That's great, thanks! I made one tweak to make the old test run on macOS, and it apparently needs some formatting now that it's uncommented, but I'll take care of that.

Thanks again!

nfriedly:
This is now released in v8.3.2 🎉 

Sorry that the history got messed up, I think that was my fault. I was trying to do the formatting in your branch, then merge into main, but I think I accidentally did the changes in main. I reversed that and fixed main, with a squash merge commit for this PR, but github hid all of the commits here when they went into main and then didn't put them back when they were removed. The merge commit, however, has references to all of the individual commits that were part of this PR, see https://github.com/express-rate-limit/express-rate-limit/commit/601b87f5d171487ed035ccdfee17ec75f5b22f2d

nfriedly:
It seems that we're still getting CI failures related to this:

https://github.com/express-rate-limit/express-rate-limit/actions/runs/24293613741/job/71299134481?pr=616


> FAIL test/library/middleware-test.ts (38.752 s)
>   ● middleware test › should decrement hits when response closes and `skipFailedRequests` is set to true (compat store) (server)
> 
>  thrown: "Exceeded timeout of 30000 ms for a test.
>  Add a timeout value to this test to increase the timeout, if this is a long-running test. See https://jestjs.io/docs/api#testname-fn-timeout."
> 
> ```
>       625 | 		['legacy', new MockLegacyStore()],
>       626 | 		['compat', new MockBackwardCompatibleStore()],
>     > 627 | 	])('should decrement hits when response closes and `skipFailedRequests` is set to true (%s store) (server)', async (name, store) => {
>           | 	  ^
>       628 | 		jest.useRealTimers()
>       629 | 		jest.setTimeout(60_000)
>       630 |
> ```

In addition to the unexpected failure, the timeout seems wrong - we're setting it to 60 seconds but the error message says 30.

PR #616, opened 2026-04-11
feat: allow usage of custom logger

### LINKED ISSUE #615: Make logger or error handling configurable
### Description

When validations are enabled, messages are printed on stdout/stderr directly, with no option to disable the behavior (except disabling the validations).

I see two main possibilities to implement that keeps the API backwards-compatible:
- Add optional logger configuration option, that if unset uses the console to print the messages.
- Add an optional validation error handler, with default implementation being the current (console.error/console.warn)

There is also the special case of passOnStoreError that also uses the console outside of the validations

Would there be an interest for this feature? If yes, I can create a PR, I'd just like to know what would be your preferred option?

If the logger object, what convention should be used?
- Same interface as console ?
- NestJS-style (message then object)
- Pino-style (object then message and others)
- other?

### Why

When indcluded in application that implements structured logging, it can create problems in the pipeline, or add noise that can be difficult to interpret as the message is multiline.

### Alternatives

I haven't found a workaround except disable validations.

piotaixr:
Closes #615 

Logger is optional, but if provided must implement the `Logger` interface to follow the same behavior as for the store.

### DIFF
--- a/docs/reference/configuration.mdx
+++ b/docs/reference/configuration.mdx
@@ -528,3 +528,25 @@ Supported validations are:
 Defaults to `true`.
 
 [stores]: /reference/stores
+
+
+##  `logger`
+
+> `Logger`
+
+The Logger instance to use to log validation errors or warnings. Default behavior
+is to use the console. The first argument is the error that was raised,
+the second is an optional message explaining the error. For example:
+
+```js
+  const limiter = rateLimit({
+    logger: {
+      warn: (error, message) => {
+        // TODO: do something with error and message
+      },
+      error: (error, message) => {
+        // TODO: do something with error and message
+      }
+    }
+  })
+```
--- a/docs/reference/configuration.mdx
+++ b/docs/reference/configuration.mdx
@@ -529,24 +529,23 @@ Defaults to `true`.
 
 [stores]: /reference/stores
 
-
-##  `logger`
+## `logger`
 
 > `Logger`
 
-The Logger instance to use to log validation errors or warnings. Default behavior
-is to use the console. The first argument is the error that was raised,
+The Logger instance to use to log validation errors or warnings. Default
+behavior is to use the console. The first argument is the error that was raised,
 the second is an optional message explaining the error. For example:
 
 ```js
-  const limiter = rateLimit({
-    logger: {
-      warn: (error, message) => {
-        // TODO: do something with error and message
-      },
-      error: (error, message) => {
-        // TODO: do something with error and message
-      }
-    }
-  })
+const limiter = rateLimit({
+	logger: {
+		warn: (error, message) => {
+			// TODO: do something with error and message
+		},
+		error: (error, message) => {
+			// TODO: do something with error and message
+		},
+	},
+})
 ```
--- a/source/logger.ts
+++ b/source/logger.ts
@@ -1,0 +1,15 @@
+export type LoggerFn = (context: unknown, message?: string) => void
+
+export type Logger = {
+	error: LoggerFn
+	warn: LoggerFn
+}
+
+export const ConsoleLogger: Logger = {
+	warn(...args): void {
+		console.warn(...args.reverse())
+	},
+	error(...args) {
+		console.error(...args.reverse())
+	},
+}
--- a/source/rate-limit.ts
+++ b/source/rate-limit.ts
@@ -11,6 +11,7 @@ import {
 	setRetryAfterHeader,
 } from './headers.js'
 import { ipKeyGenerator } from './ip-key-generator.js'
+import { ConsoleLogger, type Logger } from './logger'
 import { MemoryStore } from './memory-store.js'
 import type {
 	AugmentedRequest,
@@ -122,6 +123,7 @@ type Configuration = {
 	store: Store
 	validations: Validations
 	passOnStoreError: boolean
+	logger: Logger
 }
 
 /**
@@ -155,7 +157,10 @@ const parseOptions = (passedOptions: Partial<Options>): Configuration => {
 		omitUndefinedProperties<Partial<Options>>(passedOptions)
 
 	// Create the validator before even parsing the rest of the options.
-	const validations = getValidations(notUndefinedOptions?.validate ?? true)
+	const validations = getValidations(
+		notUndefinedOptions?.validate ?? true,
+		passedOptions.logger,
+	)
 	validations.validationsConfig()
 
 	// Warn on unknown options
@@ -284,6 +289,7 @@ const parseOptions = (passedOptions: Partial<Options>): Configuration => {
 		),
 		// Print an error to the console if a few known misconfigurations are detected.
 		validations,
+		logger: passedOptions.logger ?? ConsoleLogger,
 	}
 
 	// Ensure that the store passed implements the `Store` interface
@@ -297,6 +303,17 @@ const parseOptions = (passedOptions: Partial<Options>): Configuration => {
 	) {
 		throw new TypeError(
 			'An invalid store was passed. Please ensure that the store is a class that implements the `Store` interface.',
+		)
+	}
+
+	// Ensure that the logger passed implements the `Logger` interface
+	if (
+		typeof config.logger !== 'object' ||
+		typeof config.logger.error !== 'function' ||
+		typeof config.logger.warn !== 'function'
+	) {
+		throw new TypeError(
+			'Provided logger does not implement the Logger interface',
 		)
 	}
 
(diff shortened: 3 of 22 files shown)


piotaixr:
@gamemaker1 @nfriedly PTAL 🙏 

I went with Pino/Bunyan convention (context/error first, then the message) because the message is optional (only provided with the passOnStoreError options)

source/logger.ts
```diff
@@ -0,0 +1,15 @@
...
+}
+
+export const ConsoleLogger: Logger = {
+	warn(...args): void {
+		console.warn(...args.reverse())
+	},
+	error(...args) {
+		console.error(...args.reverse())
+	},
+}
```
gamemaker1: Could you please add jsdoc style comments to this, like it is done in `source/types.ts`?

source/rate-limit.ts
```diff
@@ -155,7 +157,10 @@ const parseOptions = (passedOptions: Partial<Options>): Configuration => {
 		omitUndefinedProperties<Partial<Options>>(passedOptions)
 
 	// Create the validator before even parsing the rest of the options.
-	const validations = getValidations(notUndefinedOptions?.validate ?? true)
+	const validations = getValidations(
+		notUndefinedOptions?.validate ?? true,
+		passedOptions.logger,
```
gamemaker1, line 188: Shouldn't this also have a `?? ConsoleLogger` suffix?

source/rate-limit.ts
```diff
@@ -300,6 +306,17 @@ const parseOptions = (passedOptions: Partial<Options>): Configuration => {
...
+	if (
+		typeof config.logger !== 'object' ||
+		typeof config.logger.error !== 'function' ||
+		typeof config.logger.warn !== 'function'
+	) {
+		throw new TypeError(
+			'Provided logger does not implement the Logger interface',
+		)
+	}
+
```
gamemaker1: Maybe move this check above the validation thing I commented on above.

test/library/logger-test.ts
```diff
@@ -0,0 +1,45 @@
+import { afterEach, beforeEach, expect } from '@jest/globals'
+import { ConsoleLogger } from '../../source/logger'
+
+describe('ConsoleLogger', () => {
```
gamemaker1, line 4: Could we also modify the external test suite (not this file, `tests/external/store-test/`) to use a pino logger and make sure it works?

docs/reference/configuration.mdx
```diff
@@ -528,3 +528,25 @@ Supported validations are:
...
+    logger: {
+      warn: (error, message) => {
+        // TODO: do something with error and message
+      },
+      error: (error, message) => {
+        // TODO: do something with error and message
+      }
+    }
+  })
+```
```
gamemaker1: In the future (not necessarily this PR) we could add a guide to integrating pino/etc using this option.

gamemaker1 (changes requested):
This looks great! Thank you for contributing :)

I have requested some changes regarding one possible bug, one more test, and some documentation. Please take a look at the comments once.

source/rate-limit.ts
piotaixr: I wanted to avoid changing the interface of the getValidations function, so I made the logger parameter default to ConsoleLogger, but as it is not exported, there is non need for this.
I updated the function and changegd the tests (that become simpler and no not require to mock the global console object).

source/rate-limit.ts
piotaixr: Yes, I see why 😓 
I moved it at the very top of the parseOptions function so we exit as early as possible if the passed logger in not valid.

source/logger.ts
piotaixr: I added the jsdoc on the interface, is it enough or should I also add it on the ConsoleLogger implementation?

test/library/validation-test.ts
nfriedly, line 1: Since the default logger still calls console.warn/error, we probably don't need as many changes here, right?

Instead of changing this file, I think we could add a couple more expect's to the 'should use the logger instead of the console on validation errors' test to ensure that the correct logger was getting called.

test/library/logger-test.ts
piotaixr: Sorry, didn't have time to look I to this. I'll take a look soon.

What do you mean by 'make sure it works'? Should I assert that the logger is effectively called, or just add it as dependency of the test project and configure it?
If I need to assert, I might need to create a separate test project because I'll need to have an application that has invalid configuration or bad behavior in order to exercise the warn/error logger.

test/library/validation-test.ts
piotaixr: Yes, I kind of started this way, but with the change in function signature of `getValidations` (now requiring a logger to be passed) we are effectively injecting a logger instance.
It would be a little sad to not use this to simplify the tests (by not mocking global variables), right?

nfriedly: Aah, ok, I didn't consider that.

test/library/logger-test.ts
piotaixr: Done!

piotaixr:
@gamemaker1 PTAL, I adressed all comments.

source/rate-limit.ts
```diff
@@ -141,6 +144,28 @@ const getOptionsFromConfig = (config: Configuration): Options => {
...
+ *
+ * @param logger {Logger | undefined}
+ *
+ * @throws {TypeError} if the provided logger incorrectly implements the {@see Logger} interface
+ * @returns {Logger} -- Thr provided Logger if valid, or the ConsoleLogger if absent.
+ */
+const validateLogger = (logger?: Logger): Logger => {
+	if (logger === undefined) {
+		return ConsoleLogger
+	}
```
nfriedly: Hum, I don't like having a function named "validate X" swap in a different X if the provided one is undefined. Maybe this should only validate and then return void if it passes?

In fact, let's move this to the validations file, and either run it in a regular validation check, or else in the getValidations() method. (A regular validation check might be difficult because we need to send the results of those to the logger method, so probably call it from getValidations().)

source/rate-limit.ts
```diff
@@ -154,8 +179,14 @@ const parseOptions = (passedOptions: Partial<Options>): Configuration => {
 	const notUndefinedOptions: Partial<Options> =
 		omitUndefinedProperties<Partial<Options>>(passedOptions)
 
+	// Make sure we have a valid Logger
+	const logger = validateLogger(passedOptions.logger)
```
nfriedly: And then here you'd just have 


```suggestion
	const logger = passedOptions.logger ?? ConsoleLogger
```

source/rate-limit.ts
nfriedly: ... and this would be 
```suggestion
		logger,
```

source/types.ts
```diff
@@ -401,6 +423,11 @@ export type Options = {
 	 * If the Store generates an error, allow the request to pass.
 	 */
 	passOnStoreError: boolean
+
+	/**
+	 * The logger to use to log errors. If absent, logs to the console.
+	 */
+	logger?: Logger
```
nfriedly: I don't think we need the questionmark here, because the rateLimit function accepts a `Partial<Options>`, so everything is already optional. We just put the question mark on the deprecated ones because they're not expected.

```suggestion
	logger: Logger
```

source/validations.ts
```diff
@@ -339,6 +341,7 @@ const validations = {
 			headers: true,
 			max: true,
 			passOnStoreError: true,
+			logger: true,
```
nfriedly, line 343: I don't think this is needed. We would only want it if we were going to make validating the logger optional, and make it a regular validation check, but I don't think we want to do either of those things.

```suggestion
```

nfriedly (commented):
Hey, thank you for this, you've done great work here! I sorry we haven't gotten back to you very quickly, I've have a really crazy couple of weeks at work, and I'm sure @gamemaker1 has had a lot going on too.

I have a few comments below, mostly about the `validateLogger()` function, but I think this is looking really good, especially the integration tests.

source/validations.ts
piotaixr: If I don't add it, Typescript complains with the following:
```
error TS2741: Property 'logger' is missing in type '{ windowMs: true; limit: true; message: true; statusCode: true; legacyHeaders: true; standardHeaders: true; identifier: true; requestPropertyName: true; skipFailedRequests: true; skipSuccessfulRequests: true; ... 9 more ...; passOnStoreError: true; }' but required in type 'KeysEnum<Options>'.
```

nfriedly: Oh, you're right, I'm sorry. I forgot what this list was, I was thinking it was a list of validations that could be disabled, not a list of options that users could pass in. Keep it here.

piotaixr:
I think I fixed all comments :)

nfriedly (approved):

nfriedly:
Ok, this should be published by CI shortly in v8.4.0 🎉 

Thanks again!

gamemaker1:
Hi, sorry about the late reply, I got really caught up in a bunch of deadlines. Thanks @nfriedly for reviewing and shipping it and thank you @piotaixr for addressing all the comments and adding this great feature :)

PR #617, opened 2026-04-18
resolve Jest timeout for server-based tests

AzmeerX:
Moved jest.setTimeout(60_000) into a beforeEach hook. this ensures the timeout is configured before the tests run
Jest's timeout enforcement now applies the extended limit from the start

Changes:

Uncommented import { platform } from 'node:process' (for platform-specific test skipping)
Wrapped the long-running server test in a describe block with proper timeout setup
Moved jest.useRealTimers() call to remain within the test function where it's needed

Fixed #611


nfriedly (approved):

nfriedly:
thanks!

nfriedly:
Oof, this didn't do the trick either: https://github.com/express-rate-limit/express-rate-limit/actions/runs/24637943753/job/72036615480

There's the 30/60 second thing which still doesn't seem to be working correctly, but that might be a red herring - this test should really be passing in under 30 seconds, so I think there's still some underlying flakiness that needs to be sussed out.

PR #621, opened 2026-04-25
feat: async store init

nfriedly:
This allows store init methods to be async / return a promise, and if it rejects, the error will be logged.

This is mainly to prevent unhandled rejections from store initialization that would otherwise take down the server. (Those could be recoverable if it's just a networking issue.) 

In particular, the redis store sends an early script upload, but it's able to retry that when handling requests, so it may recover later if the first upload fails.  But right now failures take down the server, so it doesn't get the opportunity to recover. (Also it sends the requests off in the constructor, but we'll move them to the init() method after this.)

Errors thrown synchronously are not caught (no change in behavior there), because I expect they generally wouldn't be recoverable. (And users can wrap their limiter creation in a try/catch if they believe it is a recoverable error.)

Relates to https://github.com/express-rate-limit/rate-limit-redis/issues/239 and https://github.com/express-rate-limit/express-rate-limit/issues/485

We could go farther and await the result before handling requests, but then I'm not sure what we'd do if it failed - try again, or allow/reject according to the `passOnStoreError` setting, or call store.increment() despite the failure? I think some of those might constitute breaking changes, and I think what's here is enough to handle the immediate need. So I think I want to just go with this more limited fix for now.

### DIFF
--- a/source/types.ts
+++ b/source/types.ts
@@ -169,7 +169,6 @@ export type Store = {
 	 * Called once during initialization.
 	 *
 	 * Async errors / promise rejections will be caught and logged.
-	 * (Synchronously thrown errors are not caught, and will prevent the express-rate-limit instance from being created.)
 	 *
 	 * Note that the result is not awaited - other store methods (such as increment) may be called before init returns and/or after it rejects.
 	 *
--- a/source/types.ts
+++ b/source/types.ts
@@ -166,9 +166,16 @@ export type Store = {
 	 * Method that initializes the store, and has access to the options passed to
 	 * the middleware too.
 	 *
+	 * Called once during initialization.
+	 *
+	 * Async errors / promise rejections will be caught and logged.
+	 * (Synchronously thrown errors are not caught, and will prevent the express-rate-limit instance from being created.)
+	 *
+	 * Note that the result is not awaited - other store methods (such as increment) may be called before init returns and/or after it rejects.
+	 *
 	 * @param options {Options} - The options used to setup the middleware.
 	 */
-	init?: (options: Options) => void
+	init?: (options: Options) => void | Promise<void>
 
 	/**
 	 * Method to fetch a client's hit count and reset time.
--- a/source/types.ts
+++ b/source/types.ts
@@ -168,9 +168,9 @@ export type Store = {
 	 *
 	 * Called once during initialization.
 	 *
-	 * Async errors / promise rejections will be caught and logged.
+	 * Errors / promise rejections will be caught and logged.
 	 *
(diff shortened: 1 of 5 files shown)


test/library/middleware-test.ts
```diff
@@ -174,6 +181,67 @@ describe('middleware test', () => {
 		expect(store.initWasCalled).toEqual(true)
 	})
 
+	describe('async store init', () => {
```
gamemaker1, line 184: Can we add a test that checks the difference in handling of sync/async errors?

gamemaker1 (approved):
I think this is a good change! This is my understanding of what changes with this PR:

1. When the middleware is registered, it runs `store.init()`.
2. If `store.init()` throws a/an:
	a. synchronous error, it will be thrown and the server will crash. Thus synchronous errors should only be for programmer errors.
	b. asynchronous error, it will be caught and logged.
3. If a request is received, and `store.init()` threw an async error, the `increment`/`decrement` calls will either recover from the error and work (e.g. network error), or they will fail. The recovery must be handled by the store itself, either in the `init` method by adding a `catch` there itself, or by making the `increment`/`decrement` calls resilient (like rate-limit-redis does).

This is a change from the current behaviour where both 2a and 2b crash the server and point 3 is not applicable.

Note that in case someone defines the init function using the `async` keyword, all synchronous errors throw inside it will be treated as async errors, which means even they will be caught and logged. We should make this clear in the docs.

test/library/middleware-test.ts
nfriedly: Honestly, the more I think about it, the more I want to just handle all errors from .init() the same regardless of if they were sync/async.

nfriedly: done in 5115bd2 (both catching sync errors and a test to validate the behavior)

nfriedly:
Ok, updates:

* Synchronous errors are now caught and logged, the same as async errors
* New test for synchronous errors
* Updated the create-a-store documentation

source/types.ts
```diff
@@ -166,9 +166,16 @@ export type Store = {
 	 * Method that initializes the store, and has access to the options passed to
 	 * the middleware too.
 	 *
+	 * Called once during initialization.
+	 *
+	 * Async errors / promise rejections will be caught and logged.
+	 * (Synchronously thrown errors are not caught, and will prevent the express-rate-limit instance from being created.)
```
gamemaker1: We should remove this.

```suggestion
```

gamemaker1 (approved):
LGTM

PR #622, opened 2026-04-25
feat: add forensic debug logging support to Logger and middleware

Kofacts:
This PR implements forensic debug logging support in the core `express-rate-limit` library, as suggested by maintainers in `express-slow-down#90`.

Changes:
1. Added `debug: LoggerFn` to the `Logger` interface in `source/types.ts`.
2. Implemented the `debug` method in `ConsoleLogger` using `console.debug`.
3. Added strategic debug logging points across the middleware (`source/rate-limit.ts`) for:
   - Request skipping decisions.
   - Key generation results.
   - Store increment outcomes (hit counts and reset times).
   - Limit calculation results.

This provides maintainers and developers with deep visibility into the rate-limiting process, facilitating easier auditing and troubleshooting of anomalies.

Linked: express-rate-limit/express-slow-down#90


source/types.ts
```diff
@@ -101,7 +105,13 @@ export type IncrementResponse = ClientRateLimitInfo
...
  * A modified Express request handler with the rate limit functions.
  */
-export type RateLimitRequestHandler = RequestHandler & {
+export type RateLimitRequestHandler<
+	P = any,
+	ResBody = any,
+	ReqBody = any,
+	ReqQuery = any,
+	Locals extends Record<string, any> = any,
+> = RequestHandler<P, ResBody, ReqBody, ReqQuery, Locals> & {
```
gamemaker1: Why is this required?

test/typing-issue-614.ts
gamemaker1, line 1: I do not see how this is relevant to the PR.

package-lock.json
gamemaker1, line 1: Please delete this file. We use `pnpm` to manage dependencies.

source/rate-limit.ts
gamemaker1, line 1: How about also adding debug logs for decrementing hits and setting headers?

gamemaker1 (changes requested):

source/rate-limit.ts
```diff
@@ -386,6 +387,7 @@ const rateLimit = (
 
 			// Get a unique key for the client
 			const key = await config.keyGenerator(request, response)
+			config.logger.debug(undefined, `express-rate-limit: generated key "${key}" for ${request.method} ${request.url}`)
```
gamemaker1: Perhaps we can skip this log and the one below it and directly log the `RateLimitInfo` object below with the corresponding key?

Kofacts:
Appreciate the feedback. I've corrected the errors in judgment by removing `package-lock.json`, consolidating logs, and making `debug` optional for compatibility. Verified with `pnpm lint` and `test`.

source/rate-limit.ts
```diff
@@ -440,9 +444,17 @@ const rateLimit = (
 
 			// Set the rate limit information on the augmented request object
 			augmentedRequest[config.requestPropertyName] = info
+			config.logger.debug?.(
+				undefined,
+				`express-rate-limit: key="${key}", hits=${totalHits}, limit=${limit}, remaining=${info.remaining}, resetTime=${resetTime?.toISOString()}`,
```
gamemaker1: Why not just pass the info object as the first argument, instead of manually writing it out?

source/types.ts
gamemaker1: Please undo this change.

gamemaker1 (changes requested):

source/rate-limit.ts
```diff
@@ -377,6 +377,10 @@ const rateLimit = (
 			// First check if we should skip the request
 			const skip = await config.skip(request, response)
 			if (skip) {
+				config.logger.debug?.(
+					undefined,
+					`express-rate-limit: skipping request for ${request.method} ${request.url}`,
+				)
```
nfriedly: Should debug logs get the error first? I feel like that's going to always be undefined. @piotaixr & 
@acburdine - any thoughts?

acburdine: if it's always going to be undefined it's probably worth updating the types to not have the first argument at all for the `debug` function, or to maybe pass an info object as the first arg instead of an error object

piotaixr: The reason the error is first in warn and error, and the message second is because the message was not always provided (and to follow what pino and others are doing).

Maybe we can provide a context object like `{method: request.method, url: request.url}` ?

source/console-logger.ts
```diff
@@ -7,4 +7,7 @@ export const ConsoleLogger: Logger = {
 	error(...args) {
 		console.error(...args.reverse())
 	},
+	debug(...args) {
+		console.debug(...args.reverse())
+	},
```
piotaixr: If I understand correctly, the default behavior will be to print debug logs to the console.
Not 100% sure about this but according to [NodeJS documentation](https://nodejs.org/api/console.html#consoledebugdata-args), `console.debug` is an alias for `console.log`. It means that without an additional configuration option to enable debug logs, the default behavior will now be to print all debug information to the console.

Maybe a `DebugConsoleLogger` object can be exported by the library. This way the default remains to NOT log the debug information to the console, but users of the library can provide the `DebugConsoleLogger` as logger implementation if they want the debug log to be printed to the console (or their own logger implementation if they have one)

nfriedly: Oh, yeah, that's a good point. We don't want this turned on by default. I've used https://www.npmjs.com/package/debug before, it's off by default and can be fully or selectively enabled by an environment property.

source/types.ts
Kofacts: Done, reverted in the latest commit.

test/typing-issue-614.ts
Kofacts: Removed, that file and its changes were part of a separate branch that got mixed in by mistake.

package-lock.json
Kofacts: Deleted, that was accidentally committed. Using `pnpm` as expected.

source/rate-limit.ts
Kofacts: Added in the latest commit, debug logs now cover decrementing hits and all three header variants (legacy, draft-6, draft-7, draft-8).

source/rate-limit.ts
Kofacts: Good call, consolidated into a single `debug('express-rate-limit: rate limit info', info)` call passing the full `info` object.

source/rate-limit.ts
Kofacts: Good call, consolidated into a single `debug('express-rate-limit: rate limit info', info)` call passing the full `info` object.

source/rate-limit.ts
Kofacts: Agreed. Introduced a dedicated `DebugFn` type with signature `(message: string, context?: unknown)`, no error-first argument. All debug call sites updated accordingly.

Kofacts: Great suggestion, all debug call sites now pass a structured context object (e.g. `{ key }`, `{ method, url }`, or the full `info` object) as the second argument.

source/console-logger.ts
Kofacts: Great catch. Removed `debug` from the default `ConsoleLogger` so it stays silent by default. Exported a new `DebugConsoleLogger` that users can opt into if they want console debug output. This keeps the default behavior unchanged.

nfriedly (commented):
Ok, I think the implementation of the DebugConcoleLogger is in a much better state now, thank you for all of the changes @Kofacts.

Before we ship it, though I think we need to pause and consider how the feature is going to be used and weather or not the specific logs here are the right ones. 

If I were to envision this feature, it would probably be a single line per request, which would usually be logged toward the end, right before `config.handler()` is called for requests over the limit, or right before `next()` is called for requests that are passed. (Maybe restructure that section so that it was a single debug call.)

However, if `skipSuccessfulRequests`/`skipFailedRequests` was set, then we should instead log after that was resolved.

We might want to add any data needed for the log to the `request.rateLimit` object so that the log line could be built from that, and then library consumers would also have it available.

So maybe the log looks something like this:

```ts
config.logger.debug?.('express-rate-limit:', augmentedRequest[config.requestPropertyName])
```

(Maybe we'd also want to include `${augmentedRequest.method} ${augmentedRequest.path}` in the string to help identify the request.)

And then the rate limit info (augmentedRequest[config.requestPropertyName]) would get new fields for `skipped` (maybe an enum value to differentiate between the skip function, skipFailedRequests, & skipSuccessfulRequests), a boolean `blocked` (or perhaps `handled` or `delayed` for express-slow-down), an array of headers set, and any other internal information we might find useful. 

And then express-slow-down could extend it with the delay length so that it gets included in the log. We might have to add a hook so that ESD could set the delay length before ERL calls the handler function, I need to think about that a bit more.

I don't know, as I'm writing this out, it seems like a lot of work for unknown value. Maybe it should just be an example in the docs rather than a built-in feature. What do you think @gamemaker1 ?

gamemaker1:
I thought about this a bit, and here is what I think:

1. A lot of the complexity here as well as things that would make this feature more useful like respecting environment variables are all implemented in [`debug`](https://www.npmjs.com/package/debug). I do not think we should reinvent the wheel. 
2. Express itself, along with some of its popular middlewares, already use [`debug`](https://www.npmjs.com/package/debug) for debug level logging. I believe we should follow this as it will make it easier for developers debugging an express app to configure and view debug logs if they are all in one place rather than worry about different `DebugConsoleLogger` interfaces and implementations.

As for the "one log line", we could have one log "block" instead, the way express does it:

```
> DEBUG='*' bun run index.ts

  express:application set "x-powered-by" to true +0ms
  express:application set "etag" to 'weak' +1ms
  express:application set "etag fn" to [Function: generateETag] +0ms
  express:application set "env" to 'development' +2ms
  express:application set "query parser" to 'simple' +0ms
  express:application set "query parser fn" to [Function: parse] +0ms
  express:application set "subdomain offset" to 2 +0ms
  express:application set "trust proxy" to false +0ms
  express:application set "trust proxy fn" to [Function: trustNone] +1ms
  express:application booting in development mode +1ms
  express:application set "view" to [Function: View] +0ms
  express:application set "views" to '/[...]/views' +0ms
  express:application set "jsonp callback name" to 'callback' +0ms
  router use '/' <anonymous> +0ms
  router:layer new '/' +0ms
  router:route new '/' +0ms
  router:layer new '/' +1ms
  router:route get / +0ms
  router:layer new '/' +0ms
  router dispatching GET / +4s
  router <anonymous>  : / +2ms

[...]
```

We could scope it to express-rate-limit:handler and add each attribute as a separate line.

If this sounds good, let me know, I can code it up and push the changes to this PR itself for review.

nfriedly:
Oh, I forgot express was using debug, that's a good idea. And that's a good point that express itself does a ton of individual log lines for each thing, so I suppose that is the pattern to follow. 

We could (probably?) stick debug in the default `ConsoleLogger` and drop the `DebugConsoleLogger` since it's off by default.

Perhaps we should make an example of extending the default `ConsoleLogger` so that errors and warnings go to your custom log system but the debug logs keep the same behavior. (Or just do that automatically internally?)

gamemaker1:
> We could (probably?) stick debug in the default ConsoleLogger and drop the DebugConsoleLogger since it's off by default.

> Perhaps we should make an example of extending the default ConsoleLogger so that errors and warnings go to your custom log system but the debug logs keep the same behavior. (Or just do that automatically internally?)

I think we should just drop the `debug` function from our Logger interface entirely and handle it with the debug library, since the debug logs are likely only going to be used for debugging and not in production environments where they would need to be bundled with the rest of the app's errors and warnings.

PR #625, opened 2026-05-06
fix: bump ip-address to ^10.1.1 to resolve XSS vulnerability (GHSA-v2v4-37r5-5v8g)

### LINKED ISSUE #624: Use of vulnerable version of ip-address
### Description

This library uses `ip-address@10.1.0` which has a vulnerability.

It should be updated to use `^10.1.1` to resolve this and ensure that should a similar issue arise in future it can be resolved by `npm audit fix` without needing an override or an update to this package.

### Library version

8.5.0

### Express version

5.2.1

### Node version

22.22.1

^5.8.3

### Module system

CommonJS

carrotRakko:
## Summary

Bump `ip-address` from a pinned `10.1.0` to a caret range `^10.1.1` to resolve [GHSA-v2v4-37r5-5v8g](https://github.com/advisories/GHSA-v2v4-37r5-5v8g) (XSS in Address6 HTML-emitting methods, severity: medium).

## Why a caret range instead of a fixed bump

The current pin (`"ip-address": "10.1.0"`) prevents downstream consumers from picking up patch-level fixes via `npm audit fix` / `pnpm install`. Downstream maintainers currently need to apply a local override just to get the fix.

Switching to `^10.1.1`:

- pulls in the immediate patch (10.1.1+)
- keeps future patch and minor updates flowing automatically (semver-compatible)
- lets future similar issues be resolved by consumers without an override or a new release of this package

## Closes

#624

## Test plan

- `pnpm install` resolves to a non-vulnerable `ip-address` version
- No source changes; only a dependency range adjustment

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,7 +75,7 @@
 		"prepare": "run-s compile && husky"
 	},
 	"dependencies": {
-		"ip-address": "10.1.0"
+		"ip-address": "^10.1.1"
 	},
 	"peerDependencies": {
 		"express": ">= 4.11"


nfriedly:
Okay, it looks like the lock file needs to get updated to match the package.json, but otherwise this is good. 

Ip-address recently added a new [network form feature](https://github.com/beaugunderson/ip-address/pull/199) at my request that I was planning on using, but I didn't realize there was a security issue. Let's just deal with the security issue for right now.

nfriedly:
I went ahead and put up https://github.com/express-rate-limit/express-rate-limit/pull/626 with the corrected lockfile, and also the latest version of ip-address.

PR #633, opened 2026-05-13
feat: simplify IPv6 key generation

nfriedly:
This takes advantage of a [new feature that the ip-address library added](https://github.com/beaugunderson/ip-address/pull/199) at my request to reduce the amount of string templating we need to do.


gamemaker1 (approved):
LGTM.

Is the logger external test's lockfile not updated to use the "../../.." path to the library code? It seems to be failing with the error "[..] networkForm() is not a function".

PR #637, opened 2026-05-26
Chore: update tsconfig to use nodenext instead of node

nfriedly:
This should fix the error that broke https://github.com/express-rate-limit/express-rate-limit/pull/635


nfriedly:
Ugh, this is going to need some more work.

nfriedly:
Ok, that was a bit of a pain, but it's all working now

gamemaker1 (approved):

gamemaker1:
Thank you so much!

PR #639, opened 2026-06-08
test: fix response closes skipFailedRequests (server) test

nfriedly:
The `should decrement hits when response closes and 'skipFailedRequests' is set to true (%s store) (server)` test started [failing](https://github.com/express-rate-limit/express-rate-limit/actions/runs/26865003421/job/79581145692?pr=638) in node.js v26.3.0. I poked at it a bit and figured out that the supertest `.timeout(n)` method behaves differently in v26.3 - previously it would initiate the request and wait `n` milliseconds for the request to complete before aborting it. But now it aborts immediately, without even sending the request to the server.  (And, if you wait to call `.timeout(n)`, then it waits to initiate the request.) I filed a bug at https://github.com/forwardemail/supertest/issues/891 but didn't get to the bottom of it.

For our needs, I rewrote the test to spin up an actual server and use http.get to connect to it, then manually abort the request after the server receives it. This version of the test passes on all versions of node.js. 

What we're trying to test is the server-side behavior after a connection is aborted, so it doesn't really matter how the abort happens. (It probably wasn't a great fit for supertest to begin with, because that's more concerned with the client's perspective, so it doesn't care what happens after a request ends.)

It's kind of an awkward mix of promises and callbacks right now. We could probably clean that up, but I've spent a few evenings banging my head into this test already and I'm just ready to be done with it for now.

---

On an unrelated note, I kept seeing this line get logged:

> ts-jest[config] (WARN) message TS151002: Using hybrid module kind (Node16/18/Next) is only supported in "isolatedModules: true". Please set "isolatedModules: true" in your tsconfig.json. To disable this message, you can set "diagnostics.ignoreCodes" to include 151002 in your ts-jest config. See more at https://kulshekhar.github.io/ts-jest/docs/getting-started/options/diagnostics

But, when I tried setting `isolatedModules: true`, it starts treating the tests as commonjs for some reason and fails on the first `import` statement in each file

### DIFF
--- a/test/library/middleware-test.ts
+++ b/test/library/middleware-test.ts
@@ -4,7 +4,6 @@
 import { EventEmitter } from 'node:events'
 import HTTP from 'node:http'
 import type { AddressInfo } from 'node:net'
-import { platform } from 'node:process'
 
 import {
 	afterEach,
--- a/test/library/middleware-test.ts
+++ b/test/library/middleware-test.ts
@@ -2,6 +2,8 @@
 // Tests the rate limiting middleware
 
 import { EventEmitter } from 'node:events'
+import HTTP from 'node:http'
+import type { AddressInfo } from 'node:net'
 import { platform } from 'node:process'
 
 import {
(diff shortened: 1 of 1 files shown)


test/library/middleware-test.ts
gamemaker1: Do we still need to skip it on MacOS?

gamemaker1 (approved):
LGTM! Thank you so much, this looks complicated but correct.

test/library/middleware-test.ts
nfriedly: oh, probably not. let's try it and see...

nfriedly: Ok, it took three tries, but yes, it does work in macOS now :tada:

PR #641, opened 2026-06-09
feat: use `debug` for debug logging

gamemaker1:
Supersedes & closes #622. With this, the logs look like this with `DEBUG='*'` set.

```
  express-rate-limit creating new rate limiter +0ms
  express-rate-limit set window 60000 +0ms
  express-rate-limit init for store 'MemoryStore' +1ms
  router use '/' <anonymous> +0ms
  router:layer new '/' +0ms
  router:route new '/' +0ms
  router:layer new '/' +1ms
  router:route get / +0ms
  router:layer new '/' +0ms
  router dispatching GET / +4s
  router <anonymous>  : / +1ms
  express-rate-limit request from url '/' +4s
  express-rate-limit request from ip '::ffff:127.0.0.1' +1ms
  express-rate-limit computed key '127.0.0.1' +8ms
  express-rate-limit incrementing count +1ms
  express-rate-limit set used to 1 +0ms
  express-rate-limit set remaining to 1 +0ms
  express-rate-limit set resetTime to 2026-06-09T13:53:38.269Z +0ms
  express-rate-limit set ietf draft 6 headers +1ms
```

I'm not sure if each rate limiter instance should get its own log scope, please let me know what you think.

### DIFF
--- a/source/rate-limit.ts
+++ b/source/rate-limit.ts
@@ -2,6 +2,7 @@
 // The option parser and rate limiting middleware
 
 import { isIPv6 } from 'node:net'
+import createDebugLogger from 'debug'
 import type { NextFunction, Request, RequestHandler, Response } from 'express'
 import { ConsoleLogger } from './console-logger.js'
 import {
@@ -349,6 +350,12 @@ const rateLimit = (
 	const config = parseOptions(passedOptions ?? {})
 	const options = getOptionsFromConfig(config)
 
+	// Create a debug logger for this handler
+	const debug = createDebugLogger('express-rate-limit')
+	debug('creating new rate limiter')
+	debug('set window to %o', config.windowMs)
+	debug('set limit to %o', config.limit)
+
 	// The limiter shouldn't be created in response to a request (usually)
 	config.validations.creationStack(config.store)
 	// The store instance shouldn't be shared across multiple limiters
@@ -356,6 +363,8 @@ const rateLimit = (
 
 	// Call the `init` method on the store, if it exists
 	if (typeof config.store.init === 'function') {
+		debug('init for store %o', config.store.constructor.name)
+
 		// If store.init() throws or rejects, we'll catch and log it
 		// Use .catch() rather than await, because we need to return synchronously
 		try {
@@ -393,9 +402,13 @@ const rateLimit = (
 				config.skipFailedRequests &&
 				new Promise<void>((resolve) => response.once('error', resolve))
 
+			debug('request from url %o', request.originalUrl)
+			debug('request from ip %o', request.ip)
+
 			// First check if we should skip the request
 			const skip = await config.skip(request, response)
 			if (skip) {
+				debug('skipping request')
 				next()
 				return
 			}
@@ -405,8 +418,10 @@ const rateLimit = (
 
 			// Get a unique key for the client
 			const key = await config.keyGenerator(request, response)
+			debug('computed key %o', key)
 
 			// Increment the client's hit counter by one.
+			debug('incrementing count')
 			let totalHits = 0
 			let resetTime
 			try {
@@ -449,6 +464,10 @@ const rateLimit = (
 				key,
 			}
 
+			debug('set used to %o', info.used)
+			debug('set remaining to %o', info.remaining)
+			debug('set resetTime to %o', info.resetTime)
+
 			// Set the `current` property on the object, but hide it from iteration
 			// and `JSON.stringify`. See the `./types#RateLimitInfo` for details.
 			Object.defineProperty(info, 'current', {
@@ -462,6 +481,7 @@ const rateLimit = (
 
 			// Set the `X-RateLimit` headers on the response object if enabled.
 			if (config.legacyHeaders && !response.headersSent) {
+				debug('set legacy headers')
 				setLegacyHeaders(response, info)
 			}
 
@@ -470,11 +490,13 @@ const rateLimit = (
 			if (config.standardHeaders && !response.headersSent) {
 				switch (config.standardHeaders) {
 					case 'draft-6': {
+						debug('set ietf draft 6 headers')
 						setDraft6Headers(response, info, config.windowMs)
 						break
 					}
 
 					case 'draft-7': {
+						debug('set ietf draft 7 headers')
 						config.validations.headersResetTime(info.resetTime)
 						setDraft7Headers(response, info, config.windowMs)
 						break
@@ -487,6 +509,8 @@ const rateLimit = (
 								: config.identifier
 						const name = await retrieveName
 
+						debug('set ietf draft 7 headers')
+						debug('set name to %o', name)
 						config.validations.headersResetTime(info.resetTime)
 						setDraft8Headers(response, info, config.windowMs, name, key)
 						break
@@ -512,6 +536,8 @@ const rateLimit = (
 						if (resetTime && Date.now() >= resetTime.getTime()) {
 							return
 						}
+
+						debug('decrementing count')
 						await config.store.decrement(key)
 						decremented = true
 					}
@@ -555,7 +581,9 @@ const rateLimit = (
 			// If the client has exceeded their rate limit, set the Retry-After header
 			// and call the `handler` function.
 			if (totalHits > limit) {
+				debug('limit exceeded')
 				if (config.legacyHeaders || config.standardHeaders) {
+					debug('set retry-after header')
 					setRetryAfterHeader(response, info, config.windowMs)
 				}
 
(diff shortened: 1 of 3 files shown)


source/rate-limit.ts
```diff
@@ -349,13 +350,21 @@ const rateLimit = (
 	const config = parseOptions(passedOptions ?? {})
 	const options = getOptionsFromConfig(config)
 
+	// Create a debug logger for this handler
+	const debug = createDebugLogger('express-rate-limit')
+	debug('creating new rate limiter')
+	debug('set window to %o', config.windowMs)
+	debug('set limit to %o', config.limit)
```
nfriedly: perhaps we should log all the not-undefined options here? 

Also, anything that's a function has to be run in response to a request (including the limit), so maybe we need that in the per-request logging?

source/rate-limit.ts
```diff
@@ -393,9 +402,13 @@ const rateLimit = (
 				config.skipFailedRequests &&
 				new Promise<void>((resolve) => response.once('error', resolve))
 
+			debug('request from url %o', request.originalUrl)
```
nfriedly: This is basically a duplicate of the `router dispatching GET / ` line - do you think we need it?

source/rate-limit.ts
```diff
@@ -487,6 +509,8 @@ const rateLimit = (
 								: config.identifier
 						const name = await retrieveName
 
+						debug('set ietf draft 7 headers')
```
nfriedly:
```suggestion
						debug('set ietf draft 8 headers')
```

nfriedly (approved):
Nice work!

I have a few thoughts below, but other than the draft 7/8 thing, we could ship this as-is

nfriedly:
Oh, and you'll want to rebase off of main to get the test fix I just merged in

source/rate-limit.ts
gamemaker1: Ah okay sure, will do.

source/rate-limit.ts
gamemaker1: I wanted to keep it in case someone disables the express logs and keeps only the express-rate-limit logs.

source/rate-limit.ts
```diff
@@ -349,13 +351,34 @@ const rateLimit = (
 	const config = parseOptions(passedOptions ?? {})
 	const options = getOptionsFromConfig(config)
 
+	// Create a debug logger for this handler
+	const debug = createDebugLogger('express-rate-limit')
+	debug('creating new rate limiter with %o', config.store.constructor.name)
+
+	// Log the options that are unique to this isntance of the middleware.
+	const optionsToLog = [
```
gamemaker1: I do not like these lists we are creating in different places but I am not sure what a better way is to do this. Maybe I can loop over the options and only print those that are not functions?

source/rate-limit.ts
gamemaker1: The per-request functions that are run are logged just after they are run in the middleware. I didn't want to log them separately by running the function twice in case that breaks something.

gamemaker1:
The logs look like this now:

```
  express-rate-limit set standardHeaders to 'draft-6' +0ms
  express-rate-limit set validations to { default: true } +0ms
  express-rate-limit executing init for store +2ms
  router use '/' <anonymous> +0ms
  router:layer new '/' +0ms
  router:route new '/' +0ms
  router:layer new '/' +1ms
  router:route get / +1ms
  router:layer new '/' +1ms
  router dispatching GET / +2s
  router <anonymous>  : / +0ms
  express-rate-limit requested '/' +2s
  express-rate-limit request from ip '::ffff:127.0.0.1' +1ms
  express-rate-limit computed key '127.0.0.1' +9ms
  express-rate-limit incrementing count +0ms
  express-rate-limit computed limit to be 2 +0ms
  express-rate-limit computed used to be 1 +0ms
  express-rate-limit computed remaining to be 1 +0ms
  express-rate-limit computed resetTime to be 2026-06-14T05:03:50.759Z +0ms
  express-rate-limit computed key to be '127.0.0.1' +1ms
  express-rate-limit set ietf draft 6 headers +0ms
```

source/rate-limit.ts
nfriedly: Yeah, I was imagining a loop, not a hard-coded list. I'm a little bit divided on weather or not to log the options that are default values, I was thinking maybe looping over `notUndefinedOptions`.

If we wanted to do a list so that we're not logging a `set ...` for fields the library doesn't recognize, then maybe it could be shared with the knownOptions validation?

source/rate-limit.ts
```diff
@@ -487,6 +523,8 @@ const rateLimit = (
 								: config.identifier
 						const name = await retrieveName
 
+						debug('set ietf draft 8 headers')
+						debug('set name to %o', name)
```
nfriedly, line 514: I think they call it the identifier in the standard drafts
```suggestion
						debug('set policy identifier to %o', name)
```

source/rate-limit.ts
gamemaker1: I was thinking about this, and came up with the following:

1. We should log default values; I think the logs for each instance/request should show the complete configuration at that instant for better debug-ability.
2. Can we have one big object where we define each type, and give it sub-keys like 'deprecated' or 'removed' or 'loggable' or something? We can then export the `Options` type as the set of keys of this big object that are not 'removed', and log the ones that are marked 'loggable'. Something like this (untested):

```ts
const OPTIONS = {
    // ... jsdoc
    windowMs: { loggable: true },
    draft_polli_headers: { loggable: false, deprecated: true },

    // and so on...
}

export type Options = keyof OPTIONS // where removed not true, somehow
```

nfriedly: Yeah, that could work although I don't love how complex this is getting.

We have the validation to check for unrecognized keys, so maybe this should just log whatever is in options without caring about whether or not it's a recognized key. Then it could just be a simple loop that doesn't need to be updated any time we change the configuration available.

(Also, on an unrelated note, given how much config-related code we have now, maybe we should move it to it's own config.ts or whatever?) (Not as part of this PR, though.)

gamemaker1: Replaced it with a simple loop. It generates a noisy line for validations because its a large JSON object, but it's fine for now.

I will work on separating the config code soon.

source/rate-limit.ts
```diff
@@ -98,7 +99,8 @@ const promisifyStore = (passedStore: LegacyStore | Store): Store => {
  * The internal configuration interface.
  *
  * This is copied from Options, with fields made non-readonly and deprecated
- * fields removed.
+ * fields removed. Whenever this is updated, please update the debug logs in
+ * the main middleware function as well.
```
nfriedly: We might not need this change now

source/rate-limit.ts
```diff
@@ -349,13 +351,23 @@ const rateLimit = (
 	const config = parseOptions(passedOptions ?? {})
 	const options = getOptionsFromConfig(config)
 
+	// Create a debug logger for this handler
+	const debug = createDebugLogger('express-rate-limit')
+	debug('creating new rate limiter with %o', config.store.constructor.name)
+
+	// Log the parsed options for this middleware instance.
+	for (const [key, val] of Object.entries(config))
+		debug('set %s to %o', key, val)
```
nfriedly, line 359: Yeah, I like the simplicity here. 

Does the log output look reasonable if a function is passed in?

nfriedly (approved):
A couple of minor things, but I think we're pretty much there.

source/rate-limit.ts
gamemaker1: Yes. This is what it looks like now:

```
  express-rate-limit set ipv6Subnet to 56 +1ms
  express-rate-limit set standardHeaders to 'draft-6' +0ms
  express-rate-limit set store to MemoryStore { validations: undefined, previous: Map(0) {}, current: Map(0) {}, localKeys: true } +0ms
  express-rate-limit set validations to { enabled: { default: true }, disable: [Function (anonymous)], ip: [Function (anonymous)], trustProxy: [Function (anonymous)], xForwardedForHeader: [Function (anonymous)], forwardedHeader: [Function (anonymous)], positiveHits: [Function (anonymous)], unsharedStore: [Function (anonymous)], singleCount: [Function (anonymous)], limit: [Function (anonymous)], draftPolliHeaders: [Function (anonymous)], onLimitReached: [Function (anonymous)], headersDraftVersion: [Function (anonymous)], headersResetTime: [Function (anonymous)], knownOptions: [Function (anonymous)], validationsConfig: [Function (anonymous)], creationStack: [Function (anonymous)], ipv6Subnet: [Function (anonymous)], ipv6SubnetOrKeyGenerator: [Function (anonymous)], keyGeneratorIpFallback: [Function (anonymous)], windowMs: [Function (anonymous)] } +0ms
  express-rate-limit set logger to { warn: [Function: warn], error: [Function: error] } +1ms
  express-rate-limit executing init for store +0ms
```

PR #647, opened 2026-06-23
fix: don't decrement if resetTime has passed

DanielOliverRJ:
We track use `express-rate-limit` and `rate-limit-redis` to apply a request concurrency limit, using the `skip*` options to decrement the number of “in flight” operations when a requests completes.  Within `windowMs`, the mechanism works as expected.  However, if a request starts in one window period and finishes after the counter is reset, the number of tracked “in-flight” requests can become negative.  The result is that more requests are allowed than expected.

One solution would be to set the Redis store `resetExpiryOnChange` parameter, but this would introduce the risk that a process crash never decrement the counter, leading to clients being unable to make requests within their allotment.

Another solution is to change the behaviour of `rate-limit-redis` to never allow counter to drop below 0, but this is difficult as we don’t know if the request in question is one which has crossed the reset boundary.

The best approach I could identify was to update the behaviour of the `decrement()` call so that the counter remains unchanged if the completion time has progressed beyond that of when the counter would be reset.

However, this does introduce the risk that if the user configured `resetExpiryOnChange`, a decrement is not performed when it it would in fact be valid.  I was unsure how you would prefer to resolve the conflict, given the that stores may handle this differently.  Is changing the store interface ok, or should the change in decrement behaviour be configurable?

### DIFF
--- a/test/library/middleware-test.ts
+++ b/test/library/middleware-test.ts
@@ -569,8 +569,6 @@ describe('middleware test', () => {
 	})
 
 	it('should decrement hits when request finishes before `resetTime` with `skipSuccessfulRequests`', async () => {
-		jest.useRealTimers()
-
 		const windowMs = 500
 		const requestDurationMs = 25
 		const store = new MockStore()
@@ -583,6 +581,7 @@ describe('middleware test', () => {
 			}),
 			(_request, _response, next) => {
 				setTimeout(next, requestDurationMs)
+				jest.runAllTimers()
 			},
 		])
 
@@ -592,8 +591,6 @@ describe('middleware test', () => {
 	})
 
 	it('should not decrement hits when request finishes after `resetTime` with `skipSuccessfulRequests`', async () => {
-		jest.useRealTimers()
-
 		const windowMs = 50
 		const requestDurationMs = 75
 		const store = new MockStore()
@@ -606,6 +603,7 @@ describe('middleware test', () => {
 			}),
 			(_request, _response, next) => {
 				setTimeout(next, requestDurationMs)
+				jest.runAllTimers()
 			},
 		])
 
--- a/test/library/middleware-test.ts
+++ b/test/library/middleware-test.ts
@@ -43,11 +43,11 @@ describe('middleware test', () => {
 		resetAllWasCalled = false
 
 		counter = 0
+		windowMs = 0
 
-		constructor(private readonly resetInMS?: number) {}
-
-		init(_options: Options): void {
+		init(options: Options): void {
 			this.initWasCalled = true
+			this.windowMs = options.windowMs
 		}
 
 		async get(_key: string): Promise<ClientRateLimitInfo> {
@@ -62,10 +62,7 @@ describe('middleware test', () => {
 
 			return {
 				totalHits: this.counter,
-				resetTime:
-					typeof this.resetInMS === 'number'
-						? new Date(Date.now() + this.resetInMS)
-						: undefined,
+				resetTime: new Date(Date.now() + this.windowMs),
 			}
 		}
 
@@ -574,9 +571,9 @@ describe('middleware test', () => {
 	it('should decrement hits when request finishes before `resetTime` with `skipSuccessfulRequests`', async () => {
 		jest.useRealTimers()
 
-		const windowMs = 50
+		const windowMs = 500
 		const requestDurationMs = 25
-		const store = new MockStore(windowMs)
+		const store = new MockStore()
 
 		const app = createServer([
 			rateLimit({
@@ -599,7 +596,7 @@ describe('middleware test', () => {
 
 		const windowMs = 50
 		const requestDurationMs = 75
-		const store = new MockStore(windowMs)
+		const store = new MockStore()
 
 		const app = createServer([
 			rateLimit({
--- a/test/library/middleware-test.ts
+++ b/test/library/middleware-test.ts
@@ -44,6 +44,8 @@ describe('middleware test', () => {
 
 		counter = 0
 
+		constructor(private readonly resetInMS?: number) {}
+
 		init(_options: Options): void {
 			this.initWasCalled = true
 		}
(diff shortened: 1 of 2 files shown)


test/library/middleware-test.ts
```diff
@@ -44,6 +44,8 @@ describe('middleware test', () => {
 
 		counter = 0
 
+		constructor(private readonly resetInMS?: number) {}
+
 		init(_options: Options): void {
 			this.initWasCalled = true
 		}
```
nfriedly: Instead of passing it in the constructor, we can read `windowMs` from options in the `init()` method (which is called automatically when the limiter is created):


```suggestion
		windowMs = 0
	
		init(options: Options): void {
			this.initWasCalled = true
			this.windowMs = options.windowMs
		}
```

test/library/middleware-test.ts
```diff
@@ -58,7 +60,13 @@ describe('middleware test', () => {
 			this.counter += 1
 			this.incrementWasCalled = true
 
-			return { totalHits: this.counter, resetTime: undefined }
+			return {
+				totalHits: this.counter,
+				resetTime:
+					typeof this.resetInMS === 'number'
+						? new Date(Date.now() + this.resetInMS)
+						: undefined,
```
nfriedly: Then we can simplify this because the limiter will provide a default windowMs if it wasn't specified in the config.

```suggestion
				resetTime: new Date(Date.now() + this.windowMs),
```

test/library/middleware-test.ts
```diff
@@ -563,6 +571,52 @@ describe('middleware test', () => {
 		expect(store.decrementWasCalled).toEqual(true)
 	})
 
+	it('should decrement hits when request finishes before `resetTime` with `skipSuccessfulRequests`', async () => {
+		jest.useRealTimers()
+
+		const windowMs = 50
+		const requestDurationMs = 25
+		const store = new MockStore(windowMs)
```
nfriedly:
```suggestion
		const store = new MockStore()
```

test/library/middleware-test.ts
```diff
@@ -563,6 +571,52 @@ describe('middleware test', () => {
...
+
+		expect(store.decrementWasCalled).toEqual(true)
+	})
+
+	it('should not decrement hits when request finishes after `resetTime` with `skipSuccessfulRequests`', async () => {
+		jest.useRealTimers()
+
+		const windowMs = 50
+		const requestDurationMs = 75
+		const store = new MockStore(windowMs)
```
nfriedly:
```suggestion
		const store = new MockStore()
```

nfriedly (approved):
Hey, thanks!

I'd like to consider it for a day or so, but this does seem like it's the best solution at first glance.

As far as `resetExpiryOnChange` goes, [I was already thinking about getting rid of it](https://github.com/express-rate-limit/rate-limit-redis/issues/220), as I don't think the feature makes that much sense in the context of a rate-limiter, so this just adds another reason.

I have a suggestion for the tests (I'm not 100% sure it will pass the linter, but it should give you the right idea). That said, the code change itself looks good to me, so I'm going to go ahead and approve this now.

I'd also appreciate @gamemaker1's thoughts on this.

DanielOliverRJ:
Thanks for the feedback @nfriedly.  The tests are indeed cleaner, I looked at using `init()` before, but convinced myself it wasn't being called.

I've also increased the timer on one test given the Windows failure on the previous run.

nfriedly:
I'm not quite sure why the tests are failing now, it looks like it's only on the latest node.js release, not on the LTS version, so perhaps it was a recent breaking change in node.js. 

(I also see all the ts-jest warnings, but the trouble is that if we do what it says, then a bunch of other things break and I haven't had time to sort it out.)

It might be worth dropping the `jest.useRealTimers()` and instead calling `jest.advanceTimersByTime(75)` or `jest.runAllTimers()` - it might sidestep the setImmediate problem.

DanielOliverRJ:
> It might be worth dropping the `jest.useRealTimers()` and instead calling `jest.advanceTimersByTime(75)` or `jest.runAllTimers()` - it might sidestep the setImmediate problem.

I started with `advanceTimersByTime`, but couldn't get the test to complete.  I'll take another look on both node versions.

test/library/middleware-test.ts
```diff
@@ -563,6 +568,52 @@ describe('middleware test', () => {
 		expect(store.decrementWasCalled).toEqual(true)
 	})
 
+	it('should decrement hits when request finishes before `resetTime` with `skipSuccessfulRequests`', async () => {
+		jest.useRealTimers()
```
nfriedly: This is incredibly dumb, but it might work:


```suggestion
		jest.useRealTimers()
		global.setImmediate ||= setTimeout // It should exist already 🤷
```

nfriedly: The failure might have the same root cause at https://github.com/forwardemail/supertest/issues/891 but I never got to the bottom of that one, I just rewrote the test to use an actual http server instead of going through supertest.

DanielOliverRJ:
Google to the rescue, it turns out the timer needed triggering.  Works for me on node 22 and 26, but I don't seem to have the powers to re-run the tests.

nfriedly:
Sorry, we have it set up so that a maintainer has to approve the tests before they run. There's probably a smarter way to do it where just the publishing credentials are restricted and the main tests can run for anyone, but I just approved them for now.

test/library/middleware-test.ts
```diff
@@ -563,6 +568,50 @@ describe('middleware test', () => {
...
+
+		const app = createServer([
+			rateLimit({
+				skipSuccessfulRequests: true,
+				windowMs,
+				store,
+			}),
+			(_request, _response, next) => {
+				setTimeout(next, requestDurationMs)
+				jest.runAllTimers()
```
nfriedly, line 584: aah, nice.

nfriedly (approved):

nfriedly:
🎉 all tests passed 🎉

gamemaker1 (approved):
LGTM! Thank you for the contribution @DanielOliverRJ :)

@nfriedly shall we coordinate the release of this with a major version bump of `rate-limit-redis` which removes the `resetExpiryOnChange` option?

nfriedly:
Yeah, that sounds good. I'll go ahead and merge this in now and we'll get a release out shortly.

nfriedly:
Sorry it took longer than I anticipated, but this is now published to npm in v8.6.0 - https://www.npmjs.com/package/express-rate-limit

DanielOliverRJ:
Thanks!  It's always nice to avoid local patches 😄

PR #650, opened 2026-06-30
feat: Run validations once each

### LINKED ISSUE #648: Dynamic `ipv6Subnet` validation is skipped when the first request is IPv4
## Summary

When `ipv6Subnet` is provided as a function, invalid return values appear to be validated only if an IPv6 request reaches the middleware before validations are disabled. If the first processed request is IPv4, later IPv6 requests can use an invalid dynamic subnet such as `65` without logging `ERR_ERL_IPV6_SUBNET`.

## Steps to reproduce

```ts
app.set('trust proxy', 1)

app.use(rateLimit({
  limit: 100,
  logger,
  store,
  ipv6Subnet: () => 65,
}))

await request(app)
  .get('/')
  .set('X-Forwarded-For', '203.0.113.7')

await request(app)
  .get('/')
  .set('X-Forwarded-For', '2001:db8::1')
```

## Expected behavior

The second request should trigger `ERR_ERL_IPV6_SUBNET`, because `65` is outside the valid IPv6 subnet range.

This is consistent with the control case where the first request is IPv6: `ERR_ERL_IPV6_SUBNET` is logged.

## Actual behavior

No validation error is logged after the IPv4-first request path. The later IPv6 request still generates a key using the invalid subnet:

```json
{
  "expectedErrorCode": "ERR_ERL_IPV6_SUBNET",
  "observedErrorCodes": [],
  "observedKeys": [
    "203.0.113.7",
    "2001:db8::/65"
  ]
}
```

Control case, with IPv6 first:

```json
{
  "observedErrorCodes": [
    "ERR_ERL_IPV6_SUBNET"
  ],
  "observedKeys": [
    "2001:db8::/65"
  ]
}
```

## Evidence

The apparent lifecycle gap is:

- `ipv6Subnet` can be a per-request function returning a number.
- Static subnet values are validated during option parsing, while function return values are deferred.
- The function value is only produced on the IPv6 key-generation path.
- Validations are globally disabled after the first processed request.
- If that first request is IPv4, the dynamic IPv6 subnet value has not yet been produced or checked.
- A later IPv6 request then uses `/65` without logging `ERR_ERL_IPV6_SUBNET`.

Relevant source anchors from source inspection include:

- `source/types.ts`: `ipv6Subnet` accepts a value-determining middleware.
- `source/rate-limit.ts`: function values skip parse-time subnet validation, and dynamic subnet values are later produced and validated.
- `source/rate-limit.ts`: validations are disabled after the first request.
- `source/validations.ts`: values greater than `64` should trigger `ERR_ERL_IPV6_SUBNET`.
- `source/ip-key-generator.ts`: the subnet is used to form the IPv6 key.

## Suggested fix

Avoid disabling the dynamic `ipv6Subnet` validation until a function-produced subnet value has actually been observed and checked, or track this validation separately from the global first-request validation disable.

A focused regression test could cover:

1. `ipv6Subnet: () => 65`
2. first request IPv4
3. second request IPv6
4. configured logger receives `ERR_ERL_IPV6_SUBNET`

The existing IPv6-first behavior should remain unchanged.

## Additional context / Related coverage

A current GitHub search did not identify exact existing issue or PR coverage for this IPv4-first dynamic `ipv6Subnet` path. Nearby IPv6/rate-limit configuration discussions exist, but they do not appear to cover this specific validation lifecycle case.

---
Submitted with Codex.

nfriedly:
This tweaks the validations disable logic from stop-all-after-first-request to stop-each-individual-check-after-a-single-run.

The logic change was fairly straightforward, the tests were a bit more work, but mostly because I had been a bit sloppy when writing them initially, so they are better tests now.

Fixes #648

### DIFF
--- a/test/library/validation-test.ts
+++ b/test/library/validation-test.ts
@@ -5,7 +5,7 @@ import { beforeEach, describe, expect, it, jest } from '@jest/globals'
 import express from 'express'
 import supertest from 'supertest'
 import { ipKeyGenerator, MemoryStore } from '../../source/index.js'
-import type { Logger, Store } from '../../source/types'
+import type { Logger, Options, Store } from '../../source/types'
 import { getValidations, type Validations } from '../../source/validations.js'
 
 describe('validations tests', () => {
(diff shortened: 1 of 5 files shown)


test/library/validation-test.ts
```diff
@@ -93,14 +93,16 @@ describe('validations tests', () => {
 	})
 
 	describe('forwardedHeader', () => {
-		it('should log an error when the Forwarded set to any value, but not when it is unset', () => {
+		it('should not log an error when the Forwarded header set is unset', () => {
```
gamemaker1:
```suggestion
		it('should not log an error when the Forwarded header is unset', () => {
```

gamemaker1 (approved):
LGTM. I do not see any backward-incompatible changes so I think this can be part of a minor release.

PR #652, opened 2026-07-08
docs: debugging guide, time constants, & v8.6.0 changelog

nfriedly:
Debugging guide is loosely based on https://expressjs.com/en/5x/guide/debugging/

Also added in docs for the new time constants

### DIFF
--- a/docs/guides/debugging.mdx
+++ b/docs/guides/debugging.mdx
@@ -1,0 +1,73 @@
+---
+title: Debugging Express Rate Limit
+description:
+  Learn how to enable and use debugging logs by setting the DEBUG environment
+  variable for enhanced troubleshooting.
+---
+
+Express Rate Limit supports debug logging similar to
+[Express](https://expressjs.com/en/5x/guide/debugging/), starting in v8.6.0
+
+To see all the internal logs used in Express Rate Limit, set the `DEBUG`
+environment variable to `express-rate-limit` when launching your app.
+
+```bash
+$ DEBUG=express-rate-limit node index.js
+```
+
+On Windows, use the corresponding command.
+
+```bash
+> $env:DEBUG = "express-rate-limit"; node index.js
+```
+
+Running this command on a basic configuration prints output similar to the
+following at instance creation:
+
+```bash
+$ DEBUG=express-rate-limit node server.js
+  express-rate-limit creating new rate limiter with 'MemoryStore' +0ms
+  express-rate-limit set windowMs to 900000 +2ms
+  express-rate-limit set limit to 100 +0ms
+  express-rate-limit set message to 'Too many requests, please try again later.' +0ms
+  express-rate-limit set statusCode to 429 +0ms
+  express-rate-limit set legacyHeaders to true +0ms
+  express-rate-limit set identifier to [Function: identifier] +0ms
+  express-rate-limit set requestPropertyName to 'rateLimit' +0ms
+  express-rate-limit set skipFailedRequests to false +0ms
+  express-rate-limit set skipSuccessfulRequests to false +0ms
+  express-rate-limit set requestWasSuccessful to [Function: requestWasSuccessful] +1ms
+  express-rate-limit set skip to [Function: skip] +0ms
+  express-rate-limit set keyGenerator to [AsyncFunction: keyGenerator] +0ms
+  express-rate-limit set ipv6Subnet to 56 +0ms
+  express-rate-limit set handler to [AsyncFunction: handler] +0ms
+  express-rate-limit set passOnStoreError to false +0ms
+  express-rate-limit set max to 100 +0ms
+  express-rate-limit set standardHeaders to false +0ms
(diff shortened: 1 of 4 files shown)


docs/guides/debugging.mdx
```diff
@@ -0,0 +1,73 @@
...
+  express-rate-limit set validations to { enabled: { default: true, validationsConfig: false, knownOptions: false, draftPolliHeaders: false, onLimitReached: false, keyGeneratorIpFallback: false, ipv6SubnetOrKeyGenerator: false }, disable: [Function (anonymous)], ip: [Function (anonymous)], trustProxy: [Function (anonymous)], xForwardedForHeader: [Function (anonymous)], forwardedHeader: [Function (anonymous)], positiveHits: [Function (anonymous)], unsharedStore: [Function (anonymous)], singleCount: [Function (anonymous)], limit: [Function (anonymous)], draftPolliHeaders: [Function (anonymous)], onLimitReached: [Function (anonymous)], headersDraftVersion: [Function (anonymous)], headersResetTime: [Function (anonymous)], knownOptions: [Function (anonymous)], validationsConfig: [Function (anonymous)], creationStack: [Function (anonymous)], ipv6Subnet: [Function (anonymous)], ipv6SubnetOrKeyGenerator: [Function (anonymous)], keyGeneratorIpFallback: [Function (anonymous)], windowMs: [Function (anonymous)] } +0ms
```
nfriedly: This doesn't seem optimal... but I can live with it for now.

docs/guides/debugging.mdx
```diff
@@ -0,0 +1,73 @@
...
+```bash
+  express-rate-limit requested '/' +50s
+  express-rate-limit request from ip '::ffff:127.0.0.1' +0ms
+  express-rate-limit computed key '127.0.0.1' +7ms
+  express-rate-limit incrementing count +0ms
+  express-rate-limit computed limit to be 100 +0ms
+  express-rate-limit computed used to be 1 +0ms
+  express-rate-limit computed remaining to be 99 +0ms
+  express-rate-limit computed resetTime to be 2026-07-08T21:49:50.255Z +0ms
+  express-rate-limit computed key to be '127.0.0.1' +1ms
```
nfriedly: hum.. are we computing the key twice?

gamemaker1: Oh that's actually because of these lines:

https://github.com/express-rate-limit/express-rate-limit/blob/main/source%2Frate-limit.ts#L469-L470

I could change it to print `set key to value` instead.

docs/guides/debugging.mdx
gamemaker1: I could ignore the validations key and then just log the config.validations.enabled property separately, I think that would be better right?

gamemaker1 (approved):
LGTM! Thanks for writing it out nicely.

I think I can make a patch for the two changes I mentioned above and then we can put out a release. Does that sound okay?

docs/guides/debugging.mdx
nfriedly: Yeah, although I think there's a second copy in the store that also gets logged. Like I said, I can live with this, but logging the validations enabled instead of the whole object separately probably makes sense. 

We could alternatively give the validations object a custom .toString() property instead of changing the debugging logic here. Same for the store.

docs/guides/debugging.mdx
nfriedly: Aah, I see. Yeah that's fine.

nfriedly:
> I think I can make a patch for the two changes I mentioned above and then we can put out a release. Does that sound okay?

Yeah that sounds good. Thanks!

docs/guides/debugging.mdx
gamemaker1: Okay, it turns out `debug` uses `util.inspect`, so I defined a custom inspect function that just returns the validations.enabled map. I also changed the `%o` format specifier to a `%O`, which makes it multiline and easier to read:

```
  express-rate-limit set ipv6Subnet to 56 +0ms
  express-rate-limit set standardHeaders to 'draft-6' +0ms
  express-rate-limit set store to MemoryStore {
  express-rate-limit   validations: undefined,
  express-rate-limit   previous: Map(0) {},
  express-rate-limit   current: Map(0) {},
  express-rate-limit   localKeys: true
  express-rate-limit } +0ms
  express-rate-limit set validations to {
  express-rate-limit   default: true,
  express-rate-limit   validationsConfig: false,
  express-rate-limit   knownOptions: false,
  express-rate-limit   draftPolliHeaders: false,
  express-rate-limit   onLimitReached: false,
  express-rate-limit   keyGeneratorIpFallback: false,
  express-rate-limit   ipv6SubnetOrKeyGenerator: false
  express-rate-limit } +0ms
  express-rate-limit set logger to { warn: [Function: warn], error: [Function: error] } +0ms
  express-rate-limit executing init for store +1ms
```

docs/guides/debugging.mdx
gamemaker1: Done, please see #653.

docs/guides/debugging.mdx
nfriedly: Hum, do you know why MemoryStore.validations is undefined in yours when it was set in mine?

nfriedly:
Note: waiting on https://github.com/express-rate-limit/express-rate-limit/pull/653 to get settled and merged in, then I'll update this branch.

docs/guides/debugging.mdx
gamemaker1: I passed `new MemoryStore()` directly to the rate limiter instead of the relying on the default store, which is passed as `new MemoryStore(validations)`.

PR #653, opened 2026-07-09
fix: make debug output easier to read

gamemaker1:
As per the comments in #652.

### DIFF
--- a/source/rate-limit.ts
+++ b/source/rate-limit.ts
@@ -356,7 +356,7 @@ const rateLimit = (
 
 	// Log the parsed options for this middleware instance.
 	for (const [key, val] of Object.entries(config))
-		debug('set %s to %o', key, val)
+		debug('set %s to %O', key, val)
 
 	// The limiter shouldn't be created in response to a request (usually)
 	config.validations.creationStack(config.store)
--- a/source/rate-limit.ts
+++ b/source/rate-limit.ts
@@ -356,7 +356,7 @@ const rateLimit = (
 
 	// Log the parsed options for this middleware instance.
 	for (const [key, val] of Object.entries(config))
-		debug('set %s to %O', key, val)
+		debug('set %s to %o', key, val)
 
 	// The limiter shouldn't be created in response to a request (usually)
 	config.validations.creationStack(config.store)
@@ -467,7 +467,7 @@ const rateLimit = (
 			}
 
 			for (const [key, val] of Object.entries(info))
-				debug('set %s to be %o', key, val)
+				debug('set request.%s.%s to be %o', config.requestPropertyName, key, val)
 
 			// Set the `current` property on the object, but hide it from iteration
 			// and `JSON.stringify`. See the `./types#RateLimitInfo` for details.
--- a/source/rate-limit.ts
+++ b/source/rate-limit.ts
@@ -467,7 +467,12 @@ const rateLimit = (
 			}
 
 			for (const [key, val] of Object.entries(info))
-				debug('set request.%s.%s to be %o', config.requestPropertyName, key, val)
+				debug(
+					'set request.%s.%s to be %o',
+					config.requestPropertyName,
+					key,
+					val,
+				)
 
 			// Set the `current` property on the object, but hide it from iteration
 			// and `JSON.stringify`. See the `./types#RateLimitInfo` for details.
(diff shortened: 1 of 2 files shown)


source/rate-limit.ts
```diff
@@ -356,7 +356,7 @@ const rateLimit = (
 
 	// Log the parsed options for this middleware instance.
 	for (const [key, val] of Object.entries(config))
-		debug('set %s to %o', key, val)
+		debug('set %s to %O', key, val)
```
nfriedly: So, this makes it multi-line? I'm not sure if we want that or not - I think the inspect symbol be enough on it's own.

source/validations.ts
```diff
@@ -554,5 +554,11 @@ export const getValidations = (
 			}
 	}
 
+	// When someone calls `util.inspect` on this object, print only the list of
+	// enabled validations without the list of function signatures.
+	const inspect = Symbol.for('nodejs.util.inspect.custom')
+	if (inspect)
+		(wrappedValidations as any)[inspect] = () => wrappedValidations.enabled
```
nfriedly, line 561: Nice, I didn't know about this one!

source/rate-limit.ts
```diff
@@ -467,7 +467,7 @@ const rateLimit = (
 			}
 
 			for (const [key, val] of Object.entries(info))
-				debug('computed %s to be %o', key, val)
+				debug('set %s to be %o', key, val)
```
nfriedly: Come to think of it, perhaps we should specify what we're setting here, something like


```suggestion
				debug('set request.%s.%s to be %o', config.requestPropertyName, key, val)
```

nfriedly (commented):
This looks mostly good, but I'm not sure about the multi-line thing. I think it may be preferable for each log to just be a single line. 

Also, perhaps we want to define the inspect symbol property on the MemoryStore also so that it isn't logging an extra copy of the validations config?

source/rate-limit.ts
gamemaker1: Done

source/rate-limit.ts
gamemaker1: Reverted it to %o, where it is single line.

gamemaker1:
```
  express-rate-limit set ipv6Subnet to 56 +0ms
  express-rate-limit set standardHeaders to 'draft-6' +1ms
  express-rate-limit set store to MemoryStore { validations: { default: true, validationsConfig: false, knownOptions: false, draftPolliHeaders: false, onLimitReached: false, keyGeneratorIpFallback: false, ipv6SubnetOrKeyGenerator: false }, previous: Map(0) {}, current: Map(0) {}, localKeys: true } +0ms
  express-rate-limit set validations to { default: true, validationsConfig: false, knownOptions: false, draftPolliHeaders: false, onLimitReached: false, keyGeneratorIpFallback: false, ipv6SubnetOrKeyGenerator: false } +0ms
  express-rate-limit set logger to { warn: [Function: warn], error: [Function: error] } +0ms
  express-rate-limit executing init for store +1ms

  express-rate-limit requested '/' +5s
  express-rate-limit request from ip '::ffff:127.0.0.1' +1ms
  express-rate-limit computed key '127.0.0.1' +14ms
  express-rate-limit incrementing count +0ms
  express-rate-limit set request.rateLimit.limit to be 2 +1ms
  express-rate-limit set request.rateLimit.used to be 1 +0ms
  express-rate-limit set request.rateLimit.remaining to be 1 +0ms
  express-rate-limit set request.rateLimit.resetTime to be 2026-07-16T09:46:11.574Z +0ms
  express-rate-limit set request.rateLimit.key to be '127.0.0.1' +1ms
  express-rate-limit set ietf draft 6 headers +0ms
```

nfriedly (approved):
This looks good. The only think I'm a little iffy about is the `MemoryStore` logging the `validations` object, ~~but `MemoryStory` only even uses `validations` once, to check the windowMs value - so we could check that in the `parseOptions()` function and then not pass `validations` to the memory store.~~

Edit: nevermind that idea. The check is specific to the memory store, because it's checking for the max value for node.js's `setInterval` function - other stores could support higher values. I suppose we could make the store accept the `config` (which includes the `validations` object) instead of the `options` (which doesn't), but that's an API change that I don't really want to do today. 

So, let's just merge this as-is and call it good enough for now.

PR #655, opened 2026-07-11
Feature/654 support human readable time strings for window ms

### LINKED ISSUE #654: Enhance express-rate-limit windowMs option
### Summary
Currently, `windowMs` requires developers to manually calculate the window duration in milliseconds (e.g. `10800000` for 3 hours). This is error-prone and hurts readability. I'd like to propose support for human-readable time strings as an alternative.

### Motivation
Calculating milliseconds by hand is tedious and easy to get wrong, especially for longer windows like hours or days. Allowing a string like `"3hr"` instead of `10800000` makes configs more readable and less error-prone, while still preserving the option to use raw milliseconds for developers who prefer it.

### Proposed API
```js
app.use(rateLimit({
  windowMs: "3hr", // instead of 10800000
  limit: 10,
}));
```

### Supported formats
- Numeric milliseconds (existing behavior) — unchanged, fully backwards compatible.
- String shorthand with or without a space between the number and unit:
  - `"3 min"` / `"3min"`
  - `"3 hr"` / `"3hr"`
  - `"3 day"` / `"3day"`

### Backwards compatibility
- If `windowMs` is a number, behavior is unchanged.
- If `windowMs` is a string, it gets parsed into milliseconds internally before being used.
- Invalid strings should throw a clear, descriptive error rather than failing silently.

### Open questions
- Should we support additional units (e.g. `"sec"`, `"ms"`, `"week"`)?
- Should parsing be case-insensitive (e.g. `"3HR"`, `"3Hr"`)?
- Is there an existing lightweight dependency (e.g. `ms`) we'd want to use for parsing, or should this be implemented in-house to avoid adding a dependency?

Should i work on it ?

### Why

Currently, `windowMs` requires developers to manually calculate the window duration in milliseconds (e.g. `10800000` for 3 hours). This is error-prone and hurts readability. I'd like to propose support for human-readable time strings as an alternative.

### Proposed API
```js
app.use(rateLimit({
  windowMs: "3hr", // instead of 10800000
  limit: 10,
}));
```

bit-64kce:
<html><head></head><body><h1>Readable <code>windowMs</code> with Time Constants</h1>
<p>Instead of computing milliseconds by hand, <code>express-rate-limit</code> exports time constants (<code>MINUTE</code>, <code>HOUR</code>, <code>DAY</code>) so you can express <code>windowMs</code> in plain, readable math.</p>
<pre><code class="language-js">import { rateLimit, HOUR } from 'express-rate-limit'

const limiter = rateLimit({
	windowMs: 3 * HOUR, // 3 hours, instead of 3 * 60 * 60 * 1000
	limit: 100,         // limit each IP to 100 requests per windowMs
})
</code></pre>
<h2>Available constants</h2>

Constant | Value (ms)
-- | --

HOUR | 60 * MINUTE
DAY | 24 * HOUR

<h2>Examples</h2>
<pre><code class="language-js">windowMs: 15 * MINUTE   // 15 minutes
windowMs: 1 * DAY       // 1 day
</code></pre>
<p>This avoids error-prone manual multiplication (e.g. <code>3 * 60 * 60 * 1000</code>) and makes the intent of the rate-limit window immediately clear at a glance.</p></body></html>

Fixes #654


tsconfig.json
```diff
@@ -3,6 +3,8 @@
 	"exclude": ["node_modules/"],
 	"extends": "@express-rate-limit/tsconfig",
 	"compilerOptions": {
-		"target": "ES2020"
+		"target": "ES2020",
+		"jsx": "react-jsx",
+		"types": ["react"]
```
nfriedly: This change shouldn't be needed

package-lock.json
nfriedly, line 1: Please remove this file also; we use pnpm to manage dependencies

nfriedly (commented):
Thanks! The core change looks good, but there's a bit of extra stuff here that should be removed before we merge this in.

package-lock.json
bit-64kce: Ok will make sure to do that

tsconfig.json
bit-64kce: it is unable to build due without the change on my local pc is this a build problem or my pc specific problem ?

this is the error i get :
node_modules/@types/mdx/types.d.ts(23,38): error TS2503: Canno
node_modules/@types/mdx/types.d.ts(28,23): error TS2503: Canno
node_modules/@types/mdx/types.d.ts(47,28): error TS2503: Canno
t find namespace 'JSX'.
node_modules/@types/mdx/types.d.ts(73,46): error TS2503: Canno
t find namespace 'JSX'.

Error: Compiled with errors

source/memory-store.ts
nfriedly, line 1: Please undo this change.

source/utils.ts
nfriedly, line 1: Please undo this change also. This would only be needed with the original string parsing design, not the trimmed down constant-based version we decided to go with.

source/index.ts
```diff
@@ -6,10 +6,15 @@ export { ipKeyGenerator } from './ip-key-generator.js'
...
 // Export the rateLimit function as a default export and as a named export, if
 // the default export does not work (see https://github.com/nfriedly/express-rate-limit/issues/280)
-export { default, default as rateLimit } from './rate-limit.js'
+export {
+	DAY,
+	default,
+	default as rateLimit,
+	HOUR,
+	MINUTE,
+} from './rate-limit.js'
```
nfriedly: You're missing `SECOND` here, although perhaps we should move these constants to a new file like `time-constants.ts` or similar, then we could just do `export * from './time-constants.js'`

tsconfig.json
nfriedly: It might be from using `npm install` instead of `pnpn install  --frozen-lockfile` (or `npx pnpm install --frozen-lockfile` if you don't already have pnpm installed locally). Something in our dependency tree can get out of whack with a regular npm install, but pnpm gets it to behave correctly.

See: https://express-rate-limit.mintlify.app/guides/contributing

readme.md
```diff
@@ -21,10 +21,10 @@ The [full documentation](https://express-rate-limit.mintlify.app/overview) is
 available on-line.
 
 ```ts
-import { rateLimit } from 'express-rate-limit'
+import { rateLimit, MINUTE } from 'express-rate-limit'
 
 const limiter = rateLimit({
-	windowMs: 15 * 60 * 1000, // 15 minutes
+	windowMs: 15 * MINUTE, // 15 minutes
```
nfriedly: Let's update the comment too:
```suggestion
	windowMs: 15 * MINUTE, // SECOND, MINUTE, HOUR, and DAY constants are available, or a use bare number for milliseconds
```

source/utils.ts
bit-64kce: very true, my bad

nfriedly (approved):
Great work! 

I'm going to make a couple of minor formatting changes and then merge it in. (Extra spacing in the example, and restore the original formatting in the otherwise unchanged parts of index.ts.)

We have a couple of things in progress right now, and I think we're going to roll them all into a single release, but that should be sometime within the next week or two.

nfriedly:
This is now published to npm in version 8.6.0 - https://www.npmjs.com/package/express-rate-limit - thanks again!

bit-64kce:
thank you too for your patience, i really appreciate it

nfriedly:
Hey, just a heads up so that you know what's going on: we're going to deprecate these values so that we can replace them with second-based values in the next major release. More context on https://github.com/express-rate-limit/express-rate-limit/pull/661

bit-64kce:
Okay please I would love to be involved in the project if there is anything
I can do to help

nfriedly:
I'm still thinking about this, see my idea on https://github.com/express-rate-limit/express-rate-limit/discussions/678

PR #659, opened 2026-07-14
fix: Pin safe version of @asyncapi/specs dev dep

nfriedly:
Fixes https://github.com/express-rate-limit/express-rate-limit/issues/658 by replacing the safe version we were already using as a transitive dependency with a development dependency pinned to the same version. This should prevent any accidental upgrades to the vulnerable version.


nfriedly:
FYI @kunalsingh-safedep

Issue #309, opened 2022-07-09
Per-request cost


drmrbrewer:
### Description

First, thanks for this really useful module!

So far as I can tell, it's possible at present to set `max`, and each request will count as one towards that max quota.

I'd find it useful to have a definable function which can set a per-request 'cost', so that some requests count as `1` towards the quota, while others count as more, e.g. `2` or more.

### Why

If a request is using more server resources (based on the options passed in) then it should deplete its allowed quota more quickly and be rate limited sooner.

### Alternatives

Could add multiple different limiters, each with different criteria and different `max`, but this seems overkill when a simple `cost` function could be used.

gamemaker1:
This sounds like a good idea! We could add a number/function option just like `max` called `cost`.

However, this will require us to make changes to the `Store` interface - `increment` and `decrement` will have to become `incrementBy` and `decrementBy`. Or we could just call `increment` and `decrement` multiple times.

What do you think @nfriedly?

Also, if we decide to implement this feature, @drmrbrewer would you like to make a PR for it?

drmrbrewer:
I've just noticed that [`max` can be a function,](https://github.com/nfriedly/express-rate-limit#max) so this is useful in my context (e.g. `max` could be reduced if the request is more resource intensive than normal), but I still think that a separate `cost` number/function would also be useful for additional control.

drmrbrewer:
@nfriedly thanks for the link, but I way prefer the relative simplicity of `express-rate-limit`, and a simple `cost` number/function would be a useful addition :-)

Issue #122, opened 2019-02-20
[Request] Change windowMs to function


pedrosodre:
Hey there,

in order to allow different windowMs for each keyGenerator, would be cool if windowMs became a function to allow set different milliseconds for different clients.

Cheers

nfriedly:
I've thought about this, but it would require some significant changes to the default memory store, so it's not quite as simple as it sounds.

pedrosodre:
I thought it is not as simple as it sounds, at least for memory store. For mongo, for example, maybe is a bit more simple.

In any case, I'll look forward for this feature, whenever it comes (or maybe I try out fork and work on it in near future) :)

Thank you!

bvallelunga:
This is definitely a feature I would like as well!

ovxrfl0w:
> This is definitely a feature I would like as well! If we could also do the same for the max that would be great.

Max already supports it.

brendon-stephens:
+1 for this feature request

kriskate:
upvoting this too

nfriedly:
I haven't forgotten about this request, but it would require a new Store API and significant reworking of the optimizations that various stores perform.

However, I wanted to mention that if you only have a handful of specific window sizes that you want, this can be accomplished by having multiple instances of express-rate-limit, with different window sizes, and then custom skip functions to ensure that only one applies to each request. For example:

* Anonymous users get 60 requests per minute
* Regular users get 10000 requests per hour
* Enterprise users get 999999 requests per hour

```js
import { rateLimit } from 'express-rate-limit'

// authenticate users before applying the rate limits!
app.use((req, res, next) => {
	if ( /* user is on an enterprise account */ ) {
		req.accountType = 'enterprise';
	} else if ( /* user is on a regular account */ ) {
		req.accountType = 'regular';
	}
	// all others will be treated as anonymous
	next();
}

const globalRateLimitSettings = {
	standardHeaders: 'draft-7', // draft-6: `RateLimit-*` headers; draft-7: combined `RateLimit` header
	legacyHeaders: false, // Disable the `X-RateLimit-*` headers.
	// ...
}

// anonymous limit
app.use(rateLimit({
	...globalRateLimitSettings,
	skip: req => req.accountType, // any accountType means skip this limiter
	windowMs: 60 * 1000, // 1 minute
	limit: 60,
	// store: ... , // be sure to set a different `prefix` for each store instance if using an external store!
}))

// regular limit
app.use(rateLimit({
[...]

PR #661, opened 2026-07-23
feat: add retryAfter option

### LINKED ISSUE #441: Enhanced Rate Limiting with a retryAfter option and IP Blocking Features for Improved Flexibility
### Description

Hello,

The main goal is to be able to set a different 'retry after' time. Additionally, I'd like the possibility to permanently ban certain IPs.

I propose having separate configuration parameters, something like this:

```js
app.use(rateLimit({
   windowMs: 1000, // 1seg
   limit: 10, // 10 req per sec
   /// NEW
   retryAfter:  60 * 60 * 1000  or async (req, ip) => number,   // retry in 1h
   notAllowed: [ '123.123.1.1']  or async  (req, ip) => bool
}):
```

Having the possibility to accept a function would make the system more flexible, especially in cases where the value varies depending on the user or IP.

### Why

There are many cases where you might want to allow a certain number of requests within a specific time window. However, this doesn't necessarily mean that access should be regained immediately after that same time window.

The main problem is that if I set `windowMs=1000` (1 second), users will be able to retry after just 1 second. But what I'd like to do is restrict this to a different time frame, such as 1 hour. Unfortunately, this isn't currently possible.

Additionally, I would like an option to block certain IPs. While I could create a separate middleware to check IPs and block them, it would be more suitable to leverage this library's existing advanced checks, such as proxy detection.

### Alternatives

N/A

sunnyrishy:
Closes #441.

Adds a retryAfter option that lets you customize the Retry-After header value sent on a 429 response. Can be a number or a (sync/async) function, matching the pattern already used for limit/message/identifier.

This only covers the advisory piece of the original request. As discussed in the thread, actually enforcing a longer block period (the "cool down period" idea) would need changes to how every store handles resets, so that part is out of scope here. This just changes what the header says, not how long the limiter keeps blocking.

Scoped narrowly to the plain Retry-After header. Doesn't touch the RateLimit-Reset value inside the draft-6/7/8 standardHeaders, since that represents the window's reset time (sent on every response) rather than a cooldown that only applies after a block (sent only on 429s), a different concept.

Made the option optional rather than giving it a forced default function. The natural default needs info.resetTime, which isn't available to a function shaped like the other options, (request, response) => value. 
Falls back to the existing reset time calculation when unset, so no behavior change for anyone not using it.

Added two tests (number and function forms) plus a docs section under Configuration.

Let me know if I've misread the intended scope anywhere.

### DIFF
--- a/docs/reference/configuration.mdx
+++ b/docs/reference/configuration.mdx
@@ -281,6 +281,37 @@ the Express `req` and `res` objects and then returns a string.
 Defaults to a function that returns a string of the format
 `{limit}-in-{window}`, e.g., `7-in-1.5sec`.
 
+## `retryAfter`
+
+> `number | function`
+
+Custom value (in seconds) for the `Retry-After` header sent when a client is
+rate limited.
+
+Can be the value itself as a number or a (sync/async) function that accepts the
+Express `req` and `res` objects and then returns a number.
+
+<Note>
+
+This is advisory only. It changes what the `Retry-After` header says, it does
+not change how long the limiter actually keeps blocking the client for. That is
+still controlled by [`windowMs`](#windowms).
+
+</Note>
+
+By default, the number of seconds remaining until the window resets.
+
+```ts
+const limiter = rateLimit({
+	// ...
+	retryAfter: async (req, res) => {
+		// e.g. always tell clients to back off for a fixed 10 minutes,
+		// regardless of how the window is configured.
+		return 10 * 60
+	},
+})
+```
+
 ## `store`
 
 > `Store`
--- a/docs/reference/configuration.mdx
+++ b/docs/reference/configuration.mdx
@@ -307,7 +307,7 @@ const limiter = rateLimit({
 	retryAfter: async (req, res) => {
 		// e.g. always tell clients to back off for a fixed 10 minutes,
 		// regardless of how the window is configured.
-		return 10 * 60
+		return 10 * 60 // seconds, unlike windowMs
 	},
 })
 ```
--- a/docs/reference/configuration.mdx
+++ b/docs/reference/configuration.mdx
@@ -281,15 +281,21 @@ the Express `req` and `res` objects and then returns a string.
 Defaults to a function that returns a string of the format
 `{limit}-in-{window}`, e.g., `7-in-1.5sec`.
 
-## `retryAfter`
+## `retryAfterMs`
 
 > `number | function`
 
-Custom value (in seconds) for the `Retry-After` header sent when a client is
-rate limited.
+Custom value, in milliseconds, used to compute the `Retry-After` header sent
+when a client is rate limited. Uses the same unit as [`windowMs`](#windowms) and
+the `DAY`/`HOUR`/`MINUTE`/`SECOND` constants.
 
 Can be the value itself as a number or a (sync/async) function that accepts the
 Express `req` and `res` objects and then returns a number.
+
+When set, the value is added to the current time and sent as an
+[HTTP-date](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Retry-After#syntax)
+rather than a plain delay in seconds, since HTTP dates don't carry millisecond
+precision, the header may be off by up to a second from the exact value given.
 
 <Note>
 
@@ -299,15 +305,18 @@ still controlled by [`windowMs`](#windowms).
 
 </Note>
 
-By default, the number of seconds remaining until the window resets.
+By default, this is unset, and the header instead reports the number of seconds
+remaining until the window resets.
 
 ```ts
+import { rateLimit, MINUTE } from 'express-rate-limit'
+
 const limiter = rateLimit({
 	// ...
-	retryAfter: async (req, res) => {
+	retryAfterMs: async (req, res) => {
 		// e.g. always tell clients to back off for a fixed 10 minutes,
 		// regardless of how the window is configured.
-		return 10 * 60
+		return 10 * MINUTE
 	},
 })
 ```
--- a/docs/reference/configuration.mdx
+++ b/docs/reference/configuration.mdx
@@ -281,21 +281,15 @@ the Express `req` and `res` objects and then returns a string.
 Defaults to a function that returns a string of the format
 `{limit}-in-{window}`, e.g., `7-in-1.5sec`.
 
-## `retryAfterMs`
+## `retryAfter`
 
 > `number | function`
 
-Custom value, in milliseconds, used to compute the `Retry-After` header sent
-when a client is rate limited. Uses the same unit as [`windowMs`](#windowms) and
-the `DAY`/`HOUR`/`MINUTE`/`SECOND` constants.
+Custom value (in seconds) for the `Retry-After` header sent when a client is
+rate limited.
 
 Can be the value itself as a number or a (sync/async) function that accepts the
 Express `req` and `res` objects and then returns a number.
-
-When set, the value is added to the current time and sent as an
-[HTTP-date](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Retry-After#syntax)
-rather than a plain delay in seconds, since HTTP dates don't carry millisecond
-precision, the header may be off by up to a second from the exact value given.
 
 <Note>
 
@@ -305,18 +299,15 @@ still controlled by [`windowMs`](#windowms).
 
 </Note>
 
-By default, this is unset, and the header instead reports the number of seconds
-remaining until the window resets.
+By default, the number of seconds remaining until the window resets.
 
 ```ts
-import { rateLimit, MINUTE } from 'express-rate-limit'
-
 const limiter = rateLimit({
 	// ...
-	retryAfterMs: async (req, res) => {
+	retryAfter: async (req, res) => {
 		// e.g. always tell clients to back off for a fixed 10 minutes,
 		// regardless of how the window is configured.
-		return 10 * MINUTE
+		return 10 * 60
 	},
 })
 ```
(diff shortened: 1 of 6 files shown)


nfriedly:
> The natural default needs info.resetTime, which isn't available to a function shaped like the other options, (request, response) => value.

Actually, it is available if the store provided it, because it's added to the `req` object before your new function is called:

https://github.com/express-rate-limit/express-rate-limit/blob/fad3d853e679b246f9637413aa500050b610b06a/source/rate-limit.ts#L485-L486

That said, we'd have to do some typescript trickery, because the ValueDeterminingMiddleware<T> type passes in `Request`, not our `AugmentedRequest` (because a lot of them run before the request gets augmented.)

But, all that said, what's here is probably fine.

docs/reference/configuration.mdx
```diff
@@ -281,6 +281,37 @@ the Express `req` and `res` objects and then returns a string.
...
+
+By default, the number of seconds remaining until the window resets.
+
+```ts
+const limiter = rateLimit({
+	// ...
+	retryAfter: async (req, res) => {
+		// e.g. always tell clients to back off for a fixed 10 minutes,
+		// regardless of how the window is configured.
+		return 10 * 60
```
nfriedly: probably worth calling out again that this is seconds, unlike `windowMs`

```suggestion
		return 10 * 60 // seconds, unlike windowMs
```

nfriedly (commented):
Everything here looks good on it's own, but I don't love that this operates on seconds while windowMs is in milliseconds, as are the new time helpers we just added. I feel like it'd be very easy to confuse those two, and there probably isn't a great way to distinguish between the two.

Perhaps we can add a validation check that warns if this is evenly divisible by 1000 (and alternatively, if windowMs *isn't* divisible by 1000), but I'd prefer to come up with a design that was easy to get correct rather than one that needs that warning.

I'm also considering adding a window or windowSeconds option or similar, but then the time helpers we just shipped wouldn't work there either. 

You did good work here, but I'd like to think about that a bit more before we merge and ship this.

@gamemaker1 I'd appreciate your thoughts on the issue also.

gamemaker1:
> Made the option optional rather than giving it a forced default function. The natural default needs info.resetTime, which isn't available to a function shaped like the other options, (request, response) => value.
Falls back to the existing reset time calculation when unset, so no behavior change for anyone not using it.

This is fine, even if anyone wants to use the default value in some if-condition they can just return undefined from the function.

gamemaker1:
> Perhaps we can add a validation check that warns if this is evenly divisible by 1000 (and alternatively, if windowMs isn't divisible by 1000), but I'd prefer to come up with a design that was easy to get correct rather than one that needs that warning.

Yeah, these warnings would also misfire on values that are intentionally multiples of 1000.

What about renaming this option to `retryAfterMs`, adding the milliseconds to the current date, and then return a [`Date` in the `Retry-After` header instead](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Retry-After#syntax)?

That way we do not need the validation, we keep the public API consistent and use milliseconds everywhere, and we try to keep the returned `Retry-After` header as close to the actual value as possible as well. We can add a disclaimer to the docs that it might not be as exact since [HTTP dates do not have milliseconds/nanoseconds in them](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Date).

sunnyrishy:
Renamed to retryAfterMs and changed the header value to an HTTP-date instead of a plain number, per your suggestion. It's added to the branch now.

Ran the full test suite, lint, and a type check locally, all green (243 tests passing). Also updated the docs section and the example to use the MINUTE constant so it stays consistent with the rest of the API.

Let me know if this is closer to what you had in mind.

nfriedly:
> This is fine, even if anyone wants to use the default value in some if-condition they can just return undefined from the function.

Good point!

> Yeah, these warnings would also misfire on values that are intentionally multiples of 1000.

Yeah, probably not a great idea.

> What about renaming this option to retryAfterMs, 

I like this better

> adding the milliseconds to the current date, and then return a [Date in the Retry-After header instead](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Retry-After#syntax)?

I don't like this. It's a breaking change, and I personally prefer seconds because then you don't have to worry about clocks being on the wrong timezone or anything along those lines.

I could see allowing the new option to return a `Date` object, and then using that, but then the name `retryAfterMs` wouldn't make sense.

---

Here's something I'm considering, but I'm not fully set on it yet:

* Ship a new semver-major version
* Add a new `window` or `windowSeconds` config option
* Soft-deprecate windowMs (remove it from the examples, but it continues to work)
* Delete the existing time helpers (`DAY`, `HOUR`, etc.)
* Add new time helpers with an `S` suffix (`DAYS`, `HOURS`, etc.) - ("S" for "seconds" but also reads a little nicer, e.g. "5 * MINUTES" instead of "5 * MINUTE") 
* Go back to the original version of this PR, with retry-after being in seconds, but now the time helpers work for both it and window.
* Optionally, extend the retryAfter option to support functions that return a Date object in addition to supporting numbers for seconds

That's a pretty significant change, but I'm considering it because I think it makes for a better public API. Retry-After makes sense in seconds, because that's the exact value we put in the header. The window size, on the other hand, doesn't have any "natural" unit, and milliseconds is honestly probably more precise than necessary.

The only breaking change is the removal of the existing time helpers, and that should have a relatively small impact because of how recently it shipped. Everything else here is backwards compatible.

Oh, and switching the window to seconds also aligns better with the `t` / effective window property of the draft headers standard: https://ietf-wg-httpapi.github.io/ratelimit-headers/draft-ietf-httpapi-ratelimit-headers.html#name-effective-window-parameter

gamemaker1:
My only concern about moving away from milliseconds to seconds is that it may require [breaking](https://github.com/express-rate-limit/rate-limit-memcached/blob/73162ff72a22a11c8a6cfcfe636c0af99e460c80/source/memcached-store.ts#L99) [changes](https://github.com/express-rate-limit/rate-limit-redis/blob/770f64e18e80624be79e0b9c3ddd1c7f0128269b/source/scripts.ts#L14) to stores, plus it will break for users who use millisecond-level precision...

> Oh, and switching the window to seconds also aligns better with the t / effective window property of the draft headers standard: https://ietf-wg-httpapi.github.io/ratelimit-headers/draft-ietf-httpapi-ratelimit-headers.html#name-effective-window-parameter

...but given that the official standard does not seem to mention millisecond time windows anywhere, and examples like the Memcached store, which [converts the milliseconds to seconds anyways](https://github.com/express-rate-limit/rate-limit-memcached/blob/73162ff72a22a11c8a6cfcfe636c0af99e460c80/source/memcached-store.ts#L99), I think the second concern can be set aside.

A passing thought: if we are going to make a coordinated breaking change across stores, perhaps it might be a good idea to consider implementing some other things like #122 or #309.

For now, reverting this PR to the original commit and merging is something we should go ahead with. I can work on the remaining things as separate PRs. Does that sound good?

nfriedly:
> My only concern about moving away from milliseconds to seconds is that it may require [breaking](https://github.com/express-rate-limit/rate-limit-memcached/blob/73162ff72a22a11c8a6cfcfe636c0af99e460c80/source/memcached-store.ts#L99) [changes](https://github.com/express-rate-limit/rate-limit-redis/blob/770f64e18e80624be79e0b9c3ddd1c7f0128269b/source/scripts.ts#L14) to stores, plus it will break for users who use millisecond-level precision...

We can still calculate windowMs and pass it to stores, so it doesn't have to be a breaking change. (Let's calculate whichever one is missing and pass both to stores, so that any store that want's to use seconds rather than milliseconds can have a simple peer dependency on v9+)

Also, I think I like the name `window` over `windowSeconds`. I know the later is more precise, but literally everyone will have to write this, and also with our helpers, the unit won't matter as much.

gamemaker1:
> We can still calculate windowMs and pass it to stores, so it doesn't have to be a breaking change. (Let's calculate whichever one is missing and pass both to stores, so that any store that want's to use seconds rather than milliseconds can have a simple peer dependency on v9+)

Sure, that would be good.

> Also, I think I like the name window over windowSeconds. I know the later is more precise, but literally everyone will have to write this, and also with our helpers, the unit won't matter as much.

Yes, window would be much better!

nfriedly:
I think I'm going to put out a quick update deprecating the new time helpers, so users grabbing the library today won't get instructions to use something we're taking away in the next release.

nfriedly:
One idea: instead of removing the existing `DAY` / `HOUR` / etc. we could replace them with `NaN`'s and then add a new validation that looks for `NaN` in `windowMs` and provides a more helpful error. Not sure if I like that more than just removing them, but it's a thought.

gamemaker1:
> One idea: instead of removing the existing DAY / HOUR / etc. we could replace them with NaN's and then add a new validation that looks for NaN in windowMs and provides a more helpful error. Not sure if I like that more than just removing them, but it's a thought.

I don't think that's necessary if we're going to do a major bump.

sunnyrishy:
Reverted back to the seconds-based retryAfter per this discussion, plus picked up the earlier inline suggestion about calling out the units in the example comment. Sounds good on keeping the bigger windowMs/time-unit rework as a separate effort, this PR should be back to a mergeable state now. Ran full tests/lint/typecheck locally, all green.

nfriedly:
Sorry I haven't gotten back to this yet, it's still on my radar.

Another thing that just came to mind is that we'll probably want to do a major version bump for express-slow-down also to make it work with seconds for the window and delay.

gamemaker1:
I'll try to take a stab at this today.

nfriedly (approved):
Apologies for the delay. I think you're right that this is in good shape and ready to merge. I'll get a release out with this shortly.

PR #662, opened 2026-07-26
chore: deprecate millisecond-based time constants

nfriedly:
We're going to remove these in the next major release and replace them with a slightly different variation that's based on seconds rather than milliseconds.

See discussion on https://github.com/express-rate-limit/express-rate-limit/pull/661 and also previous work at https://github.com/express-rate-limit/express-rate-limit/pull/655

### DIFF
--- a/source/time-constants.ts
+++ b/source/time-constants.ts
@@ -1,20 +1,19 @@
 /**
- * Time constants in milliseconds
- *
- * Use these with rate limit window calculations:
- * @example
- * // 5 minute window
- * windowMs: 5 * MINUTE
- *
- * // 2 hour window
- * windowMs: 2 * HOUR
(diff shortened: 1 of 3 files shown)


source/time-constants.ts
```diff
@@ -11,6 +11,8 @@
  *
  * // 1 day window
  * windowMs: 1 * DAY
+ *
+ * @deprecated these are being replaced in the next major version
```
gamemaker1, line 2: We should probably add this annotation individually to each `const` so that it appears in the type docs for them.

nfriedly: Good call, I'll do that

gamemaker1:
Rest looks good.

